PR #2, opened 2015-10-03
Crystal 0.8.0 support

veelenga:
- shards as dependency manager
 - `TimeFormat` -> `Time::Format`

### DIFF
--- a/Projectfile
+++ b/Projectfile
@@ -1,5 +1,0 @@
-deps do
-  github "waterlink/crystal-mysql"
-  github "will/crystal-pg"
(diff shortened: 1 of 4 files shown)


veelenga:
Not sure why tests are failing. Wrong db setup.

drujensen:
:+1: Thanks!

drujensen:
@veelenga I changed this to development_dependencies so it won't install all three db drivers.  You need to pick which driver(s) you want in your project.

I still haven't resolved why sqlite is failing in travis.

veelenga:
Good catch with development_dependencies :+1:

PR #18, opened 2016-11-25
decouple find from id

azah:
added specs as well, but I only tested the mysql spec. (PG won't install on ubuntu bash for windows, and I'm not sure about sqlite but I don't have it installed)

also I ran `crystal tool format`.


drujensen:
Thanks!  I will add some documentation to the readme

PR #25, opened 2017-04-15
Use DATABASE_URL by default if exists.

elorest:
https://github.com/drujensen/kemalyst-model/issues/24


elorest:
Thanks. It will make live coding for 5-10 minutes at railsconf much easier. I figure if I can build a blog in 5 minutes on stage it will beat DHH's 15 minutes from 12 years ago and convert a few people

drujensen:
@elorest Your right.  This is much easier.  Sometimes magic is good.  Thanks for the contributions.

PR #26, opened 2017-04-20
use finished macro - breaking change

drujensen:
@elorest @TechMagister I am introducing a breaking change to the models.  There is a new feature added to crystal that provides a `finished` macro to allow you to only process them at the end.

This allows us to have a model that looks like this:

```crystal
require "../src/adapter/mysql"

class Post < Kemalyst::Model
  adapter mysql
  field name : String
  field body : Text
  field total : Int32
  field draft : Bool
  timestamps
end
```

thoughts?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM drujensen/crystal:0.21.0
+FROM drujensen/crystal:0.21.1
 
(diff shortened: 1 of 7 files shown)


elorest:
Can you add this into the generators too.

PR #28, opened 2017-04-29
Added support for custom primary key column name and type

lhz:
Implements feature as discussed in issue https://github.com/drujensen/kemalyst-model/issues/27

### DIFF
--- a/src/kemalyst-model.cr
+++ b/src/kemalyst-model.cr
@@ -7,6 +7,7 @@ class Kemalyst::Model
   macro inherited
     include Kemalyst::Validators
 
+    PRIMARY = {name: id, type: Int64}
     FIELDS = {} of Nil => Nil
     SETTINGS = {} of Nil => Nil
 
@@ -35,6 +36,12 @@ class Kemalyst::Model
     {% SETTINGS[:table_name] = name.id %}
   end
 
+  # specify the primary key column and type
+  macro primary(decl)
+    {% PRIMARY[:name] = decl.var %}
+    {% PRIMARY[:type] = decl.type %}
+  end
+
   # specify the fields you want to define and types
   macro field(decl)
     {% FIELDS[decl.var] = decl.type %}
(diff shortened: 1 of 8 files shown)


src/kemalyst-model.cr
```diff
@@ -101,99 +111,104 @@ class Kemalyst::Model
       {% end %}
       return params
    end
-  end # End of Fields Macro
```
lhz: Lots of changed lines below due to continuing the macro block to include the rest of the instance methods.

src/kemalyst-model.cr
```diff
@@ -101,99 +111,104 @@ class Kemalyst::Model
...
+          params_and_pk << value
+          @@adapter.update @@table_name, @@primary_name, self.class.fields, params_and_pk
+        else
+          @created_at = Time.now
+          @updated_at = Time.now
+          {% if primary_type.id == "Int32" %}
+            @{{primary_name}} = @@adapter.insert(@@table_name, self.class.fields, params).to_i32
+          {% else %}
+            @{{primary_name}} = @@adapter.insert(@@table_name, self.class.fields, params)
+          {% end %}
```
lhz: Any other likely types for autoincremented primary keys than `Int32` and `Int64`? Not that I can think of.

src/kemalyst-model.cr
```diff
@@ -48,10 +55,13 @@ class Kemalyst::Model
   macro process_fields
     {% name_space = @type.name.downcase.id %}
     {% table_name = SETTINGS[:table_name] || name_space + "s" %}
+    {% primary_name = PRIMARY[:name] %}
+    {% primary_type = PRIMARY[:type] %}
     # Table Name
     @@table_name = "{{table_name}}"
+    @@primary_name = "{{primary_name}}"
```
lhz: These class variables may no longer be needed, since all instance methods have been moved inside the macro block, so we can use the macro expressions directly instead.

src/kemalyst-model.cr
drujensen: I could see someone wanting to use a UUID as a primary key but not sure how to implement that.  Maybe something for a future change.

drujensen:
@lhz thank you for the contribution!

PR #32, opened 2017-05-04
added ability to cast to appropriate types within the model

elorest:
so you can just send it string values and don't have to do conversions in the controller.

```cr
p = Post.new
# works with namedtuple
p.set_attributes(title: "auto casting saves time", description: "I love this feature",  tag_id: 35)
# or hash
p.set_attributes({:title => "auto casting saves time", :description => "I love this feature",  :tag_id => 35})
```

I played around with tests but there don't seem to be any existing ones that work for this object. If you could get a couple base things working for testing the model I'll add some for this.

### DIFF
--- a/src/kemalyst-model.cr
+++ b/src/kemalyst-model.cr
@@ -53,7 +53,7 @@ class Kemalyst::Model
   end
 
   macro process_fields
-    {% name_space = @type.name.gsub(/::/,"_").downcase.id %}
+    {% name_space = @type.name.gsub(/::/, "_").downcase.id %}
     {% table_name = SETTINGS[:table_name] || name_space + "s" %}
     {% primary_name = PRIMARY[:name] %}
     {% primary_type = PRIMARY[:type] %}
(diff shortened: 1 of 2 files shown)


drujensen:
this is awesome!

So you can do something like this?
```
p.set_attributes(params.select { |key| %w(name description tag_id).include? key })
```

src/kemalyst-model.cr
```diff
@@ -98,6 +98,28 @@ class Kemalyst::Model
         {% end %}
         return fields
     end
+    
+    # Cast params and set fields.
+    private def cast_to_field(name, value)
+      case name.to_s
+        {% for n, type in FIELDS %}
+        when "{{ n.id }}"
+          {% if type.id == Int32.id %}
```
drujensen: Can you add Int64?

elorest:
Int64 done.

drujensen:
released v0.5.6

elorest:
Why do we use `will/crystal-pg` instead of the default crystal one?

drujensen:
did it move to crystal-lang?

elorest:
Nevermind. I guess it didn't. I wonder why not.

PR #33, opened 2017-05-04
new and create methods also work for casting to fields.

elorest:
* Post.create
  accepts hash or namedtuple and returns a saved post model object.
* Post.new
  accepts hash or namedtuple and returns unsaved object.

### DIFF
--- a/src/kemalyst-model.cr
+++ b/src/kemalyst-model.cr
@@ -245,7 +245,7 @@ class Kemalyst::Model
     set_attributes(args.to_h)
   end
 
-  def initialize(**args)
+  def initialize(**args : Object)
     set_attributes(args.to_h)
   end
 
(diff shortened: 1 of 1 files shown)


src/kemalyst-model.cr
```diff
@@ -235,7 +235,7 @@ class Kemalyst::Model
     end
   end # End of Fields Macro
 
-  def set_attributes(args : Hash(Symbol | String, String))
+  def set_attributes(args : Hash(Symbol | String, String | Int32 | Float32))
```
drujensen: What do you need the Int32 and Float32 for?

drujensen:
released v0.5.7

src/kemalyst-model.cr
elorest: All three of those have `to_i, to_f, and to_s`. I added them to make it easier to pass params to them. In some cases you might want to pass and id or count or some other number as a hard coded value. It will make it a lot easier if you don't have to wrap them in quotes. Turns out this will cause the Time parse to mess up but so will strings without the correct formatting. We can easily just tell time to convert it to string first but we'll also have to do checks to make sure it's a parseable string. We should probably add some sort of working time fields to the generator too.

PR #34, opened 2017-05-05
Fixed a couple issues with Bool and Time

elorest:
Checked time string with regex to make sure that it contains a parseable string. Bool and Times can be passed as strings or Bool | Time.


drujensen:
released v0.5.9

PR #39, opened 2017-07-06
Dj/keep db open

drujensen:
This change avoids opening and closing the db connection for every request.  This had a significant performance increase in the TechEmpower benchmarks.

We will probably want to add error handling if the connection fails but I will leave that for a later change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: crystal
+language: generic
 
 services:
   - docker
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,8 +11,6 @@ services:
     depends_on:
       - pg
       - mysql
-    volumes:
-      - '.:/app/user'
   pg:
     image: postgres:9.5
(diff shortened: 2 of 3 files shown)


drujensen:
@elorest yeah, we need to add error handling with retry logic if the connection fails.

drujensen:
@eliasjpr I haven't tried this but I think you can just wrap with the `db.transaction` 

https://github.com/crystal-lang/crystal-db/pull/27

elorest:
IMHO, an improvement shouldn't make an existing system less reliable. This is great because is speeds things up, but I think we should make sure it reconnects if the connection dies before we release it.

drujensen:
Ok. I won't have time to address this right now. I will close this and open a ticket.

drujensen:
@elorest I tested by shutting down the database and restarting it.  The db driver's connection pool takes care of the reconnect logic.

elorest:
Sounds good. :)

PR #45, opened 2017-08-14
is/fixes issue with Time fields other than created_at and updated_at

elorest:
Fixes this issue. https://github.com/amber-crystal/granite-orm/issues/44

### DIFF
--- a/src/granite_orm.cr
+++ b/src/granite_orm.cr
@@ -131,15 +131,19 @@ class Granite::ORM
 
     # keep a hash of the params that will be passed to the adapter.
     def params
-      params = [] of DB::Any
+      parsed_params = [] of DB::Any
       {% for name, type in FIELDS %}
-        params << {{name.id}}
+        {% if type.id == Time.id %}
+          parsed_params << {{name.id}}.try(&.to_s("%F %X"))
+        {% else %}
+          parsed_params << {{name.id}}
+        {% end %}
       {% end %}
       {% if SETTINGS[:timestamps] %}
-        params << created_at.not_nil!.to_s("%F %X")
-        params << updated_at.not_nil!.to_s("%F %X")
+        parsed_params << created_at.not_nil!.to_s("%F %X")
+        parsed_params << updated_at.not_nil!.to_s("%F %X")
       {% end %}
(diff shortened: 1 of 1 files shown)


src/granite_orm.cr
```diff
@@ -131,15 +131,19 @@ class Granite::ORM
...
+          parsed_params << {{name.id}}.try(&.to_s("%F %X"))
+        {% else %}
+          parsed_params << {{name.id}}
+        {% end %}
       {% end %}
       {% if SETTINGS[:timestamps] %}
-        params << created_at.not_nil!.to_s("%F %X")
-        params << updated_at.not_nil!.to_s("%F %X")
+        parsed_params << created_at.not_nil!.to_s("%F %X")
+        parsed_params << updated_at.not_nil!.to_s("%F %X")
```
veelenga: Why we can't disallow nil in `crated_at` and `updated_at` in type definition [here](https://github.com/amber-crystal/granite-orm/blob/c49ff76e5d25cbb8217d57b27469f5b54ad7152e/src/granite_orm.cr#L71-L72) and avoid `not_nil!` runtime check?

elorest:
I submitted a pr which fixes this issue. Could you check it out?

src/granite_orm.cr
drujensen: @veelenga I think this was because sqlite doesn't have support for timestamps.  I will need to verify if this was the reason.

MSDandrea:
Hi everybody, thanks for the quick reply and fix.
I have two questions about this approach:

- By doing this aren't we dropping timezone support for every database adapter?
- Shouldn't it be the adapter responsible for parsing and formatting the params?

thanks.

src/granite_orm.cr
elorest: @veelenga If nil wasn't possible then we couldn't initialize something without setting all values.
```cr
post = Post.new
post.title = "time stuff"
post.save # Time stamps should get set at save not initialize.
```

elorest:
Yes it looks like timezone support is being dropped. I think this is an oversight. Feel free to submit a PR for that. Thanks.

PR #47, opened 2017-08-18
Ecr database.yml

danielwestendorf:
`${}` syntax is unique to this project. Something more ubiquitous
should be used like ECR.

+ Evaluate database.yml as ECR before parsing as YAML
+ Switches the pg database config to use ECR
+ Add notes to README about using ECR, +example

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,3 @@
-export PG_DATABASE_URL=postgres://drujensen@localhost:5432/kemalyst
-export MYSQL_DATABASE_URL=mysql://drujensen@localhost:3306/kemalyst
-export SQLITE_DATABASE_URL=sqlite3:./config/kemalyst.db
+export PG_DATABASE_URL=postgres://granite:password@localhost:5432/granite_db
+export MYSQL_DATABASE_URL=mysql://granite:password@localhost:3306/granite_db
+export SQLITE_DATABASE_URL=sqlite3:./config/granite.db
--- a/README.md
+++ b/README.md
@@ -31,13 +31,13 @@ dependencies:
 ```
 
 Next you will need to create a `config/database.yml`
-You can leverage environment variables using ${} syntax.
+You can leverage environment variables using ${} syntax or ECR.
 
 ```yaml
 mysql:
   database: "mysql://user:pass@mysql:3306/test"
 pg:
-  database: "postgres://postgres:@pg:5432/postgres"
+  database: "<%= ENV["PG_DATABASE_URL"] %>"
 sqlite:
   database: "sqlite3:./config/test.db"
 ```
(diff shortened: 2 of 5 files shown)


danielwestendorf:
Additionally, adding a deprecation notice for the `${}` syntax may be in order?

danielwestendorf:
Would love to alleviate this problem with just documentation and nix this PR.

This is why I created this PR:

I created my first Amber app. I generated some scaffolding and created a few records. Then I ran `crystal spec`, only to find that it was using my development db for tests.

In Rails, I'm used to tests running against a test database. I don't see a way to manage environment dependent databases, so I had planned to do something like:

```yaml
# config/database.yml
pg:
  database: "my_database_<%= ENV["AMBER_ENV"] %>"
```

In production, I would just use a `DATABASE_URL`.

Perhaps this simply isn't needed with Amber? If so, I'd love to document it, I just need some guidance.

drujensen:
@danielwestendorf 

currently, you can do:
```
pg:
  database: "my_database_${AMBER_ENV}"
```
This will pull the AMBER_ENV environment variable.  This is the syntax used by Docker and Docker Compose.

If we document this and change the templates in `amber_cmd`, will this satisfy your concerns?

danielwestendorf:
@drujensen 

Yes, I think it would satisfy my needs from a technical standpoint.

I don't use Docker and Docker Compose. Is that a requirement to use the framework? Why not both methods?

I spent about 30 minutes Googling `crystal-lang ${}` coming up with no results until I finally dug into the source to find it was a self-implemented convention. More documentation would have hopefully helped me figure it out sooner, but I'd personally prefer a convention already available in Crystal (humble opinion of language convention over implementation configuration).

drujensen:
@danielwestendorf thanks for the feedback.

I wrote this `${}` a couple years ago.   I was learning docker and really liked the syntax.  I can see if someone is not familiar with it, it may look odd.

Regarding documentation, yes, that will always be a battle with new frameworks.  A lot of work is still experimental and you don't want to document until you feel confident it's going to stick.

drujensen:
Unfortunately, the yaml standard does not define how to interpolate variables.

After searching for "yaml environment variables" and "yaml interpolation", there are quite a few frameworks using the `${}` syntax:

https://docs.docker.com/compose/environment-variables/
https://docs.puppet.com/hiera/3.3/variables.html
https://www.elastic.co/guide/en/beats/filebeat/1.2/using-environ-vars.html
https://discuss.circleci.com/t/cant-access-env-variables-from-my-yml-file/5463

drujensen:
@danielwestendorf I opened #48 to address your concerns and we will update amber-cmd to add the environment to the database configuration.

At this time, we don't feel that processing the database.yml with ECR is the right way of handling this.  If we find that we need more than just environment variables, we may consider re-opening this PR.

Again, thanks for the contribution.

Issue #30, opened 2017-04-29
add support for 1 to many relationships


drujensen:
Create two new macros: `belongs_to` and `has_many`

```crystal
require "./post"

class Comment < Kemalyst::Model
  belongs_to post
  ...
end
```
This will create a mapping to the `post_id : Int64` field and generate a method that will query the parent and return it.

```
def post
  Post.find post_id
end
```

```crystal
require "./comment"

class Post < Kemalyst::Model
  has_many comments
  ...
end
```
This will create a method called comments that will query the children records.
```
def comments
  Comments.all("JOIN posts ON posts.id = comments.post_id WHERE posts.id = ?", id)
end
```
Options to override the field names and types should be considered for legacy database mappings.

One of the difficulties is requiring the other model.  Macros do not allow you to inject `require` into the class.

Another issue is the N+1 queries for comments.  I think we can find ways of optimizing this later.

elorest:
I would suggest that all of the models should be loaded into the main project file.

```cr
#src/blog.cr
require "./models/*"
```

drujensen:
can you require all subdirectories as well?  does this work in crystal? `require "./models/**/*`

drujensen:
found it:  https://crystal-lang.org/docs/syntax_and_semantics/requiring_files.html
```crystal
require "./models/**"
```

elorest:
Both of those will work. For instance `require "./**/*"`

azah:
I'll try and take a stab at this in the next few weeks.

migeorge:
Closing since #52 was merged.

PR #52, opened 2017-09-08
One to many relationships

migeorge:
The goal of this PR is to add the `belongs_to` and `has_many` macros to Granite ORM Models to allow easier access to related data.

This addresses #30 

To Do:
- [x] `belongs_to`
- [x] `belongs_to` adapter tests
- [x] `belongs_to` setter
- [x] `has_many`
- [x] `has_many` adapter tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ The where clause will give you full control over your query.
 When using the `all` method, the SQL selected fields will always match the
 fields specified in the model.
 
-Always pass in parameters to avoid SQL Injection.  Use a `?` (or `$1`, `$2`,.. for pg)
+Always pass in parameters to avoid SQL Injection.  Use a `?`
 in your query as placeholder. Checkout the [Crystal DB Driver](https://github.com/crystal-lang/crystal-db)
 for documentation of the drivers.
 
--- a/src/adapter/pg.cr
+++ b/src/adapter/pg.cr
@@ -14,6 +14,8 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
   # configured using the sql_mapping directive in your model.  The clause and
   # params is the query and params that is passed in via .all() method
   def select(table_name, fields, clause = "", params = nil, &block)
+    clause = _ensure_clause_template(clause)
+
     statement = String.build do |stmt|
       stmt << "SELECT "
       stmt << fields.map { |name| "#{table_name}.#{name}" }.join(",")
@@ -76,5 +78,17 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
     open do |db|
       db.exec "DELETE FROM #{table_name} WHERE #{primary_name}=$1", value
     end
+  end
+
+  private def _ensure_clause_template(clause)
+    if clause.includes?("?")
+      num_subs = clause.count("?")
+
+      num_subs.times do |i|
+        clause = clause.sub("?", "$#{i+1}")
+      end
+    end
+
+    clause
   end
 end
--- a/src/granite_orm.cr
+++ b/src/granite_orm.cr
@@ -53,9 +53,11 @@ class Granite::ORM
     field {{model_name.id}}_id : Int64
 
     def {{model_name.id}}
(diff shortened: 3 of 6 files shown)


src/granite_orm.cr
```diff
@@ -48,6 +48,31 @@ class Granite::ORM
     {% SETTINGS[:timestamps] = true %}
   end
 
+  # retrive the parent relationship
+  macro belongs_to(model_name)
+    {% FIELDS["#{model_name.id}_id".id.symbolize] = Int64 %}
```
c910335: `field {{model_name.id}}_id : Int64`?

src/granite_orm.cr
```diff
@@ -48,6 +48,31 @@ class Granite::ORM
...
+
+    def {{model_name.id}}
+      parent = {{model_name.id.camelcase}}.find {{model_name.id}}_id
+      return {{model_name.id.camelcase}}.new unless parent
+      parent
```
c910335: I prefer
```cr
if parent = {{model_name.id.camelcase}}.find {{model_name.id}}_id
  parent
else
  {{model_name.id.camelcase}}.new
end
```

README.md
```diff
@@ -179,7 +179,7 @@ The where clause will give you full control over your query.
 When using the `all` method, the SQL selected fields will always match the
 fields specified in the model.
 
-Always pass in parameters to avoid SQL Injection.  Use a `?` (or `$1`, `$2`,.. for pg)
```
drujensen: any reason to remove this?

src/adapter/pg.cr
```diff
@@ -77,4 +79,16 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
       db.exec "DELETE FROM #{table_name} WHERE #{primary_name}=$1", value
     end
   end
+
+  private def _ensure_clause_template(clause)
```
drujensen: interesting, i like this!

src/granite_orm.cr
drujensen: Can you provide a setter as well?
```
user.post = post
```
This should set the `post_id` to `post.id`

src/adapter/pg.cr
migeorge: I do realize that this would make some functions of postgres inoperable, but we could also add an escape if you actually want a literal question mark in your query.

Doing a bit of digging last night I found that Laravel (php framework) does something similar and they just don't support question marks in postgres queries.

Just wanted to bring some light to this.

src/granite_orm.cr
migeorge: Ah, good point. I seem to have gotten lost in macro land :P

src/granite_orm.cr
migeorge: @c910335 I will change it to your preferred format
@drujensen Will do!

src/adapter/pg.cr
drujensen: I would hate to lose postgres functionality if we can avoid it.  However, I'm not sure we want to go down the escaping rabbit hole either.

Let's do this for now and if we see complaints regarding postgres and ?, we can evaluate a solution then.

migeorge: I'm having the same thoughts, sounds like this will be fine for now.

migeorge:
Unless you guys find anything else this implementation is complete on my end

src/granite_orm.cr
```diff
@@ -48,6 +48,39 @@ class Granite::ORM
...
+    # set the parent relationship
+    def {{model_name.id}}=(parent)
+      @{{model_name.id}}_id = parent.id
+    end
+  end
+
+  # define getter for related children
+  macro has_many(children_table)
+    def {{children_table.id}}
+      {% children_class = children_table.id[0...-1].camelcase %}
```
migeorge: This seems like a very naive way to convert from plural to singular (via removing the last character, presumably an s). Crystal doesn't seem to have an inflection library yet, which seems like something we don't want to implement in an ORM. So maybe this is okay for now?

PR #62, opened 2017-09-16
add convenience method - first

drujensen:
This adds a new query method `first` that adds a `LIMIT 1` to the query and returns a single object instead of an array.

This also allows you to pass a SQL clause just like the `all` method to perform the query.

For example:
```
first_post = Post.first("WHERE posts.active = true 
                         ORDER BY posts.updated_at DESC")
```


src/granite_orm/querying.cr
```diff
@@ -45,6 +45,16 @@ module Granite::ORM::Querying
...
   end
 
+  # First adds a `LIMIT 1` clause to the query and returns the first result
+  def first(clause = "", params = [] of DB::Any)
+    results = all([clause.strip, "LIMIT 1"].join(" "), params)
+    if results.size > 0
+      results.first
+    else
+      nil
+    end
```
c910335: How about `results[0]?`?

migeorge: Is this a stylistic choice or is there a performance difference? Just curious

c910335: We don't need to check size here because `#[]?` did it.

c910335: Oh I just found a better choice which is `#.first?`.

src/granite_orm/querying.cr
```diff
@@ -45,6 +45,16 @@ module Granite::ORM::Querying
     return rows
   end
 
+  # First adds a `LIMIT 1` clause to the query and returns the first result
+  def first(clause = "", params = [] of DB::Any)
+    results = all([clause.strip, "LIMIT 1"].join(" "), params)
+    if results.size > 0
+      results.first
+    else
+      nil
```
eliasjpr: Do we need explicitly to return `nil`?

src/granite_orm/querying.cr
drujensen: good call, I will use that

src/granite_orm/querying.cr
drujensen: @c910335 recommended `#first?` which really cleans this up.

src/granite_orm/querying.cr
```diff
@@ -45,6 +45,11 @@ module Granite::ORM::Querying
     return rows
   end
 
+  # First adds a `LIMIT 1` clause to the query and returns the first result
+  def first(clause = "", params = [] of DB::Any)
+    results = all([clause.strip, "LIMIT 1"].join(" "), params).first?
```
c910335: Now `results` is unnecessary.

drujensen: right.  thx.

Issue #59, opened 2017-09-14
Improving relationships between models 


villzor:
When using Granite-Orm I found myself missing a lot of functionalities in terms of relationships between models, and after starting extending it in my own chamber I wanted to share some thoughts on improvement, my approach so far and possibilities to contribute this to Granite. I'm still a beginner in Crystal so I'm sorry if some of my code is written poorly!

## Distinguishing between serial and primary
By distinquishing between serial (an autoincrementing field in DB) from primary(a primary key, not incrementing), it opens up better possibilities to start mapping many-to-many relationships through joining tables, in the same manner as before, and could be added as: 
`macro serial(col)
    {% SERIAL[col.var] = col.type %}
  end`

## Identiying the most important database relationships to cover
For Granite to mirror the different kind of relationships found in databases, I created the following cases and cover them one by one below:

1. many
2. many_through
3. many_tuples_through
4. one 
5. one_tuple

I've been working on a website for creating recipes with object references for ingredients, so I will use my recipe structure to exemplify the different cases.

### 1. many
This is what is called "has_many" in Granite today, a one-to-many relationship. Basically, a recipe has many steps, and each step has a primary key which links it to the specific recipe.
```
class Recipe < Model
  serial recipe_id : Int32
  field title : String
  include_timestamps
  has_many steps : Step
end

class Step < Model
  primary recipe_id : Int32
  primary step_order : Int32
  field description : String
  include_timestamps
end
```

The steps are lazily loaded within the Model, such as:
```
{% for name, type in MANY %}
      property {{name.id}} : Array({{type.id}}) { {{type.id}}.where({ 
      {% for name, type in SERIAL %}
            :{{name.id}} => @{{name.id}}, {% end %}
      {% for name, type in PRIMARY %} ::{{name.id}} => @{{name.id}}, {% end %} }) }
{% end %}
```
### 2. many_through
This is the standard many-to-many case, where two tables are joined through a joining table. Besides poor naming of my functions, here is a sketch how the adapter part of the function would behave:
```
def many_through(parent, child, joiner)
    statement = String.build do |stmt|
      stmt << "SELECT "
      stmt << child.property_names.map { |key| 
          "#{child.table}.#{key}" 
        }.join(", ") 
      stmt << " FROM "
      stmt << joiner.table
        stmt << " INNER JOIN " + child.table + " ON "
        stmt << child.keys.map { |key| 
          "#{child.table}.#{key} = #{joiner.table}.#{key}"
[...]

PR #63, opened 2017-09-17
add support for has_many through

drujensen:
This addresses #59 and adds the ability to access a many-to-many relationship using a `through` option to the `has_many` macro.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,86 @@ CREATE TABLE posts (
 CREATE INDEX 'user_id_idx' ON TABLE posts (user_id);
 ```
 
+### Many to Many Relationships
+
+Instead of using a hidden many to many table, Granite recommends always creating a model for your join tables.  For example, lets say you have many `users` that belong to many `rooms`. We recommend adding a new model called `participants` to represent the many-to-many relationship.
+
+Then you can use the `belongs_to` and `has_many` relationships going both ways.
+```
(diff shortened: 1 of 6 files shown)


drujensen:
@migeorge I added tests for pg and sqlite.

drujensen:
@elorest After several comments and feedback, it's clear that we need to provide at least `belongs_to` and `has_many`.   I wanted to draw a line at DSL for SQL, but I can see why the slippery slope happens.  The abstraction is natural since you can remove the complexity of some queries with a convenience method.  Having a `has_many :users` is much cleaner than `def users ....join query... end`.

I'm not sure where to draw the line but probably with DSL that tries to generate SQL clause.
```
post.all("WHERE users.age = ?", 10)
```
vs
```
post.where(:age == 10)
```
or variant DSL.

eliasjpr:
BTW I am all for named parameters ':age' vs position params `?`

drujensen:
@eliasjpr yeah, that would be nice.  We might be able to abstract the differences from the drivers.  

@c910335 recently encapsulated the difference of the postgres using positional parameters `$1, $2, ...` to use '?' instead.  It actually supports both, which is nice.  

Adding support for named parameters for all the drivers would be really nice.

c910335:
That was @migeorge's contribution, not mine.

Issue #50, opened 2017-09-07
Add support for to_json to models


migeorge:
When developing an API only application with Amber the need for JSON model serialization becomes apparent rather quickly.

It seems logical to add JSON mapping support to Granite models so I wanted to open this issue to start a discussion.

I'm willing to put some time towards a PR on this, but I wanted to get input from existing members of this project to flesh out an acceptable implementation before putting the effort into actually developing the solution.

Questions:

1. Should Granite::ORM implement a to_json method on models similar to the `JSON.mapping` macro?
2. Should Granite::ORM implement a JSON -> Model interface similar to the `JSON.mapping` macro?
3. Is there a good way to inherit some of this functionality (from Crystal core) without re-implementing it?

drujensen:
Hi Mike.

Thanks for opening the discussion. 

I have put a lot of thought into this over the years and have tried implementing a clean JSON mapping in granite ORM to no avail. 

The current JSON.mapping in the core requires an initializer that conflicts the granite initilizer. 

Also, To properly serialize to JSON, most users will need to map to customized key names.  In the JS world, they will want the keys to be camelCased. 

I have leaned toward writing Serializers in my own projects that take the model as an init parameter and perform the mapping with a hash. This is a great place to add other methods that can handle other mapping logic. 

However, if we did pursue adding a to_json mapping, One of the best options I have attempted is allowing for more parameters to be specified in the mapping including a key name for the JSON mapping. 



...

migeorge:
That makes sense for a simple API it might make sense to stick with `to_json` so you can return something like the following in your controller:

```
{ users: User.all }.to_json
```

But I do also see the case for Serializers. Often times in rails projects I find myself reaching for Active Model Serializers, especially once you get into side-loading relationships. Maybe Granite models could support `to_json` and we also add an optional serialization library to the Amber ecosystem. If the models support `to_json` it would probably also simplify the implementation of the serialization library.

I also like the implementation proposal you have. It would also be nice for it to infer the type of the field so the explicit configuration of the field could be optional.

As far as the timestamps go, it may be sane to exclude them from the json payload by default and add an argument to the macro to toggle settings like `{ json: { include: true, camelCase: true } }`

elorest:
My thoughts are that we should just return a Hash(Symbol, String) we could call to_json on it or pass it to another serializer. That way we easily allow for extendability without having to build a serializer in or really changing the system at all. Thoughts?

drujensen:
[...]

PR #72, opened 2017-10-17
Add to_h and to_json to Granite::ORM::Fields

migeorge:
This PR should satisfy the agreed upon resolution of #50 

It adds a `to_h` and `to_json` method to Granite ORM Models.

### DIFF
--- a/spec/granite_orm_spec.cr
+++ b/spec/granite_orm_spec.cr
@@ -16,14 +16,14 @@ describe Granite::ORM::Base do
 
   it "should provide a to_h method" do
     t = Todo.new(name: "test todo", priority: 20)
-    result = { "name" => "test todo", "priority": 20 }
+    result = { "name" => "test todo", "priority" => 20, "created_at" => "", "updated_at" => "" }
 
     t.to_h.should eq result
   end
 
   it "should provide a to_json method" do
     t = Todo.new(name: "test todo", priority: 20)
-    result = "{\"name\":\"test todo\",\"priority\":20}"
+    result = %({"name":"test todo","priority":20,"created_at":"","updated_at":""})
 
     t.to_json.should eq result
   end
--- a/spec/granite_orm_spec.cr
+++ b/spec/granite_orm_spec.cr
@@ -4,6 +4,7 @@ require "../src/adapter/pg"
 class Todo < Granite::ORM::Base
   adapter pg
   field name : String
+  field priority : Int32
   timestamps
 end
 
@@ -11,5 +12,19 @@ describe Granite::ORM::Base do
   it "should create a new todo object with name set" do
     t = Todo.new(name: "Elorest")
     t.name.should eq "Elorest"
+  end
+
+  it "should provide a to_h method" do
+    t = Todo.new(name: "test todo", priority: 20)
+    result = { "name" => "test todo", "priority": 20 }
+
+    t.to_h.should eq result
+  end
+
+  it "should provide a to_json method" do
+    t = Todo.new(name: "test todo", priority: 20)
+    result = "{\"name\":\"test todo\",\"priority\":20}"
+
+    t.to_json.should eq result
   end
 end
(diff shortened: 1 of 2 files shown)


migeorge:
Let me know if you guys see any areas to improve the `to_h` method implementation. I would have used `DB::Any` as the type for the hash values but I had to exclude `Slice(UInt8)` in order for the `to_json` call to work.

I also didn't want to cast all the values to `String` since numeric values and booleans are perfectly valid in JSON.

spec/granite_orm_spec.cr
```diff
@@ -12,4 +13,18 @@ describe Granite::ORM::Base do
     t = Todo.new(name: "Elorest")
     t.name.should eq "Elorest"
   end
+
+  it "should provide a to_h method" do
+    t = Todo.new(name: "test todo", priority: 20)
+    result = { "name" => "test todo", "priority": 20 }
```
veelenga: This is interesting:

```crystal
{ "name" => "test todo", "priority": 20 }.class # => Hash(String, Int32 | String)
{ "priority": 20 }.class # => NamedTuple(priority: Int32)
{ "priority": 20, "name" => "test todo" }.class # space not allowed between named argument name and ':'
```

I don't want to be pedantic, but the first case looks like a bug in Crystal or at least it is not documented. I would prefer to use here official way for hashes to avoid problems in future:

```crystal
{ "name" => "test todo", "priority" => 20 }.class # => Hash(String, Int32 | String)
```

src/granite_orm/fields.cr
```diff
@@ -54,6 +54,35 @@ module Granite::ORM::Fields
...
+        {% end %}
+      {% end %}
+      {% if SETTINGS[:timestamps] %}
+        if created_at
+          fields["created_at"] = created_at.not_nil!.to_s("%F %X")
+        end
+
+        if updated_at
+          fields["updated_at"] = updated_at.not_nil!.to_s("%F %X")
+        end
```
veelenga: All fields with `nil` values will be present in a hash, but `created_at` and `updated_at` will not. In this intended?

spec/granite_orm_spec.cr
migeorge: @veelenga My ruby background seems to have seeped in here pushing a fix shortly.

src/granite_orm/fields.cr
migeorge: @veelenga Good point I'll allow the nil values

spec/granite_orm_spec.cr
```diff
@@ -12,4 +13,18 @@ describe Granite::ORM::Base do
...
+
+    t.to_h.should eq result
+  end
+
+  it "should provide a to_json method" do
+    t = Todo.new(name: "test todo", priority: 20)
+    result = "{\"name\":\"test todo\",\"priority\":20}"
```
c910335: How about `%({"name":"test todo","priority":20})`?

migeorge: I like this better than escaping all the quotes

migeorge:
I've made the changes on my branch in migeorge/granite-orm but they don't seem to be propagating here. Anyone experienced this before?

src/granite_orm/fields.cr
```diff
@@ -54,6 +54,30 @@ module Granite::ORM::Fields
...
+          fields["{{name}}"] = {{name.id}}.try(&.to_s("%F %X"))
+        {% elsif type.id == Slice.id %}
+          fields["{{name}}"] = {{name.id}}.try(&.to_s(""))
+        {% else %}
+          fields["{{name}}"] = {{name.id}}
+        {% end %}
+      {% end %}
+      {% if SETTINGS[:timestamps] %}
+        fields["created_at"] = created_at ? created_at.not_nil!.to_s("%F %X") : ""
+        fields["updated_at"] = updated_at ? updated_at.not_nil!.to_s("%F %X") : ""
```
veelenga: 👍 

but why `""` instead of `nil` ?

```crystal
fields["created_at"] = created_at.try &.to_s("%F %X")
fields["updated_at"] = updated_at.try &.to_s("%F %X")
```

migeorge: @veelenga That makes more sense, fixed.

migeorge:
This should be ready to go unless anyone has more feedback

PR #73, opened 2017-10-17
Readme Testing Instructions Fix

migeorge:
I found what appears to be a typo in the instructions while setting up my testing environment.

The instructions say to `$ export .env` where you'd actually use the `source` command to process the file with bash.


drujensen:
@elorest The .env is used by the `docker-compose.yml` here: https://github.com/amberframework/granite-orm/blob/master/.env


We should probably add information about how you can test using docker-compose`

```bash
dc run spec
```

PR #75, opened 2017-10-17
Use JSON::Builder instead of to_h.to_json

veelenga:
This change ships two improvements:

1. Build json with [JSON::Builder](https://crystal-lang.org/api/0.23.1/JSON.html#build%28io%3AIO%2Cindent%3Dnil%2C%26block%29-class-method) is faster than `to_h.to_json` (because of transitional hash of course):

```crystal
# ../granite-orm/example.cr

require "benchmark"
require "./src/granite_orm"
require "./src/adapter/sqlite"

class Comment < Granite::ORM::Base
  adapter sqlite
  table_name post_comments
  field name : String
  field body : String
end

s = "string"
c = Comment.new(name: s, body: s)

Benchmark.ips do |x|
  x.report("to_h.to_json") do
    c.to_h.to_json
  end

  x.report("JSON::Builder") do
    c.to_json
  end
end
```

```sh
$ DB_NAME=test.db crystal run example.cr --release                                                                                                               
to_h.to_json 814.53k (  1.23µs) (± 0.98%)  1.49× slower
JSON::Builder   1.22M (822.64ns) (± 1.29%)       fastest
```

2. Allow `to_json` on a collection of model objects, i.e. `user.posts.to_json` should work now. Before there was a compilation error.

### DIFF
--- a/spec/granite_orm_spec.cr
+++ b/spec/granite_orm_spec.cr
@@ -16,15 +16,30 @@ describe Granite::ORM::Base do
 
   it "should provide a to_h method" do
     t = Todo.new(name: "test todo", priority: 20)
-    result = { "name" => "test todo", "priority" => 20, "created_at" => nil, "updated_at" => nil }
+    result = {"name" => "test todo", "priority" => 20, "created_at" => nil, "updated_at" => nil}
 
     t.to_h.should eq result
   end
 
-  it "should provide a to_json method" do
-    t = Todo.new(name: "test todo", priority: 20)
-    result = %({"name":"test todo","priority":20,"created_at":null,"updated_at":null})
+  describe "#to_json" do
+    it "converts object to json" do
+      t = Todo.new(name: "test todo", priority: 20)
+      result = %({"name":"test todo","priority":20,"created_at":null,"updated_at":null})
 
-    t.to_json.should eq result
+      t.to_json.should eq result
+    end
+
+    it "works with collections" do
+      todos = [
+        Todo.new(name: "todo 1", priority: 1),
+        Todo.new(name: "todo 2", priority: 2),
+        Todo.new(name: "todo 3", priority: 3),
+      ]
+
+      collection = JSON.parse todos.to_json
+      collection[0].should eq({"name" => "todo 1", "priority" => 1, "created_at" => nil, "updated_at" => nil})
+      collection[1].should eq({"name" => "todo 2", "priority" => 2, "created_at" => nil, "updated_at" => nil})
+      collection[2].should eq({"name" => "todo 3", "priority" => 3, "created_at" => nil, "updated_at" => nil})
+    end
   end
 end
--- a/src/granite_orm/fields.cr
+++ b/src/granite_orm/fields.cr
@@ -1,3 +1,5 @@
+require "json"
+
 module Granite::ORM::Fields
   macro included
     macro inherited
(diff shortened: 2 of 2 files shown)


elorest:
Pretty sure the old version would have worked with collections in the same way. JSON.parse collection is just calling to_json on every element of the collection. This is just a tiny bit faster.

veelenga:
> @migeorge Pretty sure the old version would have worked with collections in the same way. JSON.parse collection is just calling to_json on every element of the collection. This is just a tiny bit faster.

Unfortunately, it wouldn't. [`Object#to_json`](https://github.com/crystal-lang/crystal/blob/b52b2f93dd037a12f33354b05fe26a61e1b33662/src/json/to_json.cr#L2-L6) calls [`Array#to_json(json : JSON::Builder)`](https://github.com/crystal-lang/crystal/blob/b52b2f93dd037a12f33354b05fe26a61e1b33662/src/json/to_json.cr#L63-L69), which forces to have `#to_json(json : JSON::Builder)` method instead of a plain `#to_json`. Result:

```crystal
c = Comment.new(name: "name", body: "body")
[c, c].to_json

# no overload matches 'Comment#to_json' with type JSON::Builder
# Overloads are:
# - Comment#to_json()
# - Object#to_json(io : IO)
# - Object#to_json()
#
#      each &.to_json(json)

```

@elorest Did I miss something? Have you tried that on master?

veelenga:
> In most cases we'll be modifying the results of to_h in an active model serializer like way. However this is a bit faster to I'm fine with merging it. Thanks.

This is not a Crystal way to do things. Instead of allocating the result in memory, Crystal is about to providing equivalent methods in the IO itself to send the data directly to the socket, file, stream, etc... I don't have an alternative design for this, but I think it may be a big performance impact...

drujensen:
@eliasjpr oops, just saw you had a question after I merged.  Are you referring to a `has_many` relationship?

If so, we would recommend a custom serializer for that use case.

veelenga:
@eliasjpr here is an example:

```crystal
require "./src/granite_orm"
require "./src/adapter/mysql"

  adapter mysql

  has_many :posts

  field email : String
  field name : String
  timestamps
end

class Post < Granite::ORM::Base
  adapter mysql

  belongs_to :user

  field title : String
  timestamps
end

u = User.new(id: 1, email: "user1@mail.com", name: "Hero").tap &.save
Post.new(title: "Post1", user_id: u.id).tap &.save
Post.new(title: "Post2", user_id: u.id).tap &.save
Post.new(title: "Post3", user_id: u.id).tap &.save

puts u.to_json # {"email":"user1@mail.com","name":"Hero","created_at":"2017-10-18 18:04:09","updated_at":"2017-10-18 18:04:09"}

puts Post.all.to_json # [{"user_id":1,"title":"Post1","created_at":"2017-10-18 18:04:09","updated_at":"2017-10-18 18:04:09"},{"user_id":1,"title":"Post2","created_at":"2017-10-18 18:04:09","updated_at":"2017-10-18 18:04:09"},{"user_id":1,"title":"Post3","created_at":"2017-10-18 18:04:09","updated_at":"2017-10-18 18:04:09"}]
puts u.posts.to_json  # [{"user_id":1,"title":"Post1","created_at":"2017-10-18 18:04:09","updated_at":"2017-10-18 18:04:09"},{"user_id":1,"title":"Post2","created_at":"2017-10-18 18:04:09","updated_at":"2017-10-18 18:04:09"},{"user_id":1,"title":"Post3","created_at":"2017-10-18 18:04:09","updated_at":"2017-10-18 18:04:09"}]

```

PR #78, opened 2017-10-21
Adds  Support For Amber

eliasjpr:
As a developer I want to be able to initialize a Granite Model from
Amber params

See PR https://github.com/amberframework/amber/pull/315

- Adds a Type alias
- Type includes  Hash(String, JSON::Type) | Array(JSON::Type)
- Adds Fields::Type to Base.cr initializer

With this change developers will be able to populate Model objects from
Amber params.validation results.

### DIFF
--- a/src/granite_orm/fields.cr
+++ b/src/granite_orm/fields.cr
@@ -1,6 +1,8 @@
 require "json"
 
 module Granite::ORM::Fields
+  alias Type = DB::Any | Hash(String, JSON::Type) | Array(JSON::Type)
+
   macro included
     macro inherited
       FIELDS = {} of Nil => Nil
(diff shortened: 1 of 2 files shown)


src/granite_orm/fields.cr
```diff
@@ -1,6 +1,8 @@
 require "json"
 
 module Granite::ORM::Fields
+  alias Type = Bool | Float32 | Float64 | Int32 | Int64 | String | Time | Nil | Hash(String, JSON::Type) | Array(JSON::Type)
+
```
drujensen: Here are the DB::Types: https://github.com/crystal-lang/crystal-db/blob/master/src/db.cr#L72

[Nil, String, Bool, Int32, Int64, Float32, Float64, Time, Bytes]

If we add `Bytes`, we can remove the `DB::Any` or if we leave the DB::Any in, we can remove `Bool | Float32 | Float64 | Int32 | Int64 | String | Time | Nil`

eliasjpr: Type alias looks like this now `alias Type = DB::Any | Hash(String, JSON::Type) | Array(JSON::Type)`

PR #82, opened 2017-10-22
Unboxes Union Values

eliasjpr:
Since we are now supporting JSON sometimes value types will be of JSON::Any
this adds the `as(type)` to ensure the value is treated as the expected value.


eliasjpr:
WIP - Fixing specs.

faustinoaq:
Hi @eliasjpr Then, Why `.to_...` is still needed?

PR #89, opened 2017-11-08
refactor casting to support multiple types

drujensen:
This change merges the `cast_json_to_fields` and `cast_to_fields`.  It handles different scenarios including params being passed in as String or JSON::Type.  It also supports setting the types as their original declared type avoiding casting when not necessary.

This should allow for more flexibility to users so they can pass in what they have and as long as it can be converted, this will do its best to do so.

This should be backward compatible when someone passes in a Hash(String, String).

All existing specs pass.  Let me know if you think we should add more specific specs.

### DIFF
--- a/src/granite_orm/fields.cr
+++ b/src/granite_orm/fields.cr
@@ -1,7 +1,7 @@
 require "json"
 
 module Granite::ORM::Fields
-  alias Type = DB::Any | JSON::Any
+  alias Type = JSON::Type | DB::Any
   TIME_FORMAT_REGEX = /\d{4,}-\d{2,}-\d{2,}\s\d{2,}:\d{2,}:\d{2,}/
 
   macro included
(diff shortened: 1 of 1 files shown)


marksiemers:
This lgtm, but I don't have enough background to feel comfortable approving.

src/granite_orm/fields.cr
```diff
@@ -103,93 +103,42 @@ module Granite::ORM::Fields
...
+      case name.to_s
+        {% for _name, type in FIELDS %}
+        when "{{_name.id}}"
+          return @{{_name.id}} = nil if value.nil?
+          {% if type.id == Int32.id %}
+            @{{_name.id}} = value.is_a?(String) ? value.to_i32 : value.is_a?(Int64) ? value.to_s.to_i32 : value.as(Int32)
+          {% elsif type.id == Int64.id %}
+            @{{_name.id}} = value.is_a?(String) ? value.to_i64 : value.as(Int64)
+          {% elsif type.id == Float32.id %}
+            @{{_name.id}} = value.is_a?(String) ? value.to_f32 : value.is_a?(Float64) ? value.to_s.to_f32 : value.as(Float32)
```
elorest: Parenthesis or some sort of mapping method could make this logic readable...

value.is_a?(String) ? value.to_f32 : (value.is_a?(Float64) ? value.to_s.to_f32 : value.as(Float32))

PR #94, opened 2017-11-13
Dj/support setting database url directly

drujensen:
This allows you to set the database url directly using `Granite::ORM.settings.database_url`. 

If the DATABASE_URL exists, it will use that as highest priority.  If the `settings.database_url` is set directly, it will use that as the second priority.  Finally it will look for a `config/database.yml` and use that if it exists as third priority.

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -8,7 +8,7 @@ abstract class Granite::Adapter::Base
   property database : DB::Database
 
   def initialize(adapter : String)
-    if url = ENV["DATABASE_URL"]? || replace_env_vars(settings(adapter)["database"].to_s)
+    if url = ENV["DATABASE_URL"]? || Granite::ORM.settings.database_url || replace_env_vars(settings(adapter)["database"].to_s)
       @database = DB.open(url)
     else
       raise "database url needs to be set in the config/database.yml or DATABASE_URL environment variable"
--- a/src/granite_orm/querying.cr
+++ b/src/granite_orm/querying.cr
@@ -82,15 +82,4 @@ module Granite::ORM::Querying
   def scalar(clause = "", &block)
     @@adapter.open { |db| yield db.scalar(clause) }
   end
-
-  def create(**args)
-    create(args.to_h)
-  end
-
-  def create(args : Hash(Symbol | String, DB::Any))
-    instance = new
-    instance.set_attributes(args)
-    instance.save
-    instance
-  end
 end
--- a/src/granite_orm/base.cr
+++ b/src/granite_orm/base.cr
@@ -13,7 +13,6 @@ class Granite::ORM::Base
   include Associations
   include Callbacks
   include Fields
-  include Settings
   include Table
   include Transactions
   include Validators
--- a/src/granite_orm/settings.cr
+++ b/src/granite_orm/settings.cr
@@ -1,7 +1,9 @@
-module Granite::ORM::Settings
-  macro included
-    macro inherited
-      SETTINGS = {} of Nil => Nil
-    end
+module Granite::ORM
+  class Settings
+    property database_url : String? = nil
+  end
+
+  def self.settings
+    @@settings ||= Settings.new
   end
 end
--- a/src/granite_orm/table.cr
+++ b/src/granite_orm/table.cr
@@ -8,7 +8,7 @@ module Granite::ORM::Table
   # specify the database adapter you will be using for this model.
   # mysql, pg, sqlite, etc.
   macro adapter(name)
-    @@adapter = Granite::Adapter::{{name.id.capitalize}}.new("{{name.id}}")
+    @@adapter ||= Granite::Adapter::{{name.id.capitalize}}.new("{{name.id}}")
 
     def self.adapter
       @@adapter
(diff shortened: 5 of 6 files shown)


src/granite_orm/querying.cr
```diff
@@ -82,15 +82,4 @@ module Granite::ORM::Querying
   def scalar(clause = "", &block)
     @@adapter.open { |db| yield db.scalar(clause) }
   end
-
-  def create(**args)
```
drujensen: The create method should be in the transactions.cr, not the querying.cr

src/adapter/base.cr
```diff
@@ -8,7 +8,7 @@ abstract class Granite::Adapter::Base
   property database : DB::Database
 
   def initialize(adapter : String)
-    if url = ENV["DATABASE_URL"]? || replace_env_vars(settings(adapter)["database"].to_s)
+    if url = ENV["DATABASE_URL"]? || Granite::ORM.settings.database_url || replace_env_vars(settings(adapter)["database"].to_s)
```
eliasjpr: I would consider removing the `ENV["DATABASE_URL"]`, Reasons:

- Any system using granite can explicitly set the URL via `Granite::ORM.settings.database_url` including `Granite::ORM.settings.database_url = ENV["DATABASE_URL"]`
- If a DATABASE_URL evn variable is define in the system that is should NOT be use by granite it will fail to load the db url set via `Granite::ORM.settings.database_url`. Think multi-tenant apps where there is a DB holding config settings that is shared across apps and then each app has its own individual db.
- With ENV["DATABASE_URL"] Granite knows to much about `where` to look and violates `tell don't ask` principle. Now that we can set database url by composition with `Granite::ORM.settings.database_url` we can tell granite which connection to use. 

`ENV["DATABASE_URL"]` IMHO should be remove.

marksiemers: @eliasjpr - So you are saying that amber (or another app) should make use of `ENV["DATABASE_URL"]`? Rather than directly in granite?

I could see that. If an app wanted to start with `ENV["DATABASE_URL"]` and append options to it, then Granite would be overriding those options.

```
# In amber (or other app)
Granite::ORM.settings.database_url = ENV["DATABASE_URL"] + "?max_connections=#{ENV["CPU_COUNT"]}"
```

elorest: In terms of amber I think that makes sense, but if people will use granite other places then it might be easier to leave it how it is.

marksiemers: What about changing the priority?
```
Granite::ORM.settings.database_url || ENV["DATABASE_URL"]? || replace_env_vars(settings(adapter)["database"].to_s)
```

eliasjpr: It feels like magic that I set an environment variable and Granite just connects to it, unless someone knows this about granite it would be hard to figure it out. With the existence of `Granite::ORM.settings.database_url` I don't see the need of having `ENV["DATABASE_URL"]` anymore, that's all. 

As per @elorest comment if people use granite in other places is the same input they would need `Granite::ORM.settings.database_url = ENV["DATABASE_URL"]` instead of having Granite connecting implicitly to a database.

This is what I think and most likely I could be wrong, it is up to you guys if you think is worth the change.

drujensen: Amethyst-Model, Kemalyst-Model, Granite::ORM is following the Rails database configuration.  http://edgeguides.rubyonrails.org/configuring.html#configuring-a-database

Rails developers are familiar with these locations for configuring their database.  Heroku also uses the `DATABASE_URL` environment variable for easy deployment.  I would prefer keeping these for people that prefer the familiarity of Rails (i.e. me)

We can look at deprecating `config/database.yml` but I would still want to do this in a progressive manner.  I also want to keep the DATABASE_URL environment variable as the override since this is fairly well known and commonly used (by me).

Again, Granite::ORM is being used outside of Amber including old Kemalyst sites.  Let's be kind to these people(me).  If you have a really compelling reason we should create a breaking change, I would be open to it but I don't see a good enough reason at this point.

marksiemers: @drujensen would you consider this breaking?

```
if url = Granite::ORM.settings.database_url || ENV["DATABASE_URL"]? || ...
```

drujensen: No, probably wouldn't break existing apps.  But I would consider it incorrect since the expected behaviour is DATABASE_URL will override your config.  Anyway, that is the way it is in Rails.

marksiemers: Does it mean that this won't be possible?
```
# In amber (or other app)
Granite::ORM.settings.database_url = ENV["DATABASE_URL"] + "?max_connections=#{ENV["CPU_COUNT"]}"
```
If so, is there an easy way to set the connection settings?
https://crystal-lang.org/docs/database/connection_pool.html#configuration

drujensen:
You could do this:
```
ENV["DATABASE_URL"] = ENV["DATABASE_URL"] + "?max_connections=#{ENV["CPU_COUNT"]}"
```
Or just set this in the environment when you launch the app if you have control over it.

drujensen:
@marksiemers @eliasjpr @elorest @faustinoaq let me know your vote on the priority order.  I can see that setting the `settings.database_url` directly may have priority since its a push instead of a pull approach.  I am open to this change.

If you think the Environment should take priority, click the 👍.  If you think the settings.database_url should take priority, click the 👎

Issue #29, opened 2017-04-29
timestamps in sqlite


drujensen:
Support timestamps and time, date in sqlite

damianham:
fix proposed in https://github.com/amberframework/granite-orm/pull/135 (for timestamps)

PR #96, opened 2017-11-17
forces timestamps to be in UTC

### LINKED ISSUE #95: Timestamps should be in UTC, or at least consistent
Today I noticed this discrepancy with the way times are written to and read from the database:

```crystal
def test
  result = Result.new()
  result.save
  insert_time = Time.now

  return unless timestamp = result.created_at
  result = Result.find(result.id)
  return unless result
  return unless timestamp2 = result.created_at
  puts timestamp - insert_time
  puts timestamp2 - insert_time
end
test
```

```
-00:00:00.0027060
-07:00:00.7834670
```

Driver: postgres, but I haven't tested others. I believe the time is going in with a local offset, and being read out assumed in UTC. I'm in MST(-7), so there's a -7 discrepancy.

The fix might be as easy as a `#to_utc` [in a few places here](https://github.com/amberframework/granite-orm/blob/master/src/granite_orm/transactions.cr), but I'm not sure.

robacarp:
fixes #95 

This does not bring the mysql adapter into parity with postgresql because the mysql adapter returns `Time::Kind::Unspecified` instead of `Time::Kind::Utc`, however as the tests illustrate, the ticks behind the timestamp are correct.

Postgresql was the closest to consistent already, so this seems to make postgresql timestamps work as expected.

Sqlite is not addressed because of #29

### DIFF
--- a/src/granite_orm/transactions.cr
+++ b/src/granite_orm/transactions.cr
@@ -3,6 +3,9 @@ module Granite::ORM::Transactions
     {% primary_name = PRIMARY[:name] %}
     {% primary_type = PRIMARY[:type] %}
 
+    @updated_at : Time?
+    @created_at : Time?
+
     # The save method will check to see if the primary exists yet. If it does it
     # will call the update method, otherwise it will call the create method.
     # This will update the timestamps apropriately.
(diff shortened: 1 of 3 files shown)


src/granite_orm/transactions.cr
```diff
@@ -3,6 +3,9 @@ module Granite::ORM::Transactions
     {% primary_name = PRIMARY[:name] %}
     {% primary_type = PRIMARY[:type] %}
 
+    @updated_at : Time?
+    @created_at : Time?
```
robacarp: [`Time.now.utc`](https://crystal-lang.org/api/0.23.1/Time.html#to_utc-instance-method) isn't annotated, so explicit type declarations are necessary.

drujensen:
@amberframework/core-team need another approval. Storing dates without ambiguity of timezone is fairly standard practice.  Should have changed this a while ago.

robacarp:
@drujensen @faustinoaq Thanks! I appreciate your willingness to get collaborators changes in!

Issue #97, opened 2017-11-19
can't name a column created_at in SQLite 


masukomi:
this is related to #29 but can be handled independently

As long as we don't have support for timestamps in SQLite it shouldn't force created_at to be a `Time` format. 

On a related note, if i don't tell granite to use timestamps, it seems silly that it should try and enforce the format of timestamp columns. Yes, it's a bit silly to have "bob" in my `created_at` column, but as I didn't tell granite to do timestamps it shouldn't be enforcing it's opinions on what i can have in that column.

Yes, rails goes ahead and does it, but there's a difference. In rails, I don't have to specify that i want it to use timestamps, which the docs seem to suggest i DO need to do, although this bug seems to be proving the docs wrong.... either that or something is whacky.

In the meantime, i'll be attempting to monkeypatch the created_at functionality out of existence. :/

masukomi:
What follows is my solution, for anyone else encountering this, and also a field guide for devs wanting to clean up the handling of created_at and updated_at which are too smart for their own good and incorrectly setting `SETTINGS[:timestamps] = true` when it wasn't called.

monkeypatch `fields.cr` with this and then call it in your model, because if it sees a `created_at` field it's going to believe that timestamps must be true.
```
module Granite::ORM::Fields
  macro no_timestamps
    {% SETTINGS[:timestamps] = false %}
  end
end
```

then monkeypatch `transactions.cr` 
and alter the `__run_before_save` and `__run_before_create` to stop trying to set updated_at and created at... also get rid of the definitions of those instance variables that sets them to `Time` objects.

then require your new `fields.cr` and `transactions.cr` after you require granite sqlite adapter.

drujensen:
@masukomi thanks for reporting the issue.  

The plan is to support the `created_at` and `updated_at` timestamps once someone takes the time to figure it out.  Looking for help on #29.  

Regarding the issue of it enabling timestamps even though you didn't specify them, I think that is a bug introduced recently.  There should have been a check for the timestamps flag and it should exclude them as the documentation says.

Instead of providing a monkey patch, would you be interested in providing a fix for everyone?  We are always looking for help and your contributions are welcome.

drujensen:
@masukomi are you interested in working on this?

damianham:
fix proposed in https://github.com/amberframework/granite-orm/pull/135

drujensen:
Issue resolved in #135

PR #100, opened 2017-12-09
Cleaning up the test suite

robacarp:
As discussed in #87, I'm proposing code that cleans up the test suite. Specifically, I think it's important to test all of the ORM features symmetrically across all supported drivers. Hopefully this will clear up where drivers differ, and make bringing them into parity easier.

Assuming nobody throws a red flag, I'm going to go about moving the tests which are largely not adapter specific from `spec/adapter/*` to the place that will match the method definition they're testing.

cc @amberframework/contributors


robacarp:
@drujensen @eliasjpr thanks for looking. I'm nearly done with the easy part of this spec rearrangement, and I think that'll be a good place to draw a line in the sand and merge in.  Because of the broad scope of these changes, I'm trying to make each commit relevant and consumable...hopefully that helps mitigate the size of it all.

drujensen:
@robacarp That sounds like a good approach.  Let me know when you think this is ready.

spec/spec_models.cr
```diff
@@ -0,0 +1,154 @@
...
+    if adapter == "pg"
+      primary_key_sql = "BIGSERIAL PRIMARY KEY".id
+      foreign_key_sql = "BIGINT".id
+    elsif adapter == "mysql"
+      primary_key_sql = "BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY".id
+      foreign_key_sql = "BIGINT".id
+    elsif adapter == "sqlite"
+      primary_key_sql = "INTEGER NOT NULL PRIMARY KEY".id
+      foreign_key_sql = "INTEGER".id
+    end
```
veelenga: why don't you use `case` instead ?

```
case adapter
when "pg"
when "mysql"
when "sqlite"
end
```

robacarp: Good question! Unless something has changed, case statements don’t work in macros.

spec/granite_orm/fields/timestamps_spec.cr
```diff
@@ -0,0 +1,67 @@
...
+# TODO sqlite support for timestamps
+{% for adapter in ["pg", "mysql"] %}
+  {%
+    parent_constant = "Parent#{adapter.camelcase.id}".id
+
+    # TODO mysql timestamp support should work better
+    if adapter == "pg"
+      time_kind_on_read = "Time::Kind::Utc".id
+    elsif adapter == "mysql"
+      time_kind_on_read = "Time::Kind::Unspecified".id
```
robacarp: @drujensen, @eliasjpr since granite currently has some functionality out of parity across adapters, this spec is a little bespoke and needs to be more explicit than the rest. I'm assuming that the adapters will be brought into parity at some point and this can be cleaned up. #29 and #97 for example

spec/granite_orm/querying/all_spec.cr
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    # TODO Fails under MySQL
+    # it "finds records with numbered query substition" do
+    #   name = "findable model"
+    #   model = {{ model_constant }}.new(name: name).tap(&.save)
+    #   set = {{ model_constant }}.all("WHERE name = $1", [name])
+    #   set.size.should eq 1
+    #   set.first.name.should eq name
+    # end
```
robacarp: @amberframework/granite-orm-contributors is it worth enabling this test for the two adapters that it does work under? Is this functionality documented and used to the point that it should be tested?

spec/spec_models.cr
```diff
@@ -0,0 +1,204 @@
...
+        exec <<-SQL
+          CREATE TABLE {{ school_table }} (
+            custom_id {{ primary_key_sql }},
+            name VARCHAR(255)
+          )
+        SQL
+      end
+    end
+
+    class Nation::County{{ adapter_const_suffix }} < Granite::ORM::Base
```
robacarp: @amberframework/granite-orm-contributors Finding an example in the "school" namespace that fit this was difficult. I'm open to suggestions if anyone has any.

robacarp:
@drujensen this is as far as I want to go in one pull. I think it lays out enough of a pattern that it can be followed as the project develops and it should organizationally scale well. I posed a couple questions but it's ready when the conversation around it settles down.

PR #109, opened 2018-01-05
Query logger

### LINKED ISSUE #86: Allow logging
Currently Amber shows request and some response information on logs.

```
 | DELETE  "/languages/1"  | 14.55ms
Params: _method=delete&id=1
200  | GET  "/languages"  | 1.96ms
Params: 
 | DELETE  "/languages/2"  | 27.49ms
Params: _method=delete&id=2
200  | GET  "/languages"  | 1.66ms
Params: 
 | DELETE  "/languages/3"  | 24.58ms
Params: _method=delete&id=3
200  | GET  "/languages"  | 2.28ms
Params: 
 | DELETE  "/languages/4"  | 24.58ms
Params: _method=delete&id=4
200  | GET  "/languages"  | 1.68ms
Params: 
 | DELETE  "/languages/5"  | 27.02ms
Params: _method=delete&id=5
 | DELETE  "/languages/5"  | 397.0µs
Params: _method=delete&id=5
200  | GET  "/languages"  | 1.45ms
```

Would be useful to show query data on logs too, just like Rails and Phoenix do:

```
Article Create (0.000443)   INSERT INTO "articles" ("updated_at", "title", "body", "published",
 "created_at") VALUES('2008-09-08 14:52:54', 'Debugging Rails',
 'I''m learning how to print in logs!!!', 'f', '2008-09-08 14:52:54')
```

@drujensen WDYT?

robacarp:
This makes room for Granite to log queries to a logger. Hopefully it can then be integrated with Amber Logger. This will be especially helpful for prototyping changes to the ORM. I've had the basics of this code in a stash for several months and it's been very handy when experimenting with the ORM.

One of the more useful features of Rails is the transparent logging of database queries. Hopefully this can provide similar utility.

My initial stab at integrating this with the Amber logger required no changes beyond adding this line to the `initializers/granite.cr`:

```crystal
Granite::ORM.settings.logger = Amber.settings.logger
```

However, that leaves the log output a little less clean that desired. Nevertheless I believe the integration with Amber::Logger will be straightforward.

Comments appreciated.

## To do:

- [x] get initial feedback about this idea (@amberframework/core-team please chime in)
- [x] add logging statements for other adapters
- [ ] ~filter parameters, or tap into log filtering in Amber?~
- [ ] ~Better initial logger format?~
- [x] deadletter logger for testing? (Logs shouldn't show up during `crystal spec`)

## Screenshot

Specifically, the highlighted line in the middle.

image

fixes #86

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -31,6 +31,10 @@ abstract class Granite::Adapter::Base
     yield @database
   end
 
+  def log(query : String, params = [] of String) : Nil
+    Granite::ORM.settings.logger.info "#{query}: #{params}"
+  end
+
   # remove all rows from a table and reset the counter on the id.
   abstract def clear(table_name)
 
--- a/src/adapter/pg.cr
+++ b/src/adapter/pg.cr
@@ -6,7 +6,9 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
   # remove all rows from a table and reset the counter on the id.
   def clear(table_name)
     statement = "DELETE FROM #{table_name}"
+
     log statement
+
     open do |db|
       db.exec statement
     end
@@ -24,7 +26,7 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
       stmt << " FROM #{table_name} #{clause}"
     end
 
-    log statement
+    log statement, params
 
     open do |db|
       db.query statement, params do |rs|
@@ -42,7 +44,7 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
       stmt << " WHERE #{field}=$1 LIMIT 1"
     end
 
-    log statement
+    log statement, id
 
     open do |db|
       db.query_one? statement, id do |rs|
@@ -54,7 +56,7 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
   def insert(table_name, fields, params)
     statement = String.build do |stmt|
       stmt << "INSERT INTO #{table_name} ("
-      stmt << fields.map { |name| "#{name}" }.join(",")
+      stmt << fields.map { |name| "#{name}" }.join(", ")
       stmt << ") VALUES ("
       stmt << fields.map { |name| "$#{fields.index(name).not_nil! + 1}" }.join(", ")
       stmt << ")"
@@ -91,7 +93,7 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
   def delete(table_name, primary_name, value)
     statement = "DELETE FROM #{table_name} WHERE #{primary_name}=$1"
 
-    log statement
+    log statement, value
 
     open do |db|
       db.exec statement, value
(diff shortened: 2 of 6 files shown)


src/adapter/base.cr
```diff
@@ -31,6 +31,10 @@ abstract class Granite::Adapter::Base
     yield @database
   end
 
+  def log(query : String, params = [] of String) : Nil
+    Granite::ORM.settings.logger.info "#{query}: #{params}"
```
eliasjpr: maybe not default to `logger.info` since this will always will output the severity `(info)`?

src/granite_orm/settings.cr
```diff
@@ -1,6 +1,14 @@
+require "logger"
+
 module Granite::ORM
   class Settings
     property database_url : String? = nil
+    property logger : Logger
+
+    def initialize
+      @logger = Logger.new STDOUT
+      @logger.progname = "Granite"
```
eliasjpr: 💯

src/adapter/pg.cr
```diff
@@ -5,8 +5,10 @@ require "pg"
 class Granite::Adapter::Pg < Granite::Adapter::Base
   # remove all rows from a table and reset the counter on the id.
   def clear(table_name)
+    statement = "DELETE FROM #{table_name}"
+    log statement
```
eliasjpr: How can the logger be turned off?

eliasjpr:
There should be a way to turn off logging for production. Also just like Amber.logger there should be a way to hide/obfuscate sensitive data. It seems that the output on the amber side does not display `Granite`. I guess something like this would fix that

```
Granite::ORM.settings.logger = Amber.settings.logger
Granite::ORM.settings.logger.progname = "Granite"
```

drujensen:
This addresses #86.  A question regarding hiding sensitive data was brought up by @veelenga.  Any concerns about passwords, etc?

drujensen:
@eliasjpr If the logger passed to granite is set to `errors` only, then I think you won't get these messages in production, correct?

robacarp:
@eliasjpr @drujensen @veelenga thanks for a quick review!

Regarding filtering and masking data, I don't think it will be a big lift, but I would hope there would be no need to filter passwords from query logs -- plain text passwords stored in a database is a bad idea. Are there other reasons that fields would need to be filtered? I'm happy to add code which will do the filtering if so.

Regarding disabling the logger in production, I would do as @drujensen suggests and set the log level to error.

eliasjpr:
Also what about logging MySQL and SqLite?

eliasjpr:
Maybe the filtering of sensitive data from queries could be a setting for granite where developers can pass array, tuple of fields to hide info from.

drujensen:
@eliasjpr good catch.  @robacarp can you add logging to mysql and sqlite as well?  If not, we can do that in another PR

robacarp:
Unless @amberframework/granite-orm-contributors decides we need filtering now, I think I'm done here.

eliasjpr:
We can open an issue for the filtering. We should definitely filter query params IMHO it does not have to happen in the same PR

PR #114, opened 2018-01-15
[Logger] Improves query logging readability

eliasjpr:
I find it a little hard to spot the Grantite log line in the console

- Adds colorization to SQL keywords to highlight better the SQL queries
being logged to the console
- Adds setting to turn off colorization.

With this changes is easier to read the SQL logged line.

screen shot 2018-01-15 at 9 35 28 am

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -1,3 +1,4 @@
+require "colorize"
 require "../granite_orm"
 require "db"
 
@@ -5,6 +6,12 @@ require "db"
 # objects to perform actions against a specific database.  Each adapter needs
 # to implement these methods.
 abstract class Granite::Adapter::Base
+  DATABASE_YML = "config/database.yml"
+  SQL_KEYWORDS_REGEX = %r(SELECT|FROM|WHERE|LIMIT|INNER|VALUES|ORDER|UPDATE|INSERT|DELETE|DROP|
+  LEFT|RIGHT|GROUP|HAVING|GROUP|BY|CREATE|EXISTS)
+
+  Colorize.enabled = Granite::ORM.settings.colorize?
+
   property database : DB::Database
 
   def initialize(adapter : String)
@@ -14,8 +21,6 @@ abstract class Granite::Adapter::Base
       raise "database url needs to be set in the config/database.yml or DATABASE_URL environment variable"
     end
   end
-
-  DATABASE_YML = "config/database.yml"
 
   def settings(adapter : String)
     if File.exists?(DATABASE_YML) &&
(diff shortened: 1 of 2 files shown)


src/adapter/base.cr
```diff
@@ -32,7 +37,8 @@ abstract class Granite::Adapter::Base
   end
 
   def log(query : String, params = [] of String) : Nil
-    Granite::ORM.settings.logger.info "#{query}: #{params}"
+    Granite::ORM.settings.logger.info query.gsub(SQL_KEYWORDS_REGEX, "\\0".colorize(:white).mode(:bold).to_s )
```
robacarp: This colorization could probably be done once and saved? Or is the compiler just going to optimize that?

src/adapter/base.cr
```diff
@@ -1,10 +1,17 @@
+require "colorize"
 require "../granite_orm"
 require "db"
 
 # The Base Adapter specifies the interface that will be used by the model
 # objects to perform actions against a specific database.  Each adapter needs
 # to implement these methods.
 abstract class Granite::Adapter::Base
+  DATABASE_YML = "config/database.yml"
+  SQL_KEYWORDS_REGEX = %r(SELECT|FROM|WHERE|LIMIT|INNER|VALUES|ORDER|UPDATE|INSERT|DELETE|DROP|
```
robacarp: I like having all of this in one place but now this regexing happens always regardless of wether or not a logger is present. I think we need a performance test for granite. 

It would be a bit more manual but I have been thinking about putting colorization macros into each adapters query methods, that way formatting and colorization can be hand tuned for each type of statement at the adapter level. Then, when wanting to disable the query logger, the macros can be nilled and the code gets stripped at compile time.

src/adapter/base.cr
```diff
@@ -32,7 +37,8 @@ abstract class Granite::Adapter::Base
   end
 
   def log(query : String, params = [] of String) : Nil
-    Granite::ORM.settings.logger.info "#{query}: #{params}"
+    Granite::ORM.settings.logger.info query.gsub(SQL_KEYWORDS_REGEX, "\\0".colorize(:white).mode(:bold).to_s )
+    Granite::ORM.settings.logger.info params
```
robacarp: This line is just left dangling when it shows up in the logs. Can we put a label on it, or better format that somehow?

src/adapter/base.cr
eliasjpr: Not sure, but I agree maybe we can colorize and cache it

src/adapter/base.cr
eliasjpr: We can label to something like `Granite::ORM.settings.logger.info "Params: #{params}"`

src/adapter/base.cr
eliasjpr: See https://github.com/crystal-lang/crystal/blob/80a975cf0a98a4ce8464d0069145131667becd70/src/colorize.cr#L95

drujensen:
@eliasjpr What do you think about moving this colorization into the Logger instead?  That way it won't have a performance impact when you set the log level to ERROR in production.

eliasjpr:
@drujensen sounds good will create a Granite Logger that dies the colorization. Regardless

PR #117, opened 2018-01-21
Fix validate block (#110)

### LINKED ISSUE #110: Validator code doesn't seem to work (or isn't documented)
I'm not sure I understand how validators work. The readme is a mute on the subject, so I set out to do some digging and ran up against a hard wall quickly.

On a fresh project: `amber new test -p sqlite`
with a scaffolded model: `amber generate scaffold user name:string age:integer`

I add code to the user model to get this:

```crystal
  adapter sqlite
  table_name users

  validate :name, "Name cannot be blank" do
    false
  end

  field name : String
  field age : Int64
end
```

Compiling the code yields this error:

```
> crystal run src/test.cr 
Error in src/test.cr:1: while requiring "../config/*"

require "../config/*"
^

in config/application.cr:3: while requiring "../src/models/**"

require "../src/models/**"
^

in src/models/user.cr:7: expanding macro

  validate :name, "Name cannot be blank" do
  ^~~~~~~~

in src/models/user.cr:7: expanding macro

  validate :name, "Name cannot be blank" do
  ^

in macro 'validate' /Users/robert/Documents/repositories/test/lib/granite_orm/src/granite_orm/validators.cr:12, line 1:

>  1.     @@validators << {field: :base, message: :name, block: "Name cannot be blank"}
   2.   

no overload matches 'Array(NamedTuple(field: Symbol, message: String, block: Proc(User, Bool)))#<<' with type NamedTuple(field: Symbol, message: Symbol, block: String)
Overloads are:
 - Array(T)#<<(value : T)
```

Reading the error, the compiler is choosing the wrong macro, which I assume is because the variables to the macro methods aren't typed:

```crystal
  macro validate(message, block)
    @@validators << {field: :base, message: {{message}}, block: {{block}}}
  end

  macro validate(field, message, block)
    @@validators << {field: {{field}}, message: {{message}}, block: {{block}}}
  end
```

Digging in a little, I think perhaps this code was written for an older version of Crystal?

faustinoaq:
Fixes #110

### DIFF
--- a/spec/granite_orm_spec.cr
+++ b/spec/granite_orm_spec.cr
@@ -23,6 +23,10 @@ class WebSite < Granite::ORM::Base
   adapter pg
   primary custom_id : Int32
   field name : String
+
+  validate :name, "Name cannot be blank" do
+    @name.blank?
+  end
 end
 
 describe Granite::ORM::Base do
@@ -103,6 +107,27 @@ describe Granite::ORM::Base do
       s = WebSite.new(name: "Hacker News")
       s.custom_id = 3
       s.to_json.should eq %({"custom_id":3,"name":"Hacker News"})
+    end
+
+  describe "validate fields" do
+    it "invalid field" do
+      s = WebSite.new(custom_id: 1)
+      s.valid?.should eq false
+    end
+
+    it "errors" do
+      s = WebSite.new(name: "Hacker News")
+      s.errors.should eq ["Name cannot be blank"]
+    end
+
+    it "valid field" do
+      s = WebSite.new(name: "Hacker News")
+      s.valid?.should eq true
+    end
+
+    it "empty errors" do
+      s = WebSite.new(name: "Hacker News")
+      s.errors.should eq [""]
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/granite_orm_spec.cr
```diff
@@ -105,4 +109,26 @@ describe Granite::ORM::Base do
       s.to_json.should eq %({"custom_id":3,"name":"Hacker News"})
     end
   end
+
+  describe "validate fields" do
```
KevinSjoberg: There are some rules, taken from http://www.betterspecs.org/, that could be applied here to improve the readability of the test. I'd write the following.

```
describe "validating fields"
  context "without a name" do
    it "is not valid" do
      s = Website.new(name: "")

      s.valid?.should eq false
      s.errors.should contain "Name cannot be blank"
    end 
  end

  context "when name is present" do
    it "is valid" do
      s = WebSite.new(name: "Hacker News")
      
       s.valid?.should eq true
       s.errors.empty?.should eq true
    end
  end
end

robacarp:
Thank you very much @faustinoaq

PR #118, opened 2018-01-22
Validate per instance

faustinoaq:
Currently `validators` is a class variable shared by all model instances, if an object isn't valid the next objects neither:

```crystal
class WebSite < Granite::ORM::Base
  adapter pg
  primary custom_id : Int32
  field name : String

  validate :name, "Name cannot be blank" do
    !(name.not_nil!.blank?)
  end
end

s1 = WebSite.new(name: "")
s1.valid? # => false
s1.errors # => [Error(:name, "Name cannot be blank") ]

s2 = WebSite.new(name: "Foo Bar")
s2.valid? # => false ???
s2.errors # => [Error(:name, "Name cannot be blank") ] ???
```

With this PR:

```crystal
s1 = WebSite.new(name: "")
s1.valid? # => false
s1.errors # => [Error(:name, "Name cannot be blank") ]

s2 = WebSite.new(name: "Foo Bar")
s2.valid? # => true
s2.errors # => empty
```


faustinoaq:
Please edit and squash the commits before merge this PR 😅

drujensen:
That makes sense.  I think it was working before because we were passing `self` to the Proc for validation and not using the closure.  This is way better.  Thanks for working on this!  Granite needs more love. ;-)

Issue #99, opened 2017-11-25
Catch type-casting errors in set_attributes to append to errors


opiation:
Currently in `set_attributes`, casting errors simply bubble up making errors potentially ambiguous when creating records with a param hash.

```crystal
class Post < Granite::ORM::Base
  field likes : Int32
end

p = Post.new({ "likes" => "" })
```

Results in:
```
Invalid Int32:
```

These errors are not stored in p.errors as this happens before `#valid?` is called.  This means that Post hashes provided to create a new Post must be validated before being sent to Post.new.

Perhaps, `cast_to_field` could rescue type casting errors and append the type error message to `p.errors` including the field name.  This would allow some core validation to remain in Granite rather than also having it in each model or in controllers.

drujensen:
@opiation This is an excellent point.  Also, save should call validate and return an error or a boolean when it fails now that validation is integrated with granite.

drujensen:
@opiation Is this something you would like to work on?  if not, @amberframework/granite-orm-contributors anyone want to work on this?

opiation:
I believe having `Granite::ORM::Base#save` validate and return a boolean or error would indeed make Granite a little more ActiveRecord-like, though that seems already tracked separately in #81.

As for wanting to work on this, I'll review the contribution guide and my schedule.  In that time however, I welcome anyone to take on the task as I may not have much time in the near future.

PR #125, opened 2018-01-27
add error if casting fails

drujensen:
This change catches any casting errors and adds it to the `errors` instead of throwing an exception.  It addresses issue #99

### DIFF
--- a/spec/granite_orm/fields/casting_spec.cr
+++ b/spec/granite_orm/fields/casting_spec.cr
@@ -1,0 +1,25 @@
+# class Review{{ adapter_const_suffix }} < Granite::ORM::Base
+#   field name : String
+#   field downvotes : Int32
+#   field upvotes : Int64
+#   field sentiment : Float32
+#   field interest : Float64
+#   field published : Bool
+#   field created_at : Time
+# end
+
+{% for adapter in GraniteExample::ADAPTERS %}
+  {% model_constant = "Review#{adapter.camelcase.id}".id %}
+
+  describe "{{ adapter.id }} #casting_to_fields" do
+    it "casts string to int" do
+      model = {{ model_constant }}.new({ "downvotes" => "32" })
+      model.downvotes.should eq 32
+    end
+
+    it "generates an error if casting fails" do
+      model = {{ model_constant }}.new({ "downvotes" => "" })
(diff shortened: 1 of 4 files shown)


spec/granite_orm/fields/casting_spec.cr
```diff
@@ -0,0 +1,25 @@
+# class Review{{ adapter_const_suffix }} < Granite::ORM::Base
```
faustinoaq: Should we keep this comment?

src/granite_orm/fields.cr
```diff
@@ -140,6 +140,8 @@ module Granite::ORM::Fields
           {% end %}
         {% end %}
       end
+    rescue ex
+      errors << Granite::ORM::Error.new(name, ex.message)
```
eliasjpr: Nice! 👍

spec/granite_orm/fields/casting_spec.cr
eliasjpr: Agree we should remove this. The code tells the story very well on whats going on.

src/granite_orm/error.cr
```diff
@@ -1,7 +1,7 @@
 class Granite::ORM::Error
   property field, message
 
-  def initialize(@field : Symbol, @message : String)
+  def initialize(@field : (String | Symbol), @message : String?)
```
robacarp: Does this effect how model error messages are rendered? Maybe it should default to “” instead of being nilable?

src/granite_orm/fields.cr
robacarp: This code is pretty well constrained to the type casting, but if it is possible to be more specific about what is being rescued, that would probably be safer.

drujensen: @robacarp Yeah, I was hoping too as well.  I looked to see if I could catch a specific Exception but there is none provided by crystal String.  I don't want to try and determine this by the error message itself since there are many possibilities.

spec/granite_orm/fields/casting_spec.cr
drujensen: I found it difficult to find the model definition so found this useful, but I will remove it.

PR #126, opened 2018-01-27
only rescue DB errors

### LINKED ISSUE #116: Save and Destroy should not eat exceptions
The [#save and #destroy methods](https://github.com/amberframework/granite-orm/blob/master/src/granite_orm/transactions.cr#L35-L41) contain a wide open rescue which eats any exception and prevents the stack trace from being emitted.

Does it make sense to constrain the rescue to a few types emitted by the different drivers and rescue only those, so that errors which happen elsewhere aren't swallowed?

One example which just bit me: a `before_save` method raised an exception.

drujensen:
This fixes #116

### DIFF
--- a/src/granite_orm/transactions.cr
+++ b/src/granite_orm/transactions.cr
@@ -34,7 +34,7 @@ module Granite::ORM::Transactions
         return true
       rescue ex : DB::Error
         if message = ex.message
-          puts "Save Exception: #{message}"
+          Granite::ORM.settings.logger.error "Save Exception: #{message}"
           errors << Granite::ORM::Error.new(:base, message)
         end
         return false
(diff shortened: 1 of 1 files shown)


src/granite_orm/transactions.cr
```diff
@@ -32,7 +32,7 @@ module Granite::ORM::Transactions
         end
         __run_after_save
         return true
-      rescue ex
+      rescue ex : DB::Error
         if message = ex.message
           puts "Save Exception: #{message}"
```
eliasjpr: Should puts here be instead using the `Granite::Settings.logger.error` instead?

src/granite_orm/transactions.cr
```diff
@@ -48,7 +48,7 @@ module Granite::ORM::Transactions
         @@adapter.delete(@@table_name, @@primary_name, {{primary_name}})
         __run_after_destroy
         return true
-      rescue ex
+      rescue ex : DB::Error
         if message = ex.message
           puts "Destroy Exception: #{message}"
```
eliasjpr: Same as above comment

robacarp:
A test for this might be kind of weird, but the bug I was seeing is that exceptions raised in hook procs we’re getting swallowed. Maybe there can be a test around that?

Issue #81, opened 2017-10-22
Save should call validate and fail before saving


eshaiju:
I am very much new to Crystal and Amber. trying to create a basic blog application in amber.  I like to add presence validation in title and body of post model. how can we achieve this?

```
require "granite_orm/adapter/pg"

class Post < Granite::ORM::Base
  adapter pg

  field name : String
  field body : String
  field draft : Bool
  timestamps
end

```

coderhs:
You can mention it inside the DB and access those errors when it fails to save. https://github.com/amberframework/granite-orm#errors

eshaiju:
@coderhs I am asking about keywords equivalent to **  validates :name, presence: true**. I didn't find any in Doc

coderhs:
So I just read through the code and I can see that its not properly documented, so thats a good candidate for a PR.

the validation can be added as so:

```rb
require "granite_orm/adapter/sqlite"

class Comment < Granite::ORM::Base
  adapter sqlite

  table_name post_comments
  field name : String
  field body : String

  validate(:name, "Should not be blank", ->(x : self) { x.name != nil && x.name != "" })
end
```

But the interesting point to note is that validation is not checked when you run `object.save`, only validations that are checked at that point are the database validation. If you want to check the validations you mention then you need to run a `object.valid?`.

When you run the `.valid?` method it will execute the block you passed in and if it failed then place the custom message.

PS: Even if the validators you placed fail it will still get saved to the DB, but if there are any DB level constrains that fail then it won't get saved. Unlike rails it won't raise an exception, but make those errors available in the `object.errors` array.

Xosmond:
So there is not basic validations implemented ?

drujensen:
@Xosmond This was fixed in #127

PR #127, opened 2018-01-27
check if model is valid before trying to save

drujensen:
fixes ticket #81


robacarp:
Actually, there should be a test for this

Issue #76, opened 2017-10-18
Model to schema naming conventions


aarongodin:
Coming from ActiveRecord, the Model name to table name convention is great to see. I did notice that there is one difference, and I wanted to bring it up as it was the only difference I saw between Granite's and ActiveRecord's. It can be seen in this error:

```cr
icr(0.23.1) > SongThread.all
relation "songthreads" does not exist (PQ::PQError)
```

The Rails Guides outline this convention for ActiveRecord:

> Database Table - Plural with underscores separating words (e.g., book_clubs).
> Model Class - Singular with the first letter of each word capitalized (e.g., BookClub).

In this case, you'd expect the schema relation to be called `song_threads`.

Here's my SongThread class for completeness to the example:

```cr
require "granite_orm/adapter/pg"

class SongThread < Granite::ORM::Base
  adapter pg

  field bpm : Int32
  field title : String
  field primary_account : Int32
  field secondary_account : Int32

  timestamps

  def primary_account
    Account.find @primary_account
  end

  def secondary_account
    Account.find @secondary_account
  end
end
```

drujensen:
I agree, we should change this. 

We should also look at adding a pluralize helper since adding an `s` is not ideal.  we should just add a couple rules like if the word ends with an `s, ch, sh` then add `es`.  However, we shouldn't go overboard on this and add a custom dictionary of words.  It should be a straightforward predictable algorithm.

marksiemers:
Regarding the snake case table names - I think that fix is easy:
Here: https://github.com/amberframework/granite-orm/blob/master/src/granite_orm/table.cr#L30
```crystal
# Currently
{% name_space = @type.name.gsub(/::/, "_").downcase.id %}

# Change to
{% name_space = @type.name.gsub(/::/, "_").underscore.id %}
```

Regarding dealing with pluralization, I think leveraging an existing library is a good idea: https://github.com/mosop/string_inflection

robacarp:
I’d second pulling in a proper inflector. There are undoubtedly times where pedantic inflection is mote confusing to the majority of people (eg Media, Data), but more often than not I appreciated the ease of Rails inflection. In other projects I’ve attempted to live with a cheaper solution and regretted it.

eliasjpr:
There is a nice inflector created here https://github.com/luckyframework/lucky_inflector

robacarp:
[...]

PR #128, opened 2018-01-27
underscore instead of downcase

drujensen:
This fixes the original issue #76 with naming convention for tables that @marksiemers suggested


drujensen:
@robacarp yeah, working on it now.  Just want to get this out there.

drujensen:
@robacarp specs added.

PR #130, opened 2018-01-28
Implement both validate syntax (blocks and procs)

faustinoaq:
> lets see if we can support both syntaxes for the validation in Granite. That will fix this and avoid a breaking change.

This PR implements the idea commented by @drujensen 

Required by: https://github.com/amberframework/amber/pull/588

### DIFF
--- a/src/granite_orm/validators.cr
+++ b/src/granite_orm/validators.cr
@@ -5,12 +5,23 @@ module Granite::ORM::Validators
 
   macro included
     macro inherited
+      @@validators = Array({field: Symbol, message: String, block: Proc(self, Bool)}).new
       @validators = Array({field: Symbol, message: String, block: Proc(Bool)}).new
       def validate!
       end
     end
   end
 
+  # First option: syntax support using procs
+  macro validate(message, block)
+    @@validators << {field: :base, message: {{message}}, block: {{block}}}
+  end
+
+  macro validate(field, message, block)
+    @@validators << {field: {{field}}, message: {{message}}, block: {{block}}}
+  end
+
+  # Second option: syntax sypport using blocks
   macro validate(message)
     def validate!
       previous_def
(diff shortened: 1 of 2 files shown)


src/granite_orm/validators.cr
```diff
@@ -5,12 +5,23 @@ module Granite::ORM::Validators
...
       end
     end
   end
 
+  # First option: syntax support using procs
+  macro validate(message, block)
+    @@validators << {field: :base, message: {{message}}, block: {{block}}}
+  end
+
+  macro validate(field, message, block)
```
robacarp: Can these parameters be typed? When this code gets accidentally triggered by something like `validate(:name, "cant be blank", "must be present")` the compiler error messages are going to be confusing. Any help we can give the compiler will result in it providing better error feedback to devs.

Maybe crystal doesn't support this with macros, I don't remember.

src/granite_orm/validators.cr
```diff
@@ -5,12 +5,23 @@ module Granite::ORM::Validators
 
   macro included
     macro inherited
+      @@validators = Array({field: Symbol, message: String, block: Proc(self, Bool)}).new
```
robacarp: Having both a `@@validators` and a `@validators` is confusing.

robacarp: Perhaps one syntax can wrap the other? This code is an untested wild guess, but:

```crystal
macro validate(field, message, &block)
  validate field, message, -> (model : Model) do
    {{ yield }}
  end
end
```

src/granite_orm/validators.cr
```diff
@@ -25,7 +36,25 @@ module Granite::ORM::Validators
...
+  # By example:
+  # ```
+  # validate :name, "name can't be blank" do
+  #   !name.to_s.blank?
+  # end
+  #
+  # validate :name, "name can't be blank", -> (user : User) do
+  #   !user.name.to_s.blank?
+  # end
+  # ```
```
robacarp: It's probably a good idea to have this comment which shows the validation syntax, but I don't think it belongs on the `valid?` method. Perhaps on the module?

src/granite_orm/validators.cr
drujensen: I agree.  Plus it will bring back the original issue where you validate one object and all the other objects show as invalid as well.

src/granite_orm/validators.cr
drujensen: I don't think these need to be macros.  See if we can change them to `def`.

faustinoaq:
I read your suggestions and comments, new PR here 👉 https://github.com/amberframework/granite-orm/pull/131

faustinoaq:
Replaced by #131

Issue #91, opened 2017-11-12
Not null fields


jhass:
Having to handle `nil` on every field access is a PITA. Support marking fields as not null, or maybe even make that the default and support marking them as nullable. Make the primary key not null by default

drujensen:
yeah, crystal `.not_nil!` is very annoying.  I wish they chose something like swift using just an `!` or simpler syntax.

I have considered making the mapping more advanced and support `nilable` and other options that db's may support.  It would be an options hash per field.

drujensen:
closing since this is a duplicate of #57

PR #131, opened 2018-01-29
Dj/validation using blocks and procs

drujensen:
An alternative approach to support Blocks and Procs.

The main difference is that you need to pass in a parameter to the Block:
```
validate :name, "can't be blank" do |user|
  !user.name.to_s.blank?
end

validate :name, "can't be blank", -> (user : User) do
  !user.name.to_s.blank?
end
```

This also removes the majority of the macro definitions.

If we agree on this approach, I will add separate validation specs.

### DIFF
--- a/src/granite_orm/validators.cr
+++ b/src/granite_orm/validators.cr
@@ -17,7 +17,7 @@ require "./error"
 # ```
 module Granite::ORM::Validators
   getter errors = [] of Error
-  
+
   macro included
     macro inherited
       @@validators = Array({field: String, message: String, block: Proc(self, Bool)}).new
--- a/src/granite_orm/validators.cr
+++ b/src/granite_orm/validators.cr
@@ -1,34 +1,48 @@
 require "./error"
 
+# Analyze validation blocks and procs
+#
+# By example:
+# ```
+# validate :name, "can't be blank" do |user|
+#   !user.name.to_s.blank?
+# end
+#
+# validate :name, "can't be blank", -> (user : User) do
+#   !user.name.to_s.blank?
+# end
+#
+# name_required = ->(model : Granite::ORM::Base) { !model.name.to_s.blank? }
+# validate :name, "can't be blank", name_required
+# ```
 module Granite::ORM::Validators
   getter errors = [] of Error
-
+  
   macro included
     macro inherited
-      @validators = Array({field: Symbol, message: String, block: Proc(Bool)}).new
-      def validate!
+      @@validators = Array({field: String, message: String, block: Proc(self, Bool)}).new
+
+      def self.validate(message : String, &block : self -> Bool)
+        self.validate(:base, message, block)
+      end
+
+      def self.validate(field : (Symbol | String), message : String, &block : self -> Bool)
+        self.validate(field, message, block)
+      end
+
+      def self.validate(message : String, block : self -> Bool)
+        self.validate(:base, message, block)
+      end
+
+      def self.validate(field : (Symbol | String), message : String, block : self -> Bool)
+        @@validators << {field: field.to_s, message: message, block: block}
       end
     end
   end
 
-  macro validate(message)
-    def validate!
-      previous_def
-      @validators << {field: :base, message: {{message}}, block: ->{{{yield}}}}
-    end
-  end
-
-  macro validate(field, message)
-    def validate!
-      previous_def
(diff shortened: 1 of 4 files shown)


robacarp:
There's a lot of code flying around today regarding validators, so I want to pitch in what I have been using. The block/proc thing wasn't working for me and I didn't want to take the time to look into it so this is what I came up with. It's not always concise but it does have some things going for it.

```crystal
class Granite::ORM::Base
  def validate : Nil
  end

  def valid? : Bool
    clean_errors
    validate
    errors.none?
  end

  def clean_errors
    errors = [] of Error
  end

  def add_error(field : Symbol, message : String)
    errors << Error.new(field, message)
  end
end
```

```crystal
class Domain < Granite::ORM::Base
  def validate : Nil
    blank_name = true
    if name = @name
      blank_name = name.blank?
    end

    add_error :name, "cannot be blank" if blank_name
    return if blank_name

    malformed_name = true
    if name = @name
      malformed_name = ! name.index("/").nil?
      malformed_name ||= name[0...4] == "http"
    end

    if malformed_name
      add_error :name, "should be the DNS name to be checked. For example: google.com instead of http://google.com/gmail"
    end
  end
end
```

As you can see I'm running several checks on the same field.

The major pain point here is the need need to check if the field exists manually `name = @name` before asking about it's formatting because Crystal doesn't like it if you call `#index` on nil. Perhaps someone with more crystal typing chops can do some magic for a not_nil! getter. Perhaps #91 would solve that.

However, it also has the benefit that the `validate` method lives in instance-space instead of macro-space. So it's possible to do things like this admittedly contrived example:

```crystal
def validate : Nil
  case type
  when "subdomain":
    validate_subdomain
  when "ip_address":
    validate_ip_address
  else
    validate_domain
  end
end
```

The Rails model validations have `if:` and `unless:` predicates stuffed into them and it can be hard to write maintainable validations out of that stuff. It's really easy to make a mess.

faustinoaq:
@eliasjpr Should we implement this on param validators too?

drujensen:
@robacarp thanks for the information.  Hopefully the set of changes we are making will help.

@amberframework/core-team @amberframework/contributors This is ready for review.   This will address the breaking change introduced and fix Amber specs.

This also removes the macros for validation and uses methods instead.

src/granite_orm/validators.cr
```diff
@@ -1,34 +1,48 @@
...
+      def self.validate(message : String, block : self -> Bool)
+        self.validate(:base, message, block)
+      end
 
-  macro validate(field, message)
-    def validate!
-      previous_def
-      @validators << {field: {{field}}, message: {{message}}, block: ->{{{yield}}}}
+      def self.validate(field : (Symbol | String), message : String, block : self -> Bool)
+        @@validators << {field: field.to_s, message: message, block: block}
```
robacarp: doesn't this recreate the "Error on one instance, error on all instances" bug?

eliasjpr:
> @eliasjpr Should we implement this on param validators too?

We can extend the params validators to support validation for a  `Granite::ORM::Model`, I don't think it will take a lot of effort from what I can see. We did discuss last night of the possibility of normalizing validators

src/granite_orm/validators.cr
faustinoaq: I tested it and surprisingly works without that issue 😅

marksiemers:
@drujensen - Since this fixes the currently failing amber specs, as long as it doesn't introduce a new bug, I say merge it and release.

@robacarp 
> The major pain point here is the need to check if the field exists manually name = @name before asking about its formatting because Crystal doesn't like it if you call #index on nil. Perhaps someone with more crystal typing chops can do some magic for a not_nil! getter.

There is a `try` method in crystal, and it is not frowned upon using it (at least not by RX14). When called on `nil`, it just doesn't yield to the block. 

Perhaps this would clean things up a little bit:
```crystal
class Domain < Granite::ORM::Base
  ERROR_MSG = {
    dns: "should be the DNS name to be checked. For example: google.com instead of http://google.com/gmail",
    blank: "cannot be blank",
  }

  def validate : Nil
    (add_error :name, ERROR_MSG[:blank]; return) if @name.try &.blank?
    (add_error :name, ERROR_MSG[:dns]) if name.try { |name| name.index("/").nil? || name[0...4] == "http" }
  end
end
```

PR #133, opened 2018-01-31
Spec for from_sql, and a big mock for the DB class

robacarp:
I don't know if this is the right way to go about this, as always I'm open to suggestions.

I just happened to notice that this method didn't have any express testing around it, though it is being de-facto tested every time a record is being read from the database.


faustinoaq:
Fixes #159

robacarp:
@amberframework/granite-orm-contributors this is up to date with the latest spec patterns and is ready to get review

spec/mocks/db_mock.cr
```diff
@@ -0,0 +1,89 @@
...
+    @position += 1
+    @field_position = 0
+  end
+
+  def read
+    if @position >= @values.size
+      raise "Overread"
+    end
+
+
```
eliasjpr: extra line

PR #135, opened 2018-02-03
parse timestamp strings for sqlite db adapter

damianham:
A quick simple fix to handle created_at/updated_at when using sqlite database adapter, however the values must be populated (which should be the case if the timestamps settings value has been enabled)

### DIFF
--- a/spec/spec_models.cr
+++ b/spec/spec_models.cr
@@ -34,9 +34,9 @@ end
     elsif adapter == "sqlite"
       primary_key_sql = "INTEGER NOT NULL PRIMARY KEY".id
       foreign_key_sql = "INTEGER".id
-      created_at_sql = "".id
-      updated_at_sql = "".id
-      timestamp_fields = "".id
+      created_at_sql = "created_at VARCHAR,".id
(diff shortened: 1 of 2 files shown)


damianham:
I agree that it breaking if timestamps are null is not ideal but that is a consequence of adding the timestamps macro to a table with existing records.  We could say in the documentation that if you add the timestamps macro to a table with existing records you also need to update the created_at, updated_at columns to a default value.

robacarp:
Given that SQLite support has a different set of use cases than regular database support, that sounds fine. I didn’t realize it was only when adding timestamps macro to an existing table with nulls.

drujensen:
@damianham Agree with @robacarp on a couple specs here.  Also, not sure why pg syntax failed in specs.  Let me know if this is not related to your change.

damianham:
@drujensen I enabled the timestamps specs and the test suite ran fine locally when I ran docker-compose run spec  but the travis build failed when I committed the change to the specs.

drujensen:
@damianham ok, I will pull down your branch and take a look when I get a chance.

PR #137, opened 2018-02-07
Add #count

maiha:
I need `count` anyway 😄 

```crystal
  adapter mysql
  primary id : Int32
end
```

```
% amber exec User.count
491_i64
Exec
```

Best regards,


spec/granite_orm/querying/count_spec.cr
```diff
@@ -0,0 +1,22 @@
...
+
+    it "returns a number of the all records for the model" do
+      2.times do |i|
+        {{ model_constant }}.new(name: "model_#{i}").tap(&.save)
+      end
+
+      count = {{ model_constant }}.count
+      count.should eq 2
+    end
+  end
```
robacarp: ❤️ this spec, thanks @maiha

src/granite_orm/querying.cr
```diff
@@ -92,6 +92,11 @@ module Granite::ORM::Querying
     end
   end
 
+  # count returns a count of all the records
+  def count : Int64
```
robacarp: No objection here but why did you choose `Int64`?

maiha: I was thinking about handling large tables like 10 billion data.

However, in this case, special queries such as `CAST AS BIGINT` may be required. In fact, there are no rare cases beyond `Int32` at this moment.

So I'd like to modify it to use `Int32`. Thought?

robacarp: I have no preference. I think Postgres might return a different type depending on how much data comes back, so that may be an issue.

Issue #122, opened 2018-01-27
Provide something like ActiveRecord::Base.connection.quote


robacarp:
Hopefully this is buried and salty to use, but I think the ORM should provide it. Especially since this ORM is still developing (as in #121), it would be nice to be able to manually escape a string where no other option exists.

drujensen:
@robacarp will this work for you or do you think we still need the `quote` capability?

robacarp:
I think Granite should still have quote capability. That kind of depth is the right amount of depth though because it should be raising flags of warning when used. `Model.adapter.database.quote "string"` is at least [salty enough](http://catb.org/jargon/html/S/syntactic-salt.html) to be noticed by code reviewers and to cause a bit of rethinking by developers.

It's hopefully not _common_ but I have used ActiveRecord's adapter quote mechanism a time or two. I can't remember why I couldn't use a parameterized expression, but I'm sure I would have tried that road first.

Maybe it's worth waiting on the need to appear though?

faustinoaq:
Is this already being fixed by #150 ? or duplicate of #149 ?

drujensen:
Closing this issue.

PR #139, opened 2018-02-11
allow reserved words as table and column names

maiha:
An SQL syntax error occurred using reserved keywords for column names such as `key`,` all`, and so on.

```sql
... WHERE key=?

ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to
your MySQL server version for the right syntax to use near 'key=?' at line 1
```

This PR provides a quoting feature for table names and column names.
```sql
SELECT `kvs`.`id`, `kvs`.`key`, `kvs`.`value` FROM `kvs` WHERE `key`=? LIMIT 1;
INSERT INTO `select` (`all`) VALUES (?);
```

Best regards,

### DIFF
--- a/spec/spec_models.cr
+++ b/spec/spec_models.cr
@@ -18,6 +18,7 @@ end
     nation_county_table = "nation_county_#{adapter_literal}s".id
     review_table = "review_#{adapter_literal}s".id
     empty_table = "empty_#{adapter_literal}s".id
+    reserved_word_table = "select".id
 
     if adapter == "pg"
       primary_key_sql = "BIGSERIAL PRIMARY KEY".id
@@ -230,6 +231,22 @@ end
       end
     end
 
+    class ReservedWord{{ adapter_const_suffix }} < Granite::ORM::Base
+      adapter {{ adapter_literal }}
+      table_name "{{ reserved_word_table }}"
+      field all : String
+
+      def self.drop_and_create
+        exec "DROP TABLE IF EXISTS #{quoted_table_name}"
+        exec <<-SQL
+          CREATE TABLE #{quoted_table_name} (
+            id {{ primary_key_sql }},
+            #{quote("all")} VARCHAR(255)
+          )
+        SQL
+      end
+    end
+
     module GraniteExample
       @@model_classes << Parent{{ adapter_const_suffix }}
       @@model_classes << Teacher{{ adapter_const_suffix }}
@@ -240,6 +257,7 @@ end
       @@model_classes << Nation::County{{ adapter_const_suffix }}
       @@model_classes << Review{{ adapter_const_suffix }}
       @@model_classes << Empty{{ adapter_const_suffix }}
+      @@model_classes << ReservedWord{{ adapter_const_suffix }}
     end
 
     Spec.before_each do
@@ -252,6 +270,7 @@ end
       Nation::County{{ adapter_const_suffix }}.clear
       Review{{ adapter_const_suffix }}.clear
       Empty{{ adapter_const_suffix }}.clear
+      ReservedWord{{ adapter_const_suffix }}.clear
     end
 
 {% end %}
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -60,6 +60,15 @@ abstract class Granite::Adapter::Base
     Granite::Adapter::Base.env(url)
   end
 
+  # Use macro in order to read a constant defined in each subclasses.
+  macro inherited
+    # quotes table and column names
+    def quote(name : String) : String
+      char = QUOTING_CHAR
+      char + name.gsub(char, "#{char}#{char}") + char
+    end
+  end
+
   # class level method so we can test it
   def self.env(url)
     regex = /\$\{(.*?)\}/
--- a/src/adapter/mysql.cr
+++ b/src/adapter/mysql.cr
@@ -3,9 +3,11 @@ require "mysql"
 
 # Mysql implementation of the Adapter
 class Granite::Adapter::Mysql < Granite::Adapter::Base
+  QUOTING_CHAR = '`'
+
   # Using TRUNCATE instead of DELETE so the id column resets to 0
   def clear(table_name)
-    statement = "TRUNCATE #{table_name}"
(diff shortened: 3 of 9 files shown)


src/adapter/mysql.cr
```diff
@@ -5,7 +5,7 @@ require "mysql"
 class Granite::Adapter::Mysql < Granite::Adapter::Base
   # Using TRUNCATE instead of DELETE so the id column resets to 0
   def clear(table_name)
-    statement = "TRUNCATE #{table_name}"
+    statement = "TRUNCATE #{quote_table_name(table_name)}"
```
robacarp: This is part of where I was hoping to head with #122, thank you for implementing it.

spec/spec_models.cr
```diff
@@ -230,6 +231,22 @@ end
...
+      def self.drop_and_create
+        exec "DROP TABLE IF EXISTS #{quoted_table_name}"
+        exec <<-SQL
+          CREATE TABLE #{quoted_table_name} (
+            id {{ primary_key_sql }},
+            #{quoted_column_name("all")} VARCHAR(255)
+          )
+        SQL
+      end
+    end
```
robacarp: Thank you for following the pattern with your tests in this pull request, it makes it a lot easier to see how the test suite is actually covering all the code.

src/adapter/mysql.cr
```diff
@@ -20,8 +20,8 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
   def select(table_name, fields, clause = "", params = [] of DB::Any, &block)
     statement = String.build do |stmt|
       stmt << "SELECT "
-      stmt << fields.map { |name| "#{table_name}.#{name}" }.join(", ")
-      stmt << " FROM #{table_name} #{clause}"
+      stmt << fields.map { |name| "#{quote_table_name(table_name)}.#{quote_column_name(name)}" }.join(", ")
```
robacarp: It might be possible to shorten this syntax a little by providing something like this in the `Adapter::Base`:

```crystal
   def col(name : String)
      quote_column_name name
   end
```

That would make all of these statements from `quote_column_name(...)` into just `col(...)`.

src/adapter/pg.cr
```diff
@@ -111,4 +111,12 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
 
     clause
   end
+
+  def quote_table_name(name : String) : String
+    %Q("%s") % name.gsub(%Q("), %Q(""))
```
robacarp: nicely done

src/adapter/mysql.cr
```diff
@@ -91,12 +91,20 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
...
 
     log statement, value
 
     open do |db|
       db.exec statement, value
     end
   end
+
+  def quote_table_name(name : String) : String
+    %Q(`#{name.gsub("`", "``")}`)
```
robacarp: Simply for the purpose of more consistency between adapters, this could follow the same pattern as postgres.

src/granite_orm/table.cr
```diff
@@ -45,5 +45,13 @@ module Granite::ORM::Table
     def self.primary_name
       @@primary_name
     end
+
+    def self.quoted_table_name
+      @@adapter.quote_table_name(table_name)
+    end
```
robacarp: Do you think there is any drawback to making `#table_name` by default quoted? cc @drujensen

src/granite_orm/table.cr
```diff
@@ -45,5 +45,13 @@ module Granite::ORM::Table
...
+
+    def self.quoted_column_name(column_name)
```
robacarp: Since this method is providing an action, it should be named `quote_column_name` or perhaps even just `quote_column`.

src/adapter/mysql.cr
maiha: Yep. I love shorter method name! 😺 

But I'm afraid that we can't image **quoting feature** from the name of `col`. Well, how about using `quote`? In fact, both quoting logic about columns and tables are same for all three adapters. And I felt it redundant. So we can merge them into one `quote` method.

src/adapter/base.cr
```diff
@@ -60,6 +60,15 @@ abstract class Granite::Adapter::Base
     Granite::Adapter::Base.env(url)
   end
 
+  # Use macro in order to read a constant defined in each subclasses.
+  macro inherited
+    # quotes table and column names
+    def quote(name : String) : String
+      char = QUOTING_CHAR
+      char + name.gsub(char, "#{char}#{char}") + char
```
drujensen: I'm a little concerned with this logic and performance.  If someone already quoted a query:
```
select `table`.`column` from `table`;
```
what will this produce?

Also, do we need to perform a `gsub` for every table and column name or can we improve this?

maiha: Strictly speaking, it will be slightly slow, but `gsub` is fast and the target string is short so we can ignore the influence on the whole. From the performance point of view, I think there are many other places that can speed up. Furthermore, the implementation by **Crystal** is already too fast enough.

Therefore, unless it is a clear problem at the present time, I would like to prioritize "features" over "performance". Yep, I'd like to implement many features like following! 😺 
- support natural keys
- bang method like `save!`, `find!`
- avoid ugly and redundant code like `record.name.not_nil!`

drujensen: If this was called one time, I would agree but its getting called multiple times, once per each column.  I agree that the feature outweighs the cost.  I was wondering if there was a better way to do this.  I'm ok with this for now.  We can look at optimizing it later.

Thanks for your contribution.

robacarp: @drujensen hopefully granite can defer to lower level database specific adapters at some point, I'd rather not own the code that does escaping 😰

drujensen: @robacarp right, right.  I am not completely comfortable with this change.  I know we need it but something feels odd here.

drujensen:
@robacarp need a second approval.  I think we can look at optimizing this but this is a good first stab at the issue. wdyt?

PR #142, opened 2018-02-15
add belongs_to custom type support

drujensen:
This allows you to specify a custom name and type for the `belongs_to` foreign key.   I needed this for backward compatibility with a rails project that has Int(11) in mysql.
```
belongs_to :tool, tool_id : Int32
```

This does not address the `has_many` relationship.

### DIFF
--- a/spec/granite_orm/associations/belongs_to_custom_type_spec.cr
+++ b/spec/granite_orm/associations/belongs_to_custom_type_spec.cr
@@ -1,0 +1,54 @@
+require "../../spec_helper"
+
+class Tool < Granite::ORM::Base
+  adapter pg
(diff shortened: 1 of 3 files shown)


spec/granite_orm/associations/belongs_to_custom_type_spec.cr
```diff
@@ -0,0 +1,54 @@
+require "../../spec_helper"
+
+class Tool < Granite::ORM::Base
```
robacarp: @drujensen now that the excellent test macro cleanup has been merged, what do you think about tweaking this a little to run on all adapters?

faustinoaq:
Fixes #160

PR #146, opened 2018-02-16
`find`,`find_by`,`first` now ensures `T`, call with trailing `?` for `T?`

maiha:
This PR make class query methods to ensure a value exist in default like #144.
- `find`, `find_by`, `first` return `T` if exists, otherwise raises `NotFound`
- `find?`, `find_by?`, `first?` return `T?`

```crystal
User.find?(1) # => #<User:0x99a300 @id=1 ...
User.find(1)  # => #<User:0x99a300 @id=1 ...
User.find?(0) # => nil
User.find(0)  # raises Granite::ORM::Querying::NotFound.new("Couldn't find User with id=0")
```

This removes many `not_nil!` codes as in this commit.
Also this prevents newbies from asking 
- why undefined method 'xxx' for Nil?

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,28 +126,34 @@ end
 #### Find First
 
 ```crystal
-post = Post.first
+post = Post.first?
 if post
   puts post.name
 end
+
+post = Post.first # raises when no records exist
 ```
 
 #### Find
 
 ```crystal
-post = Post.find 1
+post = Post.find? 1
 if post
   puts post.name
 end
+
+post = Post.find 1 # raises when no records found
 ```
 
 #### Find By
 
 ```crystal
-post = Post.find_by :slug, "example_slug"
+post = Post.find_by? :slug, "example_slug"
 if post
   puts post.name
 end
+
+post = Post.find_by :slug, "foo" # raises when no records found
 ```
 
 #### Insert
--- a/spec/granite_orm/fields/timestamps_spec.cr
+++ b/spec/granite_orm/fields/timestamps_spec.cr
@@ -17,7 +17,7 @@ module {{adapter.capitalize.id}}
   describe "{{ adapter.id }} timestamps" do
     it "consistently uses UTC for created_at" do
       parent = Parent.new(name: "parent").tap(&.save)
-      found_parent = Parent.find(parent.id).not_nil!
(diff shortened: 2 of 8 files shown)


robacarp:
@amberframework/granite-orm-contributors  it seems like there are a couple different paradigms in Crystal about this. Some people want to use:

- `method?` to return a T?
- `method` to raise on T.nil?

and others want:

- `method` to return a T?
- `method!` to raise on a T.nil?

I personally prefer the latter, but only because of my history with Ruby/Rails. I expect `?` methods to return a boolean, and `!` methods to do something dramatic: raise, destroy data, etc.

Do we have a paradigm for this in Amber projects? Any other preferences?

drujensen:
oops. wrong button.

drujensen:
Here is a good article that explains Swift's Optionals.  Scroll down to the optional chaining to see how swift handles it.

https://hackernoon.com/swift-optionals-explained-simply-e109a4297298

robacarp:
For the release notes, is this a breaking change?

drujensen:
@robacarp no, I think the `find`, `find_by`, etc. works as it did before.

elorest:
This is a really big breaking change! We should seriously considering allowing returning value or nil instead of value or exception.

drujensen:
@robacarp @maiha I was wrong, it looks like this is a pretty big breaking change.  I was under the impression `find` and `find_by` would work as they did before and we were adding a `find?` and `find_by?`.  This is not the case and something we should consider rolling back.  WDYT?

maiha:
First, I know that this and #144 is a very big API change. But, I think that the specification where **T is default** is convenient from the following three points.

1. Since it is natural that we expect that data exists in the persistence layer 
2. Since `not_nil!` is long and ugly, the conversion cost of `T?` → `T` is high in Crystal
3. It seems very natural API to use `T` and `T?` properly according to the expectation of existence of data

This is a very big change, but I believe it is also a very big improvement.

faustinoaq:
@maiha Good points :+1: , what about using `T!` to raise an exception and making `T` niliable by default :sweat_smile:

drujensen:
https://github.com/amberframework/granite-orm/pull/171

Issue #149, opened 2018-02-20
Incorrect quoting, preventing Pg schema specification


docelic:
In one of recent commits, quoted_table_name was introduced.

This breaks functionality if table name contains a Pg schema name like schemaname.tablename.

quoted_table_name() quotes both inside double quotes to produce "schemaname.tablename", while the proper way to quote is "schemaname"."tablename".

The source of the problem is probably this function:

```
diff --git a/src/adapter/base.cr b/src/adapter/base.cr
index e29e5c5..1a5958c 100644
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -60,6 +60,15 @@ abstract class Granite::Adapter::Base
     Granite::Adapter::Base.env(url)
   end
 
+  # Use macro in order to read a constant defined in each subclasses.
+  macro inherited
+    # quotes table and column names
+    def quote(name : String) : String
+      char = QUOTING_CHAR
+      char + name.gsub(char, "#{char}#{char}") + char
+    end
+  end
```

docelic:
@maiha mentioning you here since it seems relevant.

maiha:
Nice! It's a bug I've created. I'll create a PR soon.

robacarp:
Since there's an open pull, I'm locking this issue to prevent forking the discussion. See #150 as referenced above by the github bots.

drujensen:
@robacarp do you have a recommendation on how to address this issue?

robacarp:
I can think of a couple things:

1. Provide a method called `quoted_table_name` which does the right thing for most use cases, but can be overridden in models if a developer is using a namespaced table
1. Provide a method called `quote_table_name` which simply looks for a period and divides the string if needed, then calls `quote` on the appropriate parts.

PR #150, opened 2018-02-20
Fix `quote` about "table.column", and it now guarantees idempotency.

maiha:
As @docelic said in #149, there is a bug about incorrect quoting.

```crystal
quote("foo.bar")  # => "foo.bar" (expected "foo"."bar")
```

This PR
- fixed to handle dotted input correctly
- added specs about this issue
- also now guarantees idempotency

For example, `Mysql` adapter works as follows.
```crystal
quote("foo")         # => "`foo`"
quote("foo.bar")     # => "`foo`.`bar`" (FIXED)
quote("`foo`")       # => "`foo`"       (NEW: idempotency)
quote("`foo`.`bar`") # => "`foo`.`bar`" (NEW: idempotency)
```

Best regards,

### DIFF
--- a/spec/adapter/mysql_spec.cr
+++ b/spec/adapter/mysql_spec.cr
@@ -1,1 +1,28 @@
 require "../spec_helper"
+
+private def adapter
+  Granite::Adapter::Mysql.new("mysql")
+end
+
+describe Granite::Adapter::Mysql do
+  describe "#quote" do
+    context "simple value" do
+      it "should return `foo`" do
+        adapter.quote("foo").should eq("`foo`")
+      end
+    end
+
+    context "dotted value" do
+      it "should return `foo`.`bar`" do
+        adapter.quote("foo.bar").should eq("`foo`.`bar`")
+      end
+    end
+
+    context "already quoted value" do
+      it "guarantees idempotency" do
+        adapter.quote("`foo`").should eq("`foo`")
+        adapter.quote("`foo`.`bar`").should eq("`foo`.`bar`")
+      end
+    end
+  end
+end
--- a/spec/adapter/pg_spec.cr
+++ b/spec/adapter/pg_spec.cr
@@ -1,1 +1,28 @@
 require "../spec_helper"
+
+private def adapter
+  Granite::Adapter::Pg.new("pg")
+end
+
+describe Granite::Adapter::Pg do
+  describe "#quote" do
+    context "simple value" do
+      it %(should return "foo") do
+        adapter.quote("foo").should eq(%("foo"))
+      end
+    end
+
+    context "dotted value" do
+      it %(should return "foo"."bar") do
+        adapter.quote("foo.bar").should eq(%("foo"."bar"))
+      end
+    end
+
+    context "already quoted value" do
+      it "guarantees idempotency" do
+        adapter.quote(%("foo")).should eq(%("foo"))
+        adapter.quote(%("foo"."bar")).should eq(%("foo"."bar"))
+      end
+    end
+  end
+end
--- a/spec/adapter/sqlite_spec.cr
+++ b/spec/adapter/sqlite_spec.cr
@@ -1,1 +1,28 @@
 require "../spec_helper"
+
+private def adapter
+  Granite::Adapter::Sqlite.new("sqlite")
+end
+
+describe Granite::Adapter::Sqlite do
+  describe "#quote" do
+    context "simple value" do
+      it %(should return "foo") do
+        adapter.quote("foo").should eq(%("foo"))
+      end
+    end
+
+    context "dotted value" do
+      it %(should return "foo"."bar") do
+        adapter.quote("foo.bar").should eq(%("foo"."bar"))
+      end
+    end
+
+    context "already quoted value" do
+      it "guarantees idempotency" do
+        adapter.quote(%("foo")).should eq(%("foo"))
+        adapter.quote(%("foo"."bar")).should eq(%("foo"."bar"))
+      end
+    end
+  end
+end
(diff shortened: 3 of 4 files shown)


docelic:
The approach with deleting quote chars before inserting them prevents someone from having a legitimate table with a dot as part of the name.

The commit is surely a fix for the currently bug and as such could be accepted, but how about checking whether the name already contains at least one quote char, and if yes, not making any changes on it?

robacarp:
@maiha thanks for submitting a patch for this so quickly.

I think this has a few more far reaching consequences than it seems at first. I don't think the quote method should be this clever. It should simply do a quoting, escaping characters which would interfere with that quoting. Quoting is an escaping mechanism, it shouldn't need to know anything about the contents of what it is quoting. 

Any regard to idempotency is a risk for injection vulnerability. Quoting something multiple times should produce multiple escaped quotations. Anything else is dangerous.

I don't think this is a bug with the quoting function, but missing logic around generating table names for SQL queries. 

Granite currently knows nothing about multiple database in mysql, or multiple schemas in postgres. It really only knows about _tables_ and their fields. 

My first instinct is that at this point, Granite should continue to not know anything about cross schema querying. I would much rather invest time into a proper query builder, queries with joins, allowing multiple connection strings, etc. 

@docelic I'm assuming you're working with postgres because the issue talks about multiple schemas. Can you share some of your reasoning for wanting to divide your database out into multiple schemas?

spec/adapter/mysql_spec.cr
```diff
@@ -1 +1,28 @@
...
+    end
+
+    context "dotted value" do
+      it "should return `foo`.`bar`" do
+        adapter.quote("foo.bar").should eq("`foo`.`bar`")
+      end
+    end
+
+    context "already quoted value" do
+      it "guarantees idempotency" do
```
robacarp: This is a big change, and we need to have some discussion around it.

docelic:
@robacarp even if a person is using one (non-default) schema they'd encounter this issue, unless they found a way how to invoke Pg's "set search path" command on all new connections in the connection pool.

robacarp:
> unless they found a way how to invoke Pg's "set search path" command on all new connections in the connection pool.

@docelic I think Rails solves this with multiple connections, and [doing exactly that as part of their database.yml](https://stackoverflow.com/questions/8806284/using-multiple-postgresql-schemas-with-rails-models).

maiha:
> Quoting something multiple times should produce multiple escaped quotations.

@robacarp It seems Rails(AR) doesn't provide idempotency. You're right! 😺 
```ruby
quote_table_name("foo")         # => "`foo`"
quote_table_name("foo.bar")     # => "`foo`.`bar`"
quote_table_name("`foo`")       # => "```foo```"
quote_table_name("`foo.bar`")   # => "```foo`.`bar```"
quote_table_name("`foo`.`bar`") # => "```foo```.```bar```"
```

marksiemers:
>  It seems Rails(AR) doesn't provide idempotency. You're right! 😺

Is there a reason for that decision?

robacarp:
> Is there a reason for that decision?

@marksiemers The reason you can't be idempotent and quote strings is that you cannot know what the data contains, or how many times it has already been escaped. 

value | escaped value
--------- | -------------------
`my name isn't robert` | `'my name isn''t robert'`
`'my name isn''t robert'` | `'''my name isn''''t robert'''`

And so on. Any attempt to cleverly inspect the string for evidence of escaping can be fooled, and therefore create an injection vulnerability.

marksiemers:
Quoting values would be different from quoting object names (tables, columns, schemas), right?

Would it be worth the effort to have a `quote_database_object` method that was idempotent?

robacarp:
Man, I wish the drivers would implement this so Granite doesn't have to. 

Quoting a database object is similarly unpredictable because of the possibility for dotted table names, etc. You never know what some pre-existing database someone is going to try to wire Granite into.

I think this isn't a bug in the quote method, but a bug in the way the table name macro is used to build the `FROM` part of a sql query. 

What if, instead of just [setting a const value in `table_name`](https://github.com/amberframework/granite-orm/blob/master/src/granite_orm/table.cr#L20-L22), it defines a method which can be overridden or even manually supplied?

Something like:

```crystal
macro table_name(name)
  def escaped_table_name
    @adapter.quote name.id
  end
end
```

And then update the query generation to use that method instead of the `@@table_name`.

That would allow a dev to specify whatever they want for the table name.

docelic:
Pretty nice idea @robacarp

maiha:
Well, let's clarify which **quoting** we are talking about. Because there are multiple **quoting methods** depending on the target as @marksiemers  says. And the goals may be different too.

For example, ActiveRecord has mainly three types of quoting methods, and they works differently.
1. `quote_table_name`
2. `quote_column_name`
3. `quoto_TYPE` (TYPE is a type like `string`, `true`, `date`, ...)

Nonetheless, I implemented one `quote` method with imaging 1 and 2. This choice was a mistake 
 cause many users may image 3. And this brought about a lot of discussion confusion. So, we'd better tell **quoting what** or the number.

I agreed with @robacarp that there are no needs for idempotency about 3. And agreed with @marksiemers that it's worth about 1,2.

faustinoaq:
Fixes #149

drujensen:
@robacarp Can you review this again?

robacarp:
I still maintain that idempotency is not a feature of a quoting system, but a bug.

faustinoaq:
> thanks for your input. I agree that the risk and complexity of this change is not worth the benefits. I think you're right regarding trying to make this idempotent is a bug and risky. I am going to close this PR.

Can we resuse this code somewhere? (maybe in crystal-db) like @eliasjpr said :sweat_smile:

PR #154, opened 2018-02-26
add spec for callbacks

maiha:
https://github.com/amberframework/granite-orm/pull/153#discussion_r170689455
> an exception thrown in a hook will get swallowed

This PR explicitly specifies why `DB::Error` is rescued in `#save`. After this PR, we can detect #153 is well regressed. 😄 

In addition, this also ensures the callbacks are fired and its order.
Best regard,

### DIFF
--- a/spec/spec_models.cr
+++ b/spec/spec_models.cr
@@ -235,6 +235,32 @@ end
       end
     end
 
+    class Callback < Granite::ORM::Base
+      adapter {{ adapter_literal }}
+      table_name callbacks
+      primary id : Int64
+      field name : String
+
+      property history : IO::Memory = IO::Memory.new
+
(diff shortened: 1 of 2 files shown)


spec/spec_models.cr
```diff
@@ -235,6 +235,32 @@ end
...
+      field name : String
+
+      property history : IO::Memory = IO::Memory.new
+
+      {% for name in Granite::ORM::Callbacks::CALLBACK_NAMES %}
+        {{name.id}} _{{name.id}}
+        private def _{{name.id}}
+          history << "{{name.id}}\n"
+        end
+      {% end %}
```
robacarp: hm, this is clever

faustinoaq:
Fixes #162

PR #164, opened 2018-03-21
UUID

wontruefree:
reopeing PR #102.  This has been rebased and conflicts resolved.


wontruefree:
I am adding tests for associations but travis gets diffrent results then my local setup.

drujensen:
@wontruefree Take a look at #166 Supporting Natural Keys.  I would like to use that approach for UUID support.  Let me know if that will support your use case.  If so, I would like to close this in favor of that.

wontruefree:
@drujensen thanks for the update I will check it out.  I got stuck on associatons and ran out of time to work on this.

faustinoaq:
> with the latest release, you can now specify the primary uuid : String, auto: false and then add a before_create hook to generate the uuid. I am going to close this PR. Let me know if the solution doesn't work for you and we can reopen.

@drujensen I think we can add a test for that, WDYT?

drujensen:
yes, we could do that and use it as an example when others ask how to perform UUID.

PR #166, opened 2018-03-27
Supporting natural keys

maiha:
This PR enhances `Model.primary` to accept `auto` option that works as a natural key if false.

### Example (use RDB as KVS)

```
mysql> DESCRIBE kvss;
+-------+--------------+------+-----+---------+-------+
| Field | Type         | Null | Key | Default | Extra |
+-------+--------------+------+-----+---------+-------+
| k     | varchar(255) | NO   | PRI | NULL    |       |
| v     | varchar(255) | YES  |     | NULL    |       |
+-------+--------------+------+-----+---------+-------+
```

```crystal
class Kvs < Granite::ORM::Base
  adapter mysql
  primary k : String
  field   v : String
end
```

First, if you use a primary key other than Int as described above, it will give a compile error with following suggestion.

```crystal
Failed to define Kvs#save: Primary key must be Int(32|64), or set `auto: false` for natural keys.

  primary k : String, auto: false
```

After adding `auto: false` that disables **auto number** feature, we can play with nartural keys.

```crystal
class Kvs < Granite::ORM::Base
  adapter mysql
  primary k : String, auto: false
  field   v : String
end
```

```crystal
## Create
port = Kvs.new(k: "mysql_port", v: "3306")
port.new_record? # => true
port.save        # => true
# INSERT INTO `kvs` (`v`, `k`) VALUES (?, ?): ["3306", "mysql_port"]
Kvs.count        # => 1

## Read
port = Kvs.find("mysql_port")
port.new_record? # => false
port.v           # => "3306"

## Update
port.v = "3307"
port.new_record? # => false
port.save        # => true
# UPDATE `kvs` SET `v`=? WHERE `k`=?: ["3307", "mysql_port"]

## Delete
port.destroy     # => true
Kvs.count        # => 0
```

Best regards,

### DIFF
--- a/spec/granite_orm/fields/primary_key_spec.cr
+++ b/spec/granite_orm/fields/primary_key_spec.cr
@@ -1,0 +1,24 @@
+{% for adapter in GraniteExample::ADAPTERS %}
+module {{adapter.capitalize.id}}
+  describe "{{ adapter.id }} .new" do
+    it "works when the primary is defined as `auto: true`" do
+      Parent.new
+    end
+
+    it "works when the primary is defined as `auto: false`" do
+      Kvs.new
+    end
+  end
+
+  describe "{{ adapter.id }} .new(primary_key: value)" do
+    it "ignores the value in default" do
+      Parent.new(id: 1).id?.should eq(nil)
+    end
+
+    it "sets the value when the primary is defined as `auto: false`" do
+      Kvs.new(k: "foo").k?.should eq("foo")
+      Kvs.new(k: "foo", v: "v").k?.should eq("foo")
+    end
+  end
+end
+{% end %}
--- a/README.md
+++ b/README.md
@@ -104,6 +104,18 @@ end
 
 This will override the default primary key of `id : Int64`.
 
+### Natural Key
+
+For natural keys, you can set `auto: false` option to disable auto increment insert.
+
+```crystal
+class Site < Granite::ORM::Base
+  adapter mysql
+  primary code : String, auto: false
+  field name : String
+end
+```
+
 ### SQL
 
 To clear all the rows in the database:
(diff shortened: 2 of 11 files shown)


drujensen:
@maiha Excellent work!  💯

Blacksmoke16:
```
port = Kvs.new(k: "mysql_port", v: "3306")
p port.inspect
```
```
"#<Kvs:0x5591aabc39c0 @new_record=true, @destroyed=false, @errors=[], @k=nil, @v=\"3306\", @updated_at=nil, @created_at=nil>"
```

Shouldn't the k instance variable be set to "mysql_port" when inspecting the object?

maiha:
Pushed -f to resolve conflicts.

spec/granite_orm/fields/primary_key_spec.cr
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  describe "{{ adapter.id }} .new(primary_key: value)" do
+    it "ignores the value in default" do
+      Parent.new(id: 1).id?.should eq(nil)
+    end
+
+    it "sets the value when the primary is defined as `auto: false`" do
+      Kvs.new(k: "foo").k?.should eq("foo")
+      Kvs.new(k: "foo", v: "v").k?.should eq("foo")
```
maiha: @Blacksmoke16
Yep, I just added `new` with mass-assignment yesterday into this PR!
**maiha:master** + `new(*mass-assignment*)` → **maiha:natural-key**

If you are using **maiha:master** branch, could you try again with **maiha:natural-key** or **this pr** branch? This should work!

Blacksmoke16: Ah perfect! Was behind a few commits i guess :)

drujensen: 👍

robacarp:
@maiha thank you for this contribution

PR #167, opened 2018-03-27
Fix .create class method

Adam-Stomski:
From commit history, this looks like regression after extracting some modules.

Example usage:
```
parent = Parent.create(name: "Test Parent")
```
or (more useful example)
```
params = { name: "Test Parent" }.to_h
parent = Parent.create(params)
```
instead of 
```
parent = Parent.new
parent.name = "Test Parent"
parent.save
```


robacarp:
@Adam-Stomski thank you for this fix!

Issue #156, opened 2018-03-11
Ability to bulk import array of models


Blacksmoke16:
# Use case:
Use from_json or from_yaml to produce an array of db models.  Currently (afaik) the only way to import them all is to iterate through the array and call `.save` on each model, which causes an INSERT statement to be executed for each record.

An `.import` or something similar method on the array would create a few transactions to bulk insert a few thousand records at a time.

See https://github.com/zdennis/activerecord-import as an example.

Blacksmoke16:
A workaround until a better solution is implemented

```Crystal
class Array
  def import
    self.each_slice(2500) do |s|
      c = s.first.class
      stmt = "INSERT INTO #{c.to_s.downcase}s (#{c.fields.join(",")}) VALUES "
      s.each_with_index do |o, idx|
        stmt += "(#{o.values.join(",") { |i| %("#{i}") }})"
        stmt += "," if (s.size - 1) != idx
      end
      Name.exec stmt
    end
  end
end
```

Also need a `values` model method
```Crystal
  def values
    c = self.class
    keys = c.fields
    keys << c.primary_name unless c.primary_name == "id"
    keys.map { |k| self.to_h[k] }
  end
```

robacarp:
@Blacksmoke16 aside from monkeypatching Array, this seems like it's worth fleshing out. The recently added [Adapter&quote](https://github.com/amberframework/granite-orm/blob/master/src/adapter/base.cr#L66-L69) method should help.

I think it's probably best to avoid doing any polymorphism and instead require the array to be single-model only, so that the table name isn't part of the data.

Blacksmoke16:
For sure, this sped up importing for me from a good while to like 30sec.

I'll have to check that out.  Current issue i found with my quick monkeypatch is working with null values, it quotes them so it doesn't know what to do when going to run the INSERT.  Shouldn't be a major issue to fix.  Requiring the array to be of a single model would be fine.  Simple and effective to start off would be best.

But yea, a legit feature to do this would be great to have.

Blacksmoke16:
I been messing with this for a bit wanted to see what your thoughts are on this in regards to the best way to implement it.

[...]

Issue #151, opened 2018-02-22
question: Is `ORM` needed?


maiha:
Hi all, I'm simply wondering why it's named as `Granite::ORM::Base`.
For example, ActiveRecord is quite natural.
- `ActiveRecord::Base`  # base
- `ActiveRecord::ConnectionAdapters::PostgreSQLAdapter` # and its adapters

In Granite
- `Granite::ORM::Base`
- `Granite::Adapters::Pg`

I prefer shorter naming like `Granite::Base`. 
```crystal
class User < Granite::Base
```
Furthermore, I prefer `granite` for the repository name if there are no plans to release any other `granite-xxx` projects.
Thought?

drujensen:
I am fine with this change and agree that ORM is not needed.  We will either need to make it backward compatible or coordinate closely with amber cli so the templates are updated with the correct major version change.

robacarp:
I agree, but changing the repo name is kind of painful because it breaks all builds generated with the old repository name.

maiha:
How about making a new repository named `granite` while leaving` granite-orm`?
- `granite-orm` : exists for old applications, but should be readonly (no more issues and pr)
- `granite` : a new repository used from the next version of amber like `v0.7`

Since there are only 2 bug issues and 3 PRs, I think that it is a good opportunity to execute now. 😺

eliasjpr:
What about `Granite::Model` instead of `Granite::Base`

```crystal
class User < Granite::Model
end
```

faustinoaq:
> I agree, but changing the repo name is kind of painful because it breaks all builds generated with the old repository name.

I think that's not a problem, github still redirects it for some time and amber is not v1.0.0 yet

I propose to change this repo name

BTW, I like `Granite::Model` as well :+1:

robacarp:
I know Amber isn't 1.0 yet, but I don't think we should be in the habit of abandoning users. The redirect is actually going to make things worse because it stays around long enough for everyone to forget that the old one ever existed and when things break, it's harder to find out what happened.

If we want to change the name of the repo, here are the steps I'd recommend following:

- Renaming this repo to `granite` will preserve the issues, pull requests, releases, etc.
- Then, a new `granite-orm` can be created (via fork or whatever)
- adding some macro code to `granite-orm` which will inform users that the dependency is out of date, that the shard is no longer maintained, and where to find the new replacement.
- Updating the readme to provide instructions.
- put `granite-orm` into Archive mode. 
[...]

PR #168, opened 2018-03-30
Bulk import

Blacksmoke16:
PR for #156 

## Progress
- [x] Create method/initial MySQL adapter code
- [x] Find a better way to handle quoting of String datatypes.
- [x] Test with various MySQL/granite datatypes
- [x] Handle ON DUPLICATE KEY UPDATE/IGNORE
- [x] Add Postgres/sqlite adapter versions
- [x] Add tests
- [x] Update Readme

## Questions:

1. How should the lifecycle hooks be handled? 
Either, hooks would _not_ run for each model being inserted, but would get triggered once all are inserted and such.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 /.crystal/
 /doc/
 /config/*.db
-/.idea/
 
 # Libraries don't need dependency lock
 # Dependencies will be locked in application that uses them
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /.crystal/
 /doc/
 /config/*.db
+/.idea/
 
 # Libraries don't need dependency lock
 # Dependencies will be locked in application that uses them
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
-# Granite::ORM
+# Granite
 
 [Amber](https://github.com/Amber-Crystal/amber) is a web framework written in
-the [Crystal](https://github.com/manastech/crystal) language.
+the [Crystal](https://github.com/crystal-lang/crystal) language.
 
 This project is to provide an ORM in Crystal.
 
-[![Build Status](https://img.shields.io/travis/amberframework/granite-orm.svg?maxAge=360)](https://travis-ci.org/amberframework/granite-orm)
+[![Build Status](https://img.shields.io/travis/amberframework/granite.svg?maxAge=360)](https://travis-ci.org/amberframework/granite)
 
 ## Installation
 
@@ -16,8 +16,8 @@ with kemal or any other framework as well.
 
 ```yaml
 dependencies:
-  granite_orm:
-    github: amberframework/granite-orm
+  granite:
+    github: amberframework/granite
 
   # Pick your database
   mysql:
@@ -47,12 +47,12 @@ Or you can set the `DATABASE_URL` environment variable.  This will override the 
 
 ## Usage
 
-Here is an example using Granite ORM Model
+Here is an example using Granite Model
 
 ```crystal
-require "granite_orm/adapter/mysql"
+require "granite/adapter/mysql"
 
-class Post < Granite::ORM::Base
+class Post < Granite::Base
   adapter mysql
   field name : String
   field body : String
@@ -63,9 +63,9 @@ end
 You can disable the timestamps for SqlLite since TIMESTAMP is not supported for this database:
 
 ```crystal
-require "granite_orm/adapter/sqlite"
+require "granite/adapter/sqlite"
 
-class Comment < Granite::ORM::Base
+class Comment < Granite::Base
   adapter sqlite
   table_name post_comments
   field name : String
@@ -95,7 +95,7 @@ For legacy database mappings, you may already have a table and the primary key i
 We have a macro called `primary` to help you out:
 
 ```crystal
-class Site < Granite::ORM::Base
+class Site < Granite::Base
   adapter mysql
   primary custom_id : Int32
   field name : String
@@ -109,7 +109,7 @@ This will override the default primary key of `id : Int64`.
 For natural keys, you can set `auto: false` option to disable auto increment insert.
 
 ```crystal
-class Site < Granite::ORM::Base
+class Site < Granite::Base
   adapter mysql
   primary code : String, auto: false
   field name : String
@@ -138,41 +138,41 @@ end
 #### Find First
 
 ```crystal
-post = Post.first?
+post = Post.first
 if post
   puts post.name
 end
 
-post = Post.first # raises when no records exist
+post = Post.first! # raises when no records exist
 ```
 
 #### Find
 
 ```crystal
-post = Post.find? 1
+post = Post.find 1
 if post
   puts post.name
 end
 
-post = Post.find 1 # raises when no records found
+post = Post.find! 1 # raises when no records found
 ```
 
 #### Find By
 
 ```crystal
-post = Post.find_by? :slug, "example_slug"
+post = Post.find_by :slug, "example_slug"
 if post
   puts post.name
 end
 
-post = Post.find_by :slug, "foo" # raises when no records found
+post = Post.find_by! :slug, "foo" # raises when no records found
 ```
 
 #### Insert
 
 ```crystal
 post = Post.new
-post.name = "Granite ORM Rocks!"
+post.name = "Granite Rocks!"
 post.body = "Check this out."
 post.save
 ```
@@ -251,7 +251,7 @@ post = Post.all("ORDER BY posts.name DESC LIMIT 1").first
 `belongs_to` and `has_many` macros provide a rails like mapping between Objects.
 
 ```crystal
-class User < Granite::ORM::Base
+class User < Granite::Base
   adapter mysql
 
   has_many :posts
@@ -265,7 +265,7 @@ end
 This will add a `posts` instance method to the user which returns an array of posts.
 
 ```crystal
-class Post < Granite::ORM::Base
+class Post < Granite::Base
   adapter mysql
 
   belongs_to :user
@@ -313,18 +313,18 @@ Instead of using a hidden many-to-many table, Granite recommends always creating
 Then you can use the `belongs_to` and `has_many` relationships going both ways.
 
 ```crystal
-class User < Granite::ORM::Base
+class User < Granite::Base
   has_many :participants
 
   field name : String
 end
 
-class Participant < Granite::ORM::Base
+class Participant < Granite::Base
   belongs_to :user
   belongs_to :room
 end
 
-class Room < Granite::ORM::Base
+class Room < Granite::Base
   has_many :participants
 
   field name : String
@@ -353,19 +353,19 @@ CREATE INDEX 'room_id_idx' ON TABLE participants (room_id);
 As a convenience, we provide a `through:` clause to simplify accessing the many-to-many relationship:
 
 ```crystal
-class User < Granite::ORM::Base
+class User < Granite::Base
   has_many :participants
   has_many :rooms, through: participants
 
   field name : String
 end
 
-class Participant < Granite::ORM::Base
+class Participant < Granite::Base
   belongs_to :user
   belongs_to :room
 end
 
-class Room < Granite::ORM::Base
+class Room < Granite::Base
   has_many :participants
   has_many :users, through: participants
 
@@ -393,7 +393,7 @@ end
 
 ### Errors
 
-All database errors are added to the `errors` array used by Granite::ORM::Validators with the symbol ':base'
+All database errors are added to the `errors` array used by Granite::Validators with the symbol ':base'
 
 ```crystal
 post = Post.new
@@ -408,9 +408,9 @@ There is support for callbacks on certain events.
 Here is an example:
 
 ```crystal
-require "granite_orm/adapter/pg"
+require "granite/adapter/pg"
 
-class Post < Granite::ORM
+class Post < Granite::Base
   adapter pg
 
   before_save :upcase_title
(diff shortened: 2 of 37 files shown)


src/adapter/mysql.cr
```diff
@@ -76,6 +76,31 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
...
+    stmt << "#{quote(primary_name)}, " unless primary_auto
+    stmt << fields.map { |field| quote(field) }.join(", ")
+    stmt << ") VALUES "
+
+    model_array.each do |model|
+      if model.responds_to? :to_h
+        next unless model.valid?
+        stmt << "("
+        stmt << model.to_h[primary_name].to_s + ", " unless primary_auto
+        stmt << fields.map { |field| model.to_h[field].is_a?(String) ? "'" + model.to_h[field].to_s + "'" : model.to_h[field] }.join(", ")
```
Blacksmoke16: I'm not a fan of how I'm doing this.  Suggestions welcome!

robacarp: The adapters now provide [a quote method](https://github.com/amberframework/granite-orm/blob/master/src/adapter/base.cr#L66-L69) but the functionality is has some bugs.

src/granite_orm/transactions.cr
```diff
@@ -7,9 +7,21 @@ module Granite::ORM::Transactions
     @updated_at : Time?
     @created_at : Time?
 
+    # The import class method will run a batch INSERT statement for each model in the array
+    # the array must contain only one model class
+    # invalid model records will be skipped
+    def self.import(model_array)
+      raise ArgumentError.new("Model class mismatch:  expected array of only #{self} models.") unless model_array.all? { |model| model.class == self }
```
robacarp: Can this definition be changed so that the compiler will enforce the type constraint instead of a runtime error?

First guess, something like this might work:

```crystal
def self.import(model_array : Array(self))
```

.gitignore
```diff
@@ -4,6 +4,7 @@
 /.crystal/
 /doc/
 /config/*.db
+/.idea/
```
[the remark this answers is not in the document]
robacarp: Oh, this is an editor thing. You should setup and ignore your editor files in a system wide [global gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) instead.

src/granite_orm/transactions.cr
Blacksmoke16: Ooo yes!  I will do that tonight (still getting used to static typed land :P)

src/adapter/mysql.cr
Blacksmoke16: The quoting method I do use for column/table names. However, since it quotes with the backtick that couldn't be used for string values afaik?

Blacksmoke16: That was my initial thought, however when I tried it I kept getting like `db params invalid format` or something like that (not home atm to look).  The params should just be an array of values, which the first value of array maps to first ? placeholder correct?

.gitignore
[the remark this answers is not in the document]
faustinoaq: @Blacksmoke16 interesting, Are you coding with Intellij Idea? just curious, What plugin are you using to have crystal support? :sweat_smile:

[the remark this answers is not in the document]
Blacksmoke16: @faustinoaq I am yes :)  Using Rubymine with https://plugins.jetbrains.com/plugin/10213-crystal-language-plugin.

Is pretty meh in regards to full native support like Ruby in Rubymine, but it provides syntax highlighting and that's good enough for me atm.

src/adapter/mysql.cr
```diff
@@ -76,6 +76,32 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
...
+      stmt << "INSERT INTO #{quote(table_name)} ("
+      stmt << "#{quote(primary_name)}, " unless primary_auto
+      stmt << fields.map { |field| quote(field) }.join(", ")
+      stmt << ") VALUES "
+
+      model_array.each do |model|
+        next unless model.valid?
+        stmt << '('
+        stmt << "?," unless primary_auto
+        stmt << Array.new(fields.size, '?').join(',')
```
Blacksmoke16: The primary field references here can be removed once #169 gets merged, since fields will cover it by default.

Blacksmoke16:
@robacarp What do you think the best way to handle the lifecycle hooks is for this?

Best approach/worth doing/etc?

Blacksmoke16:
Am currently having some issues with timestamps and/or the import statement setting timestamps.

https://travis-ci.org/amberframework/granite-orm/builds/360749166#L579

Am not sure super what is causing it...

robacarp:
@Blacksmoke16 

- My guess about the error you're getting is that you're not adding the id to the select statement, or that the fields are in the wrong order.
- Re: lifecycle events, what did you have in mind? Something other than the before/after create?

Blacksmoke16:
I'll look into that, thanks!

Probably best to just keep it before_create, run import statement, after_create?

robacarp:
@Blacksmoke16 ah, of course. I think they should run for each record. For example a user object which needs a password hash is not a valid record until the before_create executes. 

Proposing a workflow here:

- `Model.import(array_of_records)` is called
- `before_create` is called on each object
- every object is inserted
- bail if something errored
- `after_create` is called on each object
- `Model.import` returns

@amberframework/granite-orm-contributors thoughts?

robacarp:
@Blacksmoke16 FYI, depending on your personal git workflow, you may need to update your `.git/config` to point to the corrected granite repository, as of my mistake in #151.

src/adapter/pg.cr
```diff
@@ -76,6 +76,47 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
...
+        stmt << '('
+        stmt << fields.map_with_index { |_f, idx| "$#{index + idx + 1}" }.join(',')
+        params.concat fields.map { |field| model.to_h[field] }
+        stmt << "),"
+        index += fields.size
+      end
+    end.chomp(',')
+
+    if update_keys = options["on_duplicate_key_update"]?
+      statement += " ON CONFLICT (#{quote(primary_name)}) DO UPDATE SET "
```
robacarp: This _seems_ right, but would you mind explaining the behavior in a terse comment for future maintainers?

src/adapter/pg.cr
```diff
@@ -76,6 +76,47 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
     end
   end
 
+  def import(table_name : String, primary_name : String, fields, model_array, **options)
```
robacarp: your `**options` can be a little cleaner and have a similar effect:

```crystal
def import(table_name : String, primary_name : String, fields, model_array : Array(self), *, update_on_duplicate_key = false)
```

src/adapter/pg.cr
Blacksmoke16: Yea, I plan on updating the READ me with info on how to use everything.  But basic this was part of Postgres' 9.5 version:

https://www.postgresql.org/docs/9.5/static/sql-insert.html

It's basic the PostGres version of MySQL `ON DUPLICATE KEY UPDATE`

From the docs:
```SQL
INSERT INTO distributors (did, dname)
    VALUES (5, 'Gizmo Transglobal'), (6, 'Associated Computing, Inc')
    ON CONFLICT (did) DO UPDATE SET dname = EXCLUDED.dname;
```
Is basic saying if there is a conflict on the `did` value, then `dname` should be updated to the new value that would have otherwise been excluded.

src/adapter/pg.cr
Blacksmoke16: That would mean i would need to have a defined param for each options right?

As of now there is `on_duplicate_key_ignore: Bool`, `on_duplicate_key_update: Array(Strings) to be updated`, and possibly in future also `validate: true`, as giving option to ignore validation might speed up large imports.

robacarp: You can probably have the best of both worlds using a crystal overloaded method signature. If you need, here are [the docs on overloading](https://crystal-lang.org/docs/syntax_and_semantics/overloading.html)

robacarp: As it happens, I don’t have strong opinions on this either. I just noticed a rubyism that might be able to be cleaned up with some crystal.

Blacksmoke16:
@robacarp I got tests and readme updated.  However i'm not real sure how to implement the callbacks....Can you point me in the right direction?

I see how they are called in the other instance methods, but not sure how i would call them for each model when im in the class level import method case.

README.md
```diff
@@ -116,6 +116,69 @@ class Site < Granite::ORM::Base
...
+
+Model.import(models)
+
+Model.find!(1).name # => Fred
+
+models [
+  Model.new(id: 1, name: "George", age: 14),
+]
+
+Model.import(models, on_duplicate_key_update: %w(name))
```
robacarp: This `on_duplicate_key_update` parameter feels kind of strange, I think for two reasons. One, it's sort of a double purpose: "on duplicate update?" and "update using these fields". Second, it's really close to the the sql that it's using behind the scenes, which sort of breaks the veil that the ORM is using to obscure the dev from sql.

What if we could write something like:

```crystal
Model.import models, update_on_duplicate: true, with_fields: %(name, ...)
```

Both fields could be required with each other, so you'd never get just one of them and have a valid method call, and in Crystal we have the benefit of being able to enforce that with the compiler. Does this make it more readable, especially to someone unfamiliar with the library?

spec/granite_orm/transactions/import_spec.cr
```diff
@@ -0,0 +1,59 @@
...
+  describe "{{ adapter.id }} .import" do
+    it "should import 3 new objects" do
+      to_import = [
+        Parent.new(name: "ImportParent1"),
+        Parent.new(name: "ImportParent2"),
+        Parent.new(name: "ImportParent3"),
+      ]
+      Parent.import(to_import)
+      Parent.all("WHERE name LIKE ?", ["Import%"]).size.should eq 3
+		end
```
robacarp: Is it me or is the indentation is all strange in this file?

README.md
Blacksmoke16: I could go with that, also adds consistency to the options, as both `update_on_duplicate` and `ignore_on_duplicate` would be booleans.

robacarp: Bah, it seems like I already commented on this below...sorry

spec/granite_orm/transactions/import_spec.cr
Blacksmoke16: Yea dunno, is fine on my local so not sure what is up with that.  I'll re indent and see if that fixes it next commit.

src/granite_orm/transactions.cr
```diff
@@ -7,9 +7,20 @@ module Granite::ORM::Transactions
...
+        @@adapter.import(table_name, primary_name, fields.dup, model_array, **options)
+      rescue err
+        raise DB::Error.new(err.message)
+      end
+    end
+
     # The save method will check to see if the primary exists yet. If it does it
     # will call the update method, otherwise it will call the create method.
-    # This will update the timestamps apropriately.
+    # This will update the timestamps appropriately.
```
robacarp: 🥇

Blacksmoke16:
I'll poke around with it this weekend and we can go from there.  Thanks!

spec/granite_orm/transactions/import_spec.cr
Blacksmoke16: Yup, fixed :p

Blacksmoke16:
@robacarp from my testing I am able to trigger the callback successfully to a degree.  Since we are in a class method, the method that is to be called in the model doesn't exist.
```
in macro '__run_before_create' expanded macro: macro_139699359621456:26, line 2:

   1.       
>  2.          increment
   3.       
   4.     

undefined local variable or method 'increment'
```

Where `increment` is the name of my before_create method.

README.md
```diff
@@ -116,6 +116,69 @@ class Site < Granite::ORM::Base
 end
 ```
 
+### Bulk Insertions
+
+#### Import
+
+Each model has an `import` class level method to import an array of models in one bulk insert statement.
+```Crystal
+models [
```
Blacksmoke16: Oops, missed `=` on these, i'll fix that tonight.

src/granite_orm/transactions.cr
```diff
@@ -7,9 +7,36 @@ module Granite::ORM::Transactions
...
+    # invalid model records will be skipped
+    def self.import(model_array : Array(self))
+      begin
+        @@adapter.import(table_name, primary_name, fields.dup, model_array)
+      rescue err
+        raise DB::Error.new(err.message)
+      end
+    end
+
+    def self.import(model_array : Array(self), update_on_duplicate : Bool, columns : Array(String))
```
drujensen: since you would never use `update_on_duplicate` and `ignore_on_duplicate` what if we use an enum that has options for how to handle duplicates?  maybe `duplicates: ignore` and `duplicates: update`?

Blacksmoke16:
Sadly no, the issue is trying to run a model's instance method from the class scope.  As of now that one `__run_before_create` method does not have access to them so it cannot run.  Possibly can revisit this in another issue, to at least get the feature out.

I would have to test it if you have a specific case, otherwise iirc the whole import fails since it is just one SQL query.

src/granite_orm/transactions.cr
Blacksmoke16: Would there be a nice way to also require columns key when using the update option?

faustinoaq:
I just added Bulk import to amber documentation as well :tada: 

https://amberframework.gitbook.io/amber/guides/models/granite/bulk-insertions

PR #169, opened 2018-03-30
refactor: Unify pk and content and timestamps into `FIELDS`

maiha:
In master, when we want to iterate all fields, we have to process three kinds of fields **pk**, **content columns** and **timestamps** as following code. I feel this is a boring work and the code looks difficult.

```crystal
def to_xxx
  # for PK
  {{ PRIMARY[:name] }}

  # for Content
  {% for name, type in FIELDS %}
     ...

  # for timestamps
  {% if SETTINGS[:timestamps] %}
```

This PR merges them into one FIELDS, and the `primary` and `timestamp` now becomes just wrapper macros to the `field`. Thus, we can simply iterate `FIELDS` to scan all fields.

```crystal
def to_xxx
  {% for name, type in FIELDS %}
     ...
```

##### related refactor (Granite::ORM::Base)
- added `CONTENT_FIELDS` that removes **PK** from `FIELDS`
- added `Base.fields` to provide cached field names.
- added `Base.content_fields` to provide cached content field names.
- renamed `Base#params` to `Base#content_values` for clarity

Best regards,

### DIFF
--- a/src/granite_orm/fields.cr
+++ b/src/granite_orm/fields.cr
@@ -6,21 +6,29 @@ module Granite::ORM::Fields
 
   macro included
     macro inherited
+      CONTENT_FIELDS = {} of Nil => Nil
       FIELDS = {} of Nil => Nil
     end
   end
 
   # specify the fields you want to define and types
   macro field(decl)
-    {% FIELDS[decl.var] = decl.type %}
+    {% CONTENT_FIELDS[decl.var] = decl.type %}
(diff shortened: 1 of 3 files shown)


Blacksmoke16:
This would be quite helpful for the bulk-import PR :P

faustinoaq:
Great, @maiha Thank you so much for all your contributions!!! :smile:

PR #171, opened 2018-03-31
changes syntax for helper methods from ? to !

drujensen:
I made this change because the previous syntax was a major breaking change that
is not compatible with the Amber templates.  This accomplishes the same thing
but changes the default to be nilable and adds a bang version for `.not_nil!` equivalent.

This will allow us to avoid changing the Amber templates and is backward compatible so we don't break a bunch of peoples existing websites.

Its always recommended to check and see if a `first`, `find`, or `find_by` returns results but if you are absolutely sure you know the results will return, you can use the simpler `first!`, `find!`, or `find_by!` which will throw a runtime exception.

Same with fields, its best to check for `nil` values but if you are absolutely sure the value is not nil, then you can use the simpler `!` version of the field.  This also will throw a runtime exception if it is nil.

I recommend using these sparingly and only in quick prototypes, specs or throw away code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,34 +138,34 @@ end
 #### Find First
 
 ```crystal
-post = Post.first?
+post = Post.first
 if post
   puts post.name
 end
 
-post = Post.first # raises when no records exist
+post = Post.first! # raises when no records exist
 ```
 
 #### Find
 
 ```crystal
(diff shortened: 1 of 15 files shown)


elorest:
Love it. Thank you!

PR #172, opened 2018-03-31
add `Granite::ORM.migrator` to production codes

maiha:
Dry up codes about creating tables in specs, and put it production code as `Granite::ORM.migrator`.
This helps users
- to add tables in specs easily
- to create tables in experiment and deployment

```crystal
  adapter mysql
  field name : String
end

User.migrator.create
```

```
mysql> describe users;
+-------+--------------+------+-----+---------+----------------+
| Field | Type         | Null | Key | Default | Extra          |
+-------+--------------+------+-----+---------+----------------+
| id    | bigint(20)   | NO   | PRI | NULL    | auto_increment |
| name  | varchar(255) | YES  |     | NULL    |                |
+-------+--------------+------+-----+---------+----------------+
```

#### implementation
- mapping rules from **crystal class** to **database type** is defined in adapters.
  - `Granite::Adapter::Base::Schema::Types`
  - `Granite::Adapter::Mysql::Schema::Types`
  - ...

I just defined rules about basic classes used in current specs.
- `Bool`, `Float32`, `Float64`, `Int32`, `Int64`, `String`, `Time`

If `migrator` is exected with undefined types, it raises a runtime error as below.

```
Migrator(Granite::Adapter::Mysql) doesn't support 'Time::Span' yet. (Exception)
```

Best regards,

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,6 +451,24 @@ You can register callbacks for the following events:
 - **destroy**
 - after_destroy
 
+### Migration
+
+- `migrator` provides `drop`, `create` and `drop_and_create` methods
+
+```crystal
+class User < Granite::ORM::Base
+  adapter mysql
+  field name : String
+end
+
+User.migrator.drop_and_create
+# => "DROP TABLE IF EXISTS `users`;"
+# => "CREATE TABLE `users` (id BIGSERIAL PRIMARY KEY, name VARCHAR(255));"
+
+User.migrator(table_options: "ENGINE=InnoDB DEFAULT CHARSET=utf8").create
+# => "CREATE TABLE ... ENGINE=InnoDB DEFAULT CHARSET=utf8;"
+```
+
 ## Contributing
 
 1. Fork it ( https://github.com/amberframework/granite-orm/fork )
--- a/src/adapter/mysql.cr
+++ b/src/adapter/mysql.cr
@@ -5,6 +5,15 @@ require "mysql"
 class Granite::Adapter::Mysql < Granite::Adapter::Base
   QUOTING_CHAR = '`'
 
+  module Schema
+    TYPES = {
+      "AUTO_Int32" => "INT NOT NULL AUTO_INCREMENT PRIMARY KEY",
+      "AUTO_Int64" => "BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY",
+      "created_at" => "TIMESTAMP DEFAULT CURRENT_TIMESTAMP",
+      "updated_at" => "TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP",
+    }
+  end
+  
   # Using TRUNCATE instead of DELETE so the id column resets to 0
   def clear(table_name)
     statement = "TRUNCATE #{quote(table_name)}"
--- a/src/granite_orm/migrator.cr
+++ b/src/granite_orm/migrator.cr
@@ -1,0 +1,92 @@
+require "./error"
+
+# DB migration tool that prepares a table for the class
+#
+# ```crystal
+# class User < Granite::ORM::Base
+#   adapter mysql
+#   field name : String
+# end
+#
+# User.migrator.drop_and_create
+# # => "DROP TABLE IF EXISTS `users`;"
+# # => "CREATE TABLE `users` (id BIGSERIAL PRIMARY KEY, name VARCHAR(255));"
+#
+# User.migrator(table_options: "ENGINE=InnoDB DEFAULT CHARSET=utf8").create
+# # => "CREATE TABLE ... ENGINE=InnoDB DEFAULT CHARSET=utf8;"
+# ```
+module Granite::ORM::Migrator
+  class Base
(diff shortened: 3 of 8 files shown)


src/granite_orm/migrator.cr
```diff
@@ -0,0 +1,92 @@
...
+    def initialize(klass, @table_options = "")
+      @quoted_table_name = klass.quoted_table_name
+    end
+    
+    def drop_and_create
+      down
+      up
+    end      
+
+    def down
```
drujensen: What about renaming this to drop?

src/granite_orm/migrator.cr
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    def up
```
drujensen: and this to create?

maiha: Initially, I imagined cooperating with **micrate** as below, so I choose these names.
```
-- +micrate Up
User.migrator.up

-- +micrate Down
User.migrator.down
```

But, I found that **micrate** can handle only SQL string. Therefore, since there are no constraints now, I think that it is better to change the name as you say. 😺

src/adapter/mysql.cr
```diff
@@ -5,6 +5,15 @@ require "mysql"
 class Granite::Adapter::Mysql < Granite::Adapter::Base
   QUOTING_CHAR = '`'
 
+  module Schema
+    TYPES = {
+      "AUTO_Int32" => "INT NOT NULL AUTO_INCREMENT PRIMARY KEY",
+      "AUTO_Int64" => "BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY",
```
robacarp: The capitalization here is unexpected. Why `AUTO_Int32`?

robacarp: Can you add some quick tests for these in the adapter tests? I'd like to have them covered against accidental changes.

drujensen:
@robacarp we will not be replacing `micrate` with this anytime soon.  I used to have this exact logic in granite to automatically perform migrations, but it was too difficult to guess the intent of the developer.  For example, you didn't know what data to copy to which field if they renamed a column.

robacarp:
@drujensen Sorry, I wasn't meaning to suggest we'd provide facility for automatic migrations. Only that I can see the maintenance overhead for yet another shard being abandoned by its creator getting adopted by Amber, and absorbing the idea of a migrator into granite doesn't seem far fetched. Should that happen, the namespace `Migrator` is probably where it'd live.

drujensen:
@robacarp gotcha.  Don't get me wrong, I loved it when the migrations were built in.  It was additive so it never broke anything, but it couldn't do certain things and those limitations would require something like micrate anyway.  This PR is a good compromise and really cleans up the code.  I am very happy to see this solution.  It fits in between the gap that I lost when moving to micrate.

src/adapter/mysql.cr
maiha: Well, I'd like to resolve three kinds of **crystal objects** here.
1. Auto number (**value** + **metadata** like `Int32` or `Int64` + **this is autonumber**)
2. All crystal classes with module name (**value** like `Int32`, `Foo::Data`)
3. special field variables (**value** like `created_at` , `updated_at`)

Here, since they can be represented by **strings itself** for the case of 2 and 3, we can use simple  `Hash(String, String)` for the mapping dictionary.
Then, I faced a problem of how to embed the meta information of **autonumber** in the payload for the case 1.

Although there are various correct and complicated implementations, I prioritized the simplicity of implementation, and I embedded the meta information as "uppercase letters + _" in the string for the time being. 😸

maiha: Yes, I think it would be better to have a test. But I can not image a simple test. But, even now, Int32 and Int64 are used as primary keys in other specs, so I think the code has been covered. 😺

README.md
```diff
@@ -451,6 +451,24 @@ You can register callbacks for the following events:
...
+
+```crystal
+class User < Granite::ORM::Base
+  adapter mysql
+  field name : String
+end
+
+User.migrator.drop_and_create
+# => "DROP TABLE IF EXISTS `users`;"
+# => "CREATE TABLE `users` (id BIGSERIAL PRIMARY KEY, name VARCHAR(255));"
```
Blacksmoke16: That should probably be `id BIGINT` or `id INT` since `BIGSERIAL` is PostgreSQL specific.

drujensen:
@maiha Something is breaking the Migrator. I updated to the latest master branch and looks like I broke it:
```
Migrator(Granite::Adapter::Pg) doesn't support '{type: String}' yet. (Exception)
  from spec/spec_models.cr:255:3 in 'create'
```
Can you take a look when you get a chance?

maiha:
I am sorry, my main business has been very busy since April, and I have not had time to contribute to OSS recently.
Unfortunately, this situation seems to last for half a year, so I will close this PR once. 😢

drujensen:
@maiha Ok, i will fork this and fix it.  Thanks for your contributions!  Hope to see you back soon.  🥇

Issue #121, opened 2018-01-27
Unable to easily delete has_many without iterating and deleting each


robacarp:
This pattern isn't ideal:

```crystal
class User < Granite::ORM::Base
  has_many :posts

  before_destroy :clean_up_posts

  def clean_up_posts
    posts.map &.destroy
  end
end

class Post < Granite::ORM::Base
  belongs_to :user
end
```

Because it causes N delete queries to be executed. It would be nice to have a way to destroy an entire collection in one go. So I tried some SQL:

```crystal
  def clean_up_posts
    query = <<-SQL
    DELETE FROM posts WHERE user_id = ?
    SQL

    Post.exec query, [id]
  end
```

This may be a completely unrelated issue, but the [exec and scalar query methods](https://github.com/amberframework/granite-orm/blob/master/src/granite_orm/querying.cr#L95-L105) aren't parameterized on the models, so:

```
wrong number of arguments for 'Post.exec' (given 2, expected 0..1)
Overloads are:
 - Granite::ORM::Querying#exec(clause = "", &block)
```

As a result, I ended up with this:

```crystal
  def clean_up_posts
    query = <<-SQL
    DELETE FROM posts WHERE user_id = #{id}
    SQL

    Post.exec query
  end
```

I don't think this is immediately going to result in a compromisable injection vulnerability...in my case. But it's probably worth adding parameterized `exec` and `scalar` methods to the base.

shayneoneill:
Wouldnt the better way to handle this be to ensure DB foreign key Constraints are properly set to cascade? Keep in mind cascade also has to deal with update contingencies too. This is something modern RDBMS handle well already. Why duplicate functionality?

robacarp:
@shayneoneill leveraging RDBMS constraints could be a great way to handle this

drujensen:
@shayneoneill @robacarp I think adding db constraints in the migration scripts would be perfect.  I always hated rails for not properly supporting RDBMS constraints.  We should look into that solution further.

Maybe we can add this as a feature of #182 WDYT?

Adam-Stomski:
I'm pretty sure it should be easy to implement here: https://github.com/amberframework/granite/blob/master/src/granite/association_collection.cr

drujensen:
@Adam-Stomski yeah, we can add a `delete_all` here fairly easily. We can also support cascade deleting children rows on a `has_many` relationship.  

[...]

PR #176, opened 2018-04-05
Change .all and has_many to return *Collection object

Adam-Stomski:
My proposal for lazy loaded collections. 

Now using `.all` method queries database immediately, after this change it would return a `Granite::ORM::Collection(Model)` which forwards missing methods to the array of records.
```
# before:
Klass.all # queries database [Klass1, Klass2]

# after:
Klass.all # does not query => Granite::ORM::Collection(Klass)
Klass.all.each do |klass| # queries db
  # (...)
end
```

Now using `has_many` association methods, loads all related records from the database. After this change, it would return a `Granite::ORM::AssociationCollection(Owner, Target)`. This allows easy extensions on associations (for example https://github.com/amberframework/granite/issues/121 should be easy to implement as `association.delete_all`). Included `find_by` and `all` as an example for this.
```
# before:
student.klasss # => queries database [Klass1, Klass2]

# after:
student.klasss # does not query => Granite::ORM::AssociationCollection(Student, Klass)
student.klasss.each do |klass| # queries db
  # (...)
end
```

**new methods**
```
klasses = student.klasss.all("AND klasss.type = ? LIMIT ?", ["maths", 10]) # => Granite::ORM::Collection(Klass)
klasses.map(&.name) # queries db

klass = student.klasss.find_by(:name, "Klass A1")
klass = student.klasss.find_by!(:name, "Klass A1")
klass = student.klasss.find(1)
klass = student.klasss.find!(1)
```

I didn't want any new query syntax, just reuse what Granite already has, making associations less dumb, and prevent unnecessary db queries. Also didn't add any docs for now, would like to hear you opinion on this first :)

### DIFF
--- a/spec/granite_orm/associations/has_many_spec.cr
+++ b/spec/granite_orm/associations/has_many_spec.cr
@@ -25,6 +25,34 @@ module {{adapter.capitalize.id}}
       teacher.klasss.size.should eq 2
     end
 
+    it "can query the association" do
+      teacher = Teacher.new
+      teacher.name = "test teacher"
+      teacher.save
+
+      class1 = Klass.new
+      class1.name = "Test class X"
+      class1.teacher = teacher
+      class1.save
+
+      class2 = Klass.new
+      class2.name = "Test class X"
+      class2.teacher = teacher
+      class2.save
+
+      class3 = Klass.new
+      class3.name = "Test class with different name"
+      class3.teacher = teacher
+      class3.save
+
+      klasses = teacher.klasss.all("AND name = ? ORDER BY id DESC", ["Test class X"])
+      klasses.map(&.id).should eq [class2.id, class1.id]
+
+      klass = teacher.klasss.find_by(:name, "Test class with different name").not_nil!
+      klass.id.should eq class3.id
+      klass.name.should eq "Test class with different name"
+    end
+
     describe "#has_many, through:" do
       it "provides a method to retrieve associated objects through another table" do
         student = Student.new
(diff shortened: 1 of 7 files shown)


robacarp:
Just looking at the doc you have here in the pull, I think it’s a good idea. Granite is currently pretty chatty and I’d like to see that cleaned up however possible. 

I haven’t looked at the code here yet, but I suspect I’ve implemented something vaguely similar in the query builder (so I’m obviously onboard with the idea). 

Is the idea here to prevent queries from being run on variables that are never used?

spec/granite_orm/associations/has_many_spec.cr
```diff
@@ -66,6 +94,36 @@ module {{adapter.capitalize.id}}
...
+        enrollment2.student = student
+        enrollment2.klass = klass2
+        enrollment2.save
+
+        enrollment3 = Enrollment.new
+        enrollment3.klass = klass3
+        enrollment3.student = student
+        enrollment3.save
+
+        klasses = student.klasss.all("AND klasss.name = ? ORDER BY klasss.id DESC", ["Test class X"])
```
robacarp: This syntax is an improvement for sure

Adam-Stomski:
@robacarp I want to improve way working with relations. Since I needed something in the middle to provide methods like `student.teachers.all` I thought that it would be cool to have consistent API with lazy load using `all` method without relation too.

I believe that preventing queries from unused variables is a very good thing. I've seen a lot of code written with passing down collection through few methods to not be used because of one condition at the end. 

Since these collections forwards missing to the array the only breaking change there is returned type. I believe this is a small problem compared to possibilities it gives and "free" optimization

eliasjpr:
@faustinoaq we should label this with a `breaking change` label. 

@Adam-Stomski this is nice addition 👍

spec/granite_orm/associations/has_many_spec.cr
```diff
@@ -25,46 +25,146 @@ module {{adapter.capitalize.id}}
...
-        klass3.name = "Test class"
+        klass3.name = "Test class with different name"
+        klass3.teacher = teacher
         klass3.save
 
-        enrollment1 = Enrollment.new
-        enrollment1.student = student
-        enrollment1.klass = klass1
-        enrollment1.save
+        klass = teacher.klasss.find_by!(:name, "Test class with different name").not_nil!
```
drujensen: I don't think you need the `not_nil!` since you are using the `find_by!`

spec/granite_orm/associations/has_many_spec.cr
```diff
@@ -25,46 +25,146 @@ module {{adapter.capitalize.id}}
...
-        enrollment2 = Enrollment.new
-        enrollment2.student = student
-        enrollment2.klass = klass2
-        enrollment2.save
 
-        enrollment3 = Enrollment.new
-        enrollment3.klass = klass2
-        enrollment3.student = unrelated_student
-        enrollment3.save
+        klass = teacher.klasss.find!(klass1.id).not_nil!
```
drujensen: same here

spec/granite_orm/associations/has_many_through_spec.cr
```diff
@@ -0,0 +1,200 @@
...
+        enrollment2.student = student
+        enrollment2.klass = klass2
+        enrollment2.save
+
+        enrollment3 = Enrollment.new
+        enrollment3.klass = klass3
+        enrollment3.student = student
+        enrollment3.save
+
+        klass = student.klasss.find_by!(:name, "Test class with different name").not_nil!
```
drujensen: and here

spec/granite_orm/associations/has_many_through_spec.cr
```diff
@@ -0,0 +1,200 @@
...
+        enrollment2.student = student
+        enrollment2.klass = klass2
+        enrollment2.save
+
+        enrollment3 = Enrollment.new
+        enrollment3.klass = klass3
+        enrollment3.student = student
+        enrollment3.save
+
+        klass = student.klasss.find!(klass1.id).not_nil!
```
drujensen: here

drujensen:
@robacarp need another approval when you get a chance

Issue #57, opened 2017-09-10
Provide field! for property!


c910335:
Most of the times, we are very sure that a field is not nil (by validator, default value or SQL not null constraint), so it is annoying to check nil every time and [property!](https://crystal-lang.org/api/0.23.1/Object.html#property%21%28%2Anames%29-macro) works well for this situation.

drujensen:
nice!  I love seeing the language grow.  We used to always use `not_nil!` which is super annoying.

faustinoaq:
Is this issue already fixed?

drujensen:
Yes, they are.  closing this issue.  Thanks @faustinoaq.

PR #181, opened 2018-04-11
Add field! for property!

c910335:
relate: #57, #120, #144, #171

This PR supports the two syntaxes in #171 by #57.

For example:

```crystal
class Foo < Granite::ORM::Base
  # ...
  field number : Int32
end

foo = Foo.new(number: 0)
foo.number += 1
```

This gives:

```
Error in src/test.cr:7: undefined method '+' for Nil (compile-time type is (Int32 | Nil))

foo.number += 1
^

Rerun with --error-trace to show a complete error trace.
```

It is still annoying to use `foo.number = foo.number! + 1`.

With this PR:

```crystal
class Foo < Granite::ORM::Base
  # ...
  field! number : Int32
end

foo = Foo.new(number: 0)
foo.number += 1 # It works!
```

This PR also includes the change from `type` to `options` in #120 because it is easier to implement `field!` with this feature.
This PR also removes the property in `table.cr` because it is duplicated.

### DIFF
--- a/spec/granite_orm/fields/field_spec.cr
+++ b/spec/granite_orm/fields/field_spec.cr
@@ -1,0 +1,34 @@
+require "../../spec_helper"
+
+class Field < Granite::ORM::Base
+  adapter pg
+
+  field normal : Int32
+  field! raise_on_nil : Int32
+end
+
+describe Granite::ORM::Fields do
+  describe "field" do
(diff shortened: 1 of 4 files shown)


c910335:
@faustinoaq 
Specs were done. 😄

PR #183, opened 2018-04-22
Some enhancements about callbacks

c910335:
#### Add #abort!

A long time ago(maybe a few months), someone asked me how to abort a transaction within a callback in Amber's Gitter.
I did say that you can't, but now the feature is implemented.

Example:

```crystal
class Foo < Granite::ORM::Base
  # ...
  
  before_create do
    abort! if some_condition
  end
end
```

#### Allow multiple callbacks in one line

I don't know why I missed this in my previous PR #56.

Example:

```crystal
class Foo < Granite::ORM::Base
  # ...

  before_save method_a, method_b

  # definition of #method_a and #method_b
end
```

#### Support passing block to be a callback

This is convenient when the callback is simple.

The first example is also an example of this.

#### Move the updates of `@new_record` and `@destroyed` to proper places

Because `#abort!` in `after_create` or `after_destroy` doesn't abort the main transaction, but it would abort the update of `@new_record` and `@destroyed` before.

### DIFF
--- a/spec/granite_orm/callbacks/abort_spec.cr
+++ b/spec/granite_orm/callbacks/abort_spec.cr
@@ -1,0 +1,133 @@
+require "../../spec_helper"
+
+{% for adapter in GraniteExample::ADAPTERS %}
+module {{adapter.capitalize.id}}
+  describe "{{ adapter.id }} #abort!" do
+    context "when create" do
+      it "doesn't run other callbacks if abort at before_save" do
+        cwa = CallbackWithAbort.new(abort_at: "before_save", do_abort: true)
+        cwa.save
+
+        cwa.errors.map(&.to_s).should eq(["Aborted at before_save."])
+        cwa.history.to_s.strip.should eq("")
+        CallbackWithAbort.find("before_save").should be_nil
+      end
+
+      it "only runs before_save if abort at before_create" do
+        cwa = CallbackWithAbort.new(abort_at: "before_create", do_abort: true)
+        cwa.save
+
+        cwa.errors.map(&.to_s).should eq(["Aborted at before_create."])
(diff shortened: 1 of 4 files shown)


drujensen:
@c910335 Thanks for following up on this.

In rails, you would return true/false.  What are your thoughts on having the callbacks return a boolean instead?  Also, do you know if this will rollback any transactions it may be in?

c910335:
> Because `after_create` and `after_destroy` may abort the transaction now.

Just notice this mistake.
I meant `#abort!` in `after_create` and `after_destroy` doesn't abort the main transaction, but they would abort the update of @new_record and @destroyed before.

c910335:
If an `#abort!` is called in a callback, other callbacks and the main transaction after this callback will not be executed.

For example, here is the callback chain of create.

create: `before_save`, `before_create`, **save**, `after_create`, `after_save`

If abort at `before_create`, `after_create` and `after_save` will not be executed and the object will also not be saved.
However, if abort at `after_create`, only `after_save` will not be executed and the object will be saved successfully.

c910335:
In rails, we would `throw :abort`. 
reference: http://guides.rubyonrails.org/active_record_callbacks.html#halting-execution, rails/rails#17227

drujensen:
Question: Should we wrap all the callbacks in a transaction so that `abort!` in the `after_*` will rollback?

c910335:
I haven't thought about rollback. Should it rollback if abort at `after_*`?

robacarp:
> Should it rollback if abort at after_*?

I don't think so.

drujensen:
:dunce_hat:  man, now that i think about it, of course it shouldn't rollback in an `after_*` since that implies it completed successfully.

PR #186, opened 2018-04-27
Support belongs_to custom method name

c910335:
Example:

```crystal
class Article < Granite::ORM::Base
  # ...
  # `#author` and `#author=` with `field author_id : Int64`
  belongs_to author : User
end
```

Or both this feature and custom id in #142

```crystal
class Article < Granite::ORM::Base
  # ...
  # `#author` and `#author=` with `field author_id : String`
  belongs_to author : User,  author_id : String
end
```

This is very useful because `article.author` is much more readable than `article.user`, and an article may need to belong to more than one kind of user, for example, author and approver.


src/granite_orm/associations.cr
```diff
@@ -1,24 +1,38 @@
...
+    {% if model.is_a? TypeDeclaration %}
+      belongs_to {{model.var}}, {{model.type.id}}, {{model.var}}_id : Int64
+    {% else %}
+      belongs_to {{model.id}}, {{model.id.camelcase}}, {{model.id}}_id : Int64
+    {% end %}
   end
 
-  macro belongs_to(model_name, decl)
+  # ditto
+  macro belongs_to(model, decl)
```
robacarp: It looks like `decl` is a pattern that is already used in this file, but what does it stand for? It seems like it means `foreign_key`?

c910335:
@robacarp
I've renamed `decl` to `foreign_key`.
Thus the more readable syntax has shown.

```crystal
  belongs_to author : User, foreign_key: author_id : String
```

And I think this syntax should be recommended.

conradwt:
@robacarp Is it possible to use `belongs_to author : User` and have the foreign key inferred to be id within the users table?

PR #187, opened 2018-04-27
Refactor transactions

c910335:
I found that `#save` was too complicated while I was working on #183.
Now the callback chains are clear.


src/granite_orm/transactions.cr
```diff
@@ -34,100 +47,111 @@ module Granite::ORM::Transactions
...
+          {% raise "Failed to define #{@type.name}#save: Primary key must be Int(32|64), or set `auto: false` for natural keys.\n\n  primary #{primary_name} : #{primary_type}, auto: false\n" %}
+        {% else %}
+          if @{{primary_name}}
+            @@adapter.insert(@@table_name, fields, params, lastval: false)
+          else
+            message = "Primary key('{{primary_name}}') cannot be null"
+            errors << Granite::ORM::Error.new("{{primary_name}}", message)
+            raise DB::Error.new
+          end
+        {% end %}
```
robacarp: Could the macro code here for the primary key logic be moved into its own method?

drujensen:
@robacarp I think we can remove the extra layer of abstraction:
```
running hooks: __create __update __destroy
```

And remove the `_run_` from:
```
the actual database manipulations: __run_create __run_update and __run_destroy
```

WDYT?  This will

robacarp:
@drujensen I agree, good call

Issue #190, opened 2018-05-03
[Bulk Imports] Ability to specify batch size


Blacksmoke16:
Adding this issue for tracking purposes.

Basic will add a new optional param to the import method that determines the number of records to import in one INSERT statement.  Plan is to default it to the size of the model array, i.e. only 1 INSERT statement.

Will have PR for this prob this weekend.

Blacksmoke16:
@faustinoaq Can remove WIP this has been ready for review.

Blacksmoke16:
Done via #191

PR #191, opened 2018-05-05
Bulk import batch size

Blacksmoke16:
PR for #190


src/granite_orm/transactions.cr
```diff
@@ -10,25 +10,31 @@ module Granite::ORM::Transactions
     # The import class method will run a batch INSERT statement for each model in the array
     # the array must contain only one model class
     # invalid model records will be skipped
-    def self.import(model_array : Array(self))
+    def self.import(model_array : Array(self), batch_size : Int32 = model_array.size)
       begin
-        @@adapter.import(table_name, primary_name, fields.dup, model_array)
+        model_array.each_slice(batch_size, true) do |slice|
+          @@adapter.import(table_name, primary_name, fields.dup, slice)
```
robacarp: Can you get away with calling `#dup` on fields only once and reusing that duped copy?

Blacksmoke16: I'll take a look, good catch.

src/granite_orm/transactions.cr
```diff
@@ -12,8 +12,9 @@ module Granite::ORM::Transactions
     # invalid model records will be skipped
     def self.import(model_array : Array(self), batch_size : Int32 = model_array.size)
       begin
+        fields_duplicate = fields.dup
         model_array.each_slice(batch_size, true) do |slice|
-          @@adapter.import(table_name, primary_name, fields.dup, slice)
+          @@adapter.import(table_name, primary_name, fields_duplicate, slice)
```
drujensen: 🥇

Blacksmoke16:
Also @robacarp @drujensen do you have any better ideas to test that two separate inserts were executed?  Current test is just testing that everything was inserted correctly, which would at least mean it worked?

drujensen:
@Blacksmoke16 maybe look at the `created_at` timestamp? I'm not sure that is reliable though.

Issue #185, opened 2018-04-27
Support callbacks on bulk imports


Blacksmoke16:
Now that #168 is merged we need to revisit how to best implement callback support for this method.

The issue i was experiencing is when i call `__run_before_create`, since the import method is a class method, it does not have access to the instance methods.  

I wanted to get this issue made to see if anyone more familiar with the callbacks had any clever ideas for this.

Otherwise the idea i came up with was having the user define a `before_import` and/or `after_import` method in their models that can specifically be called in a .each before/after the import; since the names are not dynamic.  

I think this solution could work, the user could also do `before_create :before_import` so that the same method would be used for both normal creates and the import method.

WDYT?

drujensen:
I think we should just document it that import is using SQL INSERT statement and that instances are not instantiated, so the callbacks will not happen.  If you need callbacks then its best to iteratively create instances and call `.save`.

We should also document that the import may have a limitation on size of the SQL statement so batching the imports is recommended in that case.

Blacksmoke16:
That would be one option.  However it would be nice for the option to exist.  

For example my use case is me importing data from a json api into an array of models.  Each model has a `correct` property if the courier's reward is correct (based on my current pricing).  It would be much more efficient to run a quick loop to set the instance variables in a before_import than it would to save each model individually.  I'm sure there would be other use cases for this.

If something happens in the before_import you could just break out of the loop and throw and error and it would never actually get imported.  But if something happened in after_import, is there an ability to rollback?

Also, i'm planning on having a PR soonish for the user to be able to set a custom batch size, that would default to the number of elements in the array.

Blacksmoke16:
More i been thinking about this the more i agree with @drujensen to not include them by default.  The best solution to this is for the user to manually run their callback functions on the array.

Example:

```Crystal
  class Item < Granite::ORM::Base
    adapter mysql
    table_name items

    primary type_id : Int64, auto: false
    field type_name : String
    field price : Float64
    field published : Bool

    before_save :publish

    def publish
      @published = true
    end
  end
```
```Crystal
  items = [] of Item
  items << EveToolsApi::Models::Item.new(type_id: 1.to_i64, type_name: "item1", price: 0.00, published: false)
  items << EveToolsApi::Models::Item.new(type_id: 2.to_i64, type_name: "item2", price: 1.00, published: false)
  items << EveToolsApi::Models::Item.new(type_id: 3.to_i64, type_name: "item3", price: 2.00, published: false)
[...]

PR #195, opened 2018-05-09
Manually trigger callbacks

Blacksmoke16:
Depends on #194 As i used the Item class with the  `before_create` to base my tests off of.  Diff will just be two files once that gets merged in.

This PR wraps the callback run macros into functions that allow the users to call it to trigger a callback.  This is mainly intended to be a better solution to #185 as it allows for similar behavior without the extra caveats/issues.  

Example (from the UUID spec)
```Crystal
    adapter mysql
    table_name items

    primary item_id : String, auto: false
    field item_name : String

    before_create :generate_uuid

    def generate_uuid
      @item_id = UUID.random.to_s
    end
  end
```

```Crystal
items = [] of Item
items << Item.new(item_name: "item1")
items << Item.new(item_name: "item2")
items << Item.new(item_name: "item3")
items << Item.new(item_name: "item4")

items.all? { |item| item.item_id.nil? }.should be_true

items.each(&.run_before_create)

items.all? { |item| item.item_id.is_a?(String) }.should be_true

Item.import(items)
```
As in the above example this preserve the usefulness of having callbacks without integrating them into the `.import` function itself.

The only issue with this is if someone were to manually trigger their callbacks, then use `.save` for example, it would run that callback twice.  I don't really see this being an issue as there would be no reason to manually run them in that case.  

WDYT?


Blacksmoke16:
I'm open to suggestions.  Would have to test if just `before_create` would interfere with the `before_create` macro tho?

drujensen:
I think `run_before_create` is clear.

Blacksmoke16:
@drujensen I just tested it with method names same as callback name and it seemed to work fine.  All tests passed.  I can change it to that if that would be preferable?

drujensen:
@Blacksmoke16 thanks for your work on this.  excellent feature!  💯

PR #196, opened 2018-05-10
Update README for manually running callbacks

Blacksmoke16:
I forgot to update README with usage of using manually trigger callbacks with `.import` so here it is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ end
 
 #### Import
 
-**Note:  As of now, imports do not trigger callbacks.**
+**Note:  Imports do not trigger callbacks automatically.  See [Running Callbacks](#running-callbacks).**
 
 Each model has an `import` class level method to import an array of models in one bulk insert statement.
 ```Crystal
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,7 +139,7 @@ end
 
 #### Import
 
-**Note:  As of now, imports do not trigger callbacks.**
+**Note:  Imports do not trigger callbacks automatically.  See [Running Callbacks](#running-callbacks).**
```
drujensen: Should we mention that it doesn't run `validate` either?

README.md
```diff
@@ -216,6 +216,51 @@ Model.import(models, batch_size: 2)
...
+  Item.new(item_name: "item2"),
+  Item.new(item_name: "item3"),
+  Item.new(item_name: "item4"),
+]
+
+# If we did `Item.import(items)` now, it would fail since the item_id wouldn't get set before saving the record, violating the primary key constraint.
+
+# Manually run the callback on each model to generate the item_id.
+items.each(&.before_create)
+
```
drujensen: Can we manually run `validate()` here  as well?

README.md
Blacksmoke16: As of now it will skip invalid records.

https://github.com/amberframework/granite/blob/master/src/adapter/mysql.cr#L94

drujensen: @Blacksmoke16 This is always a pain with bulk import one how we handle it.  I will merge this since it's not related to the readme (yet).

PR #199, opened 2018-05-10
Fix Bulk Imports

Blacksmoke16:
This PR addresses some issues i uncovered recently, removing too much when implementing the batch size, while the tests passed.

1. Re add the extra params for `ignore_on_duplicate` and `update_on_duplicate`
2. Add `Granite::ORM::Collection(self)` typing to `model_array` param so that records can be fetched from DB, iterated upon, and reimported easily.  See [This Spec](https://github.com/Blacksmoke16/granite-orm/blob/fix-bulk-import/spec/granite_orm/transactions/import_spec.cr#L40-L47)
1. Greatly expands the test suite of bulk imports to cover standard usages, `batch_size`, `ignore_on_duplicate` and `update_on_duplicate` for each of the following:
    * Non AUTO INCREMENT default name PK
    * AUTO INCREMENT default name PK
    * Non AUTO INCREMENT custom name PK
    * AUTO INCREMENT custom name PK

### DIFF
--- a/spec/granite_orm/transactions/import_spec.cr
+++ b/spec/granite_orm/transactions/import_spec.cr
@@ -3,69 +3,186 @@ require "../../spec_helper"
 {% for adapter in GraniteExample::ADAPTERS %}
 module {{adapter.capitalize.id}}
   describe "{{ adapter.id }} .import" do
-    it "should import 3 new objects" do
-      to_import = [
-        Parent.new(name: "ImportParent1"),
-        Parent.new(name: "ImportParent2"),
-        Parent.new(name: "ImportParent3"),
-      ]
-      Parent.import(to_import)
-      Parent.all("WHERE name LIKE ?", ["Import%"]).size.should eq 3
-    end
+    describe "using the defualt primary key" do
+      context "with an AUTO INCREMENT PK" do
+        it "should import 3 new objects" do
+          to_import = [
+            Parent.new(name: "ImportParent1"),
+            Parent.new(name: "ImportParent2"),
+            Parent.new(name: "ImportParent3"),
+          ]
+          Parent.import(to_import)
+          Parent.all("WHERE name LIKE ?", ["ImportParent%"]).size.should eq 3
+        end
 
-    it "should work with on_duplicate_key_update" do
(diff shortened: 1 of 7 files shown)


Blacksmoke16:
Oo good call, yea can do.

Blacksmoke16:
@drujensen That ought to be good.

Blacksmoke16:
I'll look into it.  Will add a spec for this as well.

Blacksmoke16:
@drujensen How would i see what sqlite version the tests use?

drujensen:
good question.  In the `Docker` file, we include the library here:

https://github.com/amberframework/granite/blob/master/Dockerfile#L3

You can run `docker-compose run spec bash` to get a bash shell inside the docker image.  Not sure what command to use to get the sqlite version from there though.  Maybe write a small crystal test to run `select sqlite_version();`?

drujensen:
ok, this needs a second reviewer and then we can merge.

Blacksmoke16:
I'll update this with new spec stuff, sec

Blacksmoke16:
Rgr, seems some of the specs fail now, ill see what's wrong.

Blacksmoke16:
There is a bug with the migrator that doesn't apply `PRIMARY KEY` to column that is not AUTO INCREMENT.

I logged out the table_name k/value of the PK by doing
`puts "#{@quoted_table_name} - #{k} #{v}"`
```
"kvss" - "k" VARCHAR(255)
"people" - "id" INTEGER NOT NULL PRIMARY KEY
"companies" - "id" INTEGER NOT NULL PRIMARY KEY
"books" - "id" INTEGER NOT NULL PRIMARY KEY
"book_reviews" - "id" INTEGER NOT NULL PRIMARY KEY
"items" - "item_id" VARCHAR(255)
"non_auto_default_pk" - "id" INTEGER
"non_auto_custom_pk" - "custom_id" INTEGER
```
Notice how all the models with `auto: false` do not have `PRIMARY KEY`

drujensen:
@Blacksmoke16 ahhh, yup.  Let me push a fix for that.  thanks!

PR #203, opened 2018-05-12
move primary key out of schema types

drujensen:
This removes the PRIMARY KEY declaration out of the type and into the macro.  This allows for fields that are not auto-incremental to still be a Primary Key and for non Primary Keys to be incremental.

### DIFF
--- a/src/adapter/mysql.cr
+++ b/src/adapter/mysql.cr
@@ -7,13 +7,13 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
 
   module Schema
     TYPES = {
-      "AUTO_Int32" => "INT NOT NULL AUTO_INCREMENT PRIMARY KEY",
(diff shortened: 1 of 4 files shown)


Blacksmoke16:
Seems that even before the `auto: false` did not have `PRIMARY KEY` as natural key specs are failing due to PK duplicate entries, which should be the correct behavior imo.  Espcially when using the `primary` macro AND `auto: false`

Possibly worth adding a new spec for models that do not use `primary`, for the case of creating a table without PK for both `AUTO INCREMENT`.

drujensen:
@Blacksmoke16 yeah but if you do not specify a primary, we assume you want `primary id : Int64`

Blacksmoke16:
Ah right right, disregard that part then.  Still should update specs now that `auto: false` is having PK constraints enforced of course.

drujensen:
@Blacksmoke16 can you review?

PR #211, opened 2018-05-20
Allow `find_by` to accept multiple arguments

### LINKED ISSUE #206: find_by feature
Add ability to select only one record by few params 
**Model**.find_by(name: name_string, other_field: other_field_int)

http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by

Blacksmoke16:
This PR (#206) allows `find_by` and `find_by!` to accept multiple criteria.

This does change the syntax of the method, but i think this is a better solution long term than having two different methods with different syntaxes for each.

This touches a fair bit of the `querying.cr` since `find` was wrapping `find_by`, and since you cannot declare a dynamic keyed NamedTuple to pass to the new `find_by` I had to replace it with `first`

fixes #206

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,12 +305,13 @@ post = Post.find! 1 # raises when no records found
 #### Find By
 
 ```crystal
-post = Post.find_by :slug, "example_slug"
+post = Post.find_by(slug: "example_slug")
 if post
   puts post.name
 end
 
-post = Post.find_by! :slug, "foo" # raises when no records found
+post = Post.find_by!(slug: "foo") # raises when no records found.
+other_post = Post.find_by(slug: "foo", type: "bar") # Also works for multiple arguments.
 ```
 
 #### Insert
--- a/spec/granite/querying/find_by_spec.cr
+++ b/spec/granite/querying/find_by_spec.cr
@@ -11,10 +11,10 @@ module {{adapter.capitalize.id}}
       model.name = name
       model.save
 
-      found = Parent.find_by("name", name)
+      found = Parent.find_by(name: name)
       found.not_nil!.id.should eq model.id
 
-      found = Parent.find_by!("name", name)
+      found = Parent.find_by!(name: name)
       found.should be_a(Parent)
     end
 
@@ -26,10 +26,10 @@ module {{adapter.capitalize.id}}
       model.name = name
       model.save
 
-      found = Parent.find_by(:name, name)
+      found = Parent.find_by(name: name)
       found.not_nil!.id.should eq model.id
 
-      found = Parent.find_by!(:name, name)
+      found = Parent.find_by!(name: name)
       found.id.should eq model.id
     end
 
@@ -41,20 +41,20 @@ module {{adapter.capitalize.id}}
       model.all = value
       model.save
 
-      found = ReservedWord.find_by("all", value)
+      found = ReservedWord.find_by(all: value)
       found.not_nil!.id.should eq model.id
 
-      found = ReservedWord.find_by!(:all, value)
+      found = ReservedWord.find_by!(all: value)
       found.id.should eq model.id
     end
 
     it "returns nil or raises if no result" do
       Parent.clear
-      found = Parent.find_by("name", "xxx")
+      found = Parent.find_by(name: "xxx")
       found.should be_nil
 
-      expect_raises(Granite::Querying::NotFound, /Couldn't find .*Parent.* with name=xxx/) do
-        Parent.find_by!("name", "xxx")
+      expect_raises(Granite::Querying::NotFound, /Couldn't find .*Parent.* with name = xxx/) do
+        Parent.find_by!(name: "xxx")
       end
     end
   end
--- a/spec/granite/querying/find_by_spec.cr
+++ b/spec/granite/querying/find_by_spec.cr
@@ -18,22 +18,23 @@ module {{adapter.capitalize.id}}
       found.should be_a(Parent)
     end
 
-    it "finds an object with a symbol field" do
-      Parent.clear
-      name = "robinson"
+    it "works with multiple arguments" do
+      Review.clear
 
-      model = Parent.new
-      model.name = name
-      model.save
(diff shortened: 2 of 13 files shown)


src/granite/querying.cr
```diff
@@ -73,28 +73,26 @@ module Granite::Querying
...
+  # find returns the row with the primary key specified. Otherwise nil.
   def find(value)
-    return find_by(@@primary_name, value)
+    first("WHERE #{@@primary_name} = ?", value)
   end
 
+  # find returns the row with the primary key specified. Otherwise raises an exception.
   def find!(value)
-    return find_by!(@@primary_name, value)
+    first("WHERE #{@@primary_name} = ?", value) || raise Granite::Querying::NotFound.new("Couldn't find #{{{@type.name.stringify}}} with #{@@primary_name} = #{value}")
```
drujensen: You could call `find` instead of `first` here.

src/granite/querying.cr
```diff
@@ -73,28 +73,26 @@ module Granite::Querying
     first(clause, params) || raise NotFound.new("Couldn't find " + {{@type.name.stringify}} + " with first(#{clause})")
   end
 
-  # find returns the row with the primary key specified.
-  # it checks by primary by default, but one can pass
-  # another field for comparison
+  # find returns the row with the primary key specified. Otherwise nil.
   def find(value)
-    return find_by(@@primary_name, value)
+    first("WHERE #{@@primary_name} = ?", value)
```
drujensen: I think this is fine.  I can't see any downsides calling `first` instead of `find_by`.  In the end, the same query will be performed.

README.md
```diff
@@ -305,12 +305,13 @@ post = Post.find! 1 # raises when no records found
 #### Find By
 
 ```crystal
-post = Post.find_by :slug, "example_slug"
+post = Post.find_by(slug: "example_slug")
```
drujensen: I like this syntax better.  💯

Blacksmoke16:
@drujensen If you would, could you review latest commit?  I realized that the `select_one` was kinda irrelevant as i could just pass the query to `first` and get the same outcome.

spec/granite/querying/find_by_spec.cr
```diff
@@ -11,50 +11,51 @@ module {{adapter.capitalize.id}}
...
-      model.save
+      Review.create(name: "review1", upvotes: 2.to_i64)
+      Review.create(name: "review2", upvotes: 0.to_i64)
+      Review.create(name: "review1", upvotes: 10.to_i64)
 
-      found = Parent.find_by(:name, name)
-      found.not_nil!.id.should eq model.id
+      Review.find_by(name: "review1").not_nil!.id.should eq 1
+      Review.find_by(upvotes: 0).not_nil!.id.should eq 2
+      Review.find_by(name: "review1", upvotes: 10).not_nil!.id.should eq 3
```
robacarp: do you need `.not_nil!` here?

src/adapter/sqlite.cr
```diff
@@ -46,24 +46,6 @@ class Granite::Adapter::Sqlite < Granite::Adapter::Base
...
-
-    log statement, id
-
-    open do |db|
-      db.query_one? statement, id do |rs|
-        yield rs
-      end
-    end
-  end
-
```
robacarp: 👍 on removing these

src/granite/association_collection.cr
```diff
@@ -11,26 +11,23 @@ class Granite::AssociationCollection(Owner, Target)
     )
   end
 
-  def find_by(field : String | Symbol, value)
+  def find_by(**args)
```
robacarp: I like the new syntax, should the old overload stay around too?

src/granite/querying.cr
```diff
@@ -73,28 +73,24 @@ module Granite::Querying
...
   end
 
+  # find returns the row with the primary key specified. Otherwise raises an exception.
   def find!(value)
-    return find_by!(@@primary_name, value)
+    find(value) || raise Granite::Querying::NotFound.new("Couldn't find #{{{@type.name.stringify}}} with #{@@primary_name} = #{value}")
```
robacarp: This error message when rendered will read: `Couldn't find User with id = 24`. Could we make that one of:

`Couldn't find a(n) User with id = 24`
`No User found where id = 24`
`No User found with id = 24`

Similarly, for the messages generated find_by! and AssociationCollection#find_by!

src/granite/association_collection.cr
```diff
@@ -11,26 +11,23 @@ class Granite::AssociationCollection(Owner, Target)
...
     Target.first(
-      [query, "AND #{Target.table_name}.#{field} = ?"].join(" "),
-      [owner.id, value]
+      "#{query} AND #{args.map { |arg| "#{Target.table_name}.#{arg} = ?" }.join(" AND ")}",
```
robacarp: Should this table name and field names be quoted as well?

src/granite/association_collection.cr
Blacksmoke16: That was the original plan, but i didn't like the inconsistency between the two.  IMO id just keep the one and people can take a few min to change their statements to new format, which is better as well, as they would have to anyway to use the multiple criteria feature anyway.

src/granite/association_collection.cr
Blacksmoke16: Thanks

Blacksmoke16:
I ran the tests again locally and passed fine, must have been a fluke.

drujensen:
@robacarp can you approve?

faustinoaq:
@drujensen this is a breaking change, so, this shouldn't be included on `v0.11.1` but on `v0.12.0`, right? :)

drujensen:
@faustinoaq yes.  Marking at breaking change.  Thanks.

drujensen:
@faustinoaq @robacarp let's get this approved and merged.  I will make sure to release as 0.12.0

PR #214, opened 2018-05-22
Use DB.open block so it closes connection

### LINKED ISSUE #213: always more than max_prepared_stmt_count
add auto statements managing 

 
```
Can't create more than max_prepared_stmt_count statements (current value: 16382) (Exception)
  from lib/mysql/src/mysql/connection. cr:95:5 in 'handle_err_packet'
  from lib/mysql/src/mysql/connection.cr:112:5 in 'raise_if_err_packet'
  from lib/mysql/src/mysql/statement.cr:20:7 in 'initialize'
  from lib/mysql/src/mysql/statement.cr:4:3 in 'new'
  from lib/mysql/src/mysql/connection.cr:151:5 in 'build_prepared_statement'
  from lib/db/src/db/connection.cr:7:15 in 'fetch_or_build_prepared_statement'
  from lib/db/src/db/session_methods.cr:58:9 in 'build'
  from lib/db/src/db/pool_prepared_statement.cr:37:16 in 'build_statement'
  from lib/db/src/db/pool_prepared_statement.cr:53:22 in 'initialize'
  from lib/db/src/db/pool_prepared_statement.cr:11:5 in 'new'
  from lib/db/src/db/database.cr:89:7 in 'build_prepared_statement'
  from lib/db/src/db/database.cr:7:15 in 'fetch_or_build_prepared_statement'
  from lib/db/src/db/session_methods.cr:23:9 in 'build'
  from lib/db/src/db/query_methods.cr:38:7 in 'query'
  from lib/granite/src/granite/querying.cr:53:7 in 'raw_all'
  from ~procProc(Ar  from ~procProc(Array(ProductItemImage))@lib/granite/src/granite/querying.cr:64
Unhandled exception in spawn:
End of file reached (IO::EOFError)
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/io.cr:519:27 in 'read_fully'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/io/byte_format.cr:131:3 in 'decode'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/int.cr:489:5 in 'from_io'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/io.cr:895:5 in 'read_bytes'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/event/signal_handler.cr:47:7 in 'run'
  from ~procProc(Nil)@/usr/local/Cellar/crystal-lang/0.24.2_1/src/event/signal_handler.cr:93
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/fiber.cr:255:3 in 'run'
  from ~proc2Proc(Fiber, (IO::FileDescriptor | Nil))@/usr/local/Cellar/crystal-lang/0.24.2_1/src/fiber.cr:29
iber, (IO::FileDescriptor | Nil))@/usr/local/Cellar/crystal-la  from lib/granite/src/granite/collection.cr:255:3 in 'collection'
  from Granite::Collection(ProductItemImage)@Granite::Collection(M)#first?:(ProductItemImage | Nil)
  from lib/granite/src/granite/querying.cr:69:5 in 'first'
  from lib/granite/src/granite/querying.cr:68:3 in 'first'
  from db/seeds.cr:105:30 in '__crystal_main'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/crystal/main.cr:11:3 in '_crystal_main'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/crystal/main.cr:112:5 in 'main_user_code'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/crystal/main.cr:101:7 in 'main'
  from /usr/local/Cellar/crystal-lang/0.24.2_1/src/crystal/main.cr:135:3 in 'main'
```

Blacksmoke16:
Use DB.open block so that it will close connection when query is finished.  Resolves #213

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -5,11 +5,11 @@ require "db"
 # objects to perform actions against a specific database.  Each adapter needs
 # to implement these methods.
 abstract class Granite::Adapter::Base
-  property database : DB::Database
+  property database_url : String
 
   def initialize(adapter : String)
     if url = ENV["DATABASE_URL"]? || Granite.settings.database_url || replace_env_vars(settings(adapter)["database"].to_s)
-      @database = DB.open(url)
+      @database_url = url
     else
       raise "database url needs to be set in the config/database.yml or DATABASE_URL environment variable"
     end
@@ -28,7 +28,9 @@ abstract class Granite::Adapter::Base
   end
 
   def open(&block)
-    yield @database
+    DB.open(@database_url) do |db|
+      yield db
+    end
   end
 
   def log(query : String, params = [] of String) : Nil


Blacksmoke16:
https://github.com/amberframework/granite/commit/38dd6ffea35c1dd4b22cc1af9f943fac79300b51 Apparently you did :p

drujensen:
Oh, wow.  now I remember.  This was a major performance enhancement.  Well, darn.

Blacksmoke16:
Hmm is there a middleground? Say like giving a connection a set amount of time that it can be open so queries within x seconds/minutes or whatever can use the same connection until it is recycled?

faustinoaq:
@Blacksmoke16 I think you're right, maybe that's the reason why we're having [these errors](https://tfb-status.techempower.com/unzip/results.2018-05-20-08-15-39-766.zip/results/20180517221830/amber/run/amber.log)

We removed this before to enhance Granite performance, although maybe is too much :sweat_smile: 

WDYT?

Ref: https://tfb-status.techempower.com/results/ae097e87-3b23-402c-8f51-1608e16dd11a
Also see: https://tfb-status.techempower.com/

Blacksmoke16:
I'm sure it helped with performance; but having it unrestricted is almost worse.  I'll see if i can come up with a way to limit the requests per connection, 1000? 100?.  Should help with bursts of queries while never hitting the limits.

faustinoaq:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/C/h2o/start-servers.sh#L13

^^ Interesting @drujensen @Blacksmoke16 

^^ h2o (one of the fastest frameworks on TFB) uses 3 or 4 connections per request.

faustinoaq:
> No way, they are already too fast, it's unfair

^^ https://github.com/TechEmpower/FrameworkBenchmarks/pull/3751 :sweat_smile:

drujensen:
@Blacksmoke16 @faustinoaq Instead of closing the connection on every request which is quite slow, what if we try to figure out how to end the transaction or whatever is causing the prepared statements to not clear?

Also, the drivers have connection pooling built in so I don't think we need more than one.  Unless there is something blocking???

drujensen:
We may need help from the db driver team on this one.

bcardiff:
In crystal-db by default the user is able to create unlimited amount of connections and unless something goes wrong they are not closed.

Currently there is no life cycle policy about when an existing connection should be closed, the only pool options are [documented here](https://crystal-lang.org/docs/database/connection_pool.html).

After preparing a statements usually a handle is returned so the client can execute give the handler (and not the sql statement) and argument the query. That handler is per connection. But to allow statement reutilization a cache sql-query -> handle is needed per connection (or the user is forced to use statements rather than query methods from the pool).

If the amount of allocated resources per connection need to be reduced I see two options:

1. use unprepared statements. that is per driver bases but can be configured in the connection string.
2. work on a proposal for connection life cycle policy
3. `DB::Pool#clear` should clear all existing connection of the pool without dropping the whole pool, but I never tried that scenario
4. Use a single connection pool to create actually just a connection and not a pool.

drujensen:
@bcardiff very useful information

So is there any logic in the db driver that caches the prepared statement and reuse it if the same query is provided or are we currently always creating a new prepared statement for each call?  I realize this is per connection but it should only create the prepared statement once per connection.  This could be a huge performance increase if we are not properly using prepared statements.

Is there a way to clear the prepared statements without needing to close the connection?

Blacksmoke16:
I looked through the code and didn't see anything.  However this is not my forte, can put this PR on hold and see if we can come up with a better solution?

drujensen:
@bcardiff @Blacksmoke16  I looked through the db driver logic and it seems we are doing everything correctly according to the documentation.  I am going to close this and recommend that @jaysneg limits the pooling size documented (here)[https://crystal-lang.org/docs/database/connection_pool.html]

Blacksmoke16:
:+1: sounds good.

bcardiff:
@drujensen The related code for the prepared statements in the pool is in the [cache of each Connection](https://github.com/crystal-lang/crystal-db/blob/master/src/db/connection.cr#L40) and
how the [PoolPreparedStatement](https://github.com/crystal-lang/crystal-db/blob/f4c1c3626b4b303a432c309474709023db29a16b/src/db/pool_prepared_statement.cr#L35) try to reuse connections if possible where the statement was already prepared. All this happens without any configuration.

Issue #200, opened 2018-05-11
SELECT feature


jaysneg:
add ability to edit SELECT of Query

Is not posible to add custom fields

SELECT *, **categories.name AS category_name** FROM products JOIN categories ON categories.id = products.category_id

like there: **https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/select**

noahlh:
This is under discussion @ #184 

Working on it now!

robacarp:
Closing as duplicate of #184

Issue #184, opened 2018-04-24
Support for raw SQL queries


noahlh:
Proposal for discussion:

One of the things I really like about Granite is the philosophy that you should be able to get into the nitty gritty of SQL and let the ORM handle only the task of mapping the results to the object.  The `all` method does a great job of this.  

**I'd like to propose abstracting one level further and adding a method called `sql` (or maybe `raw`?) which allows a full SQL query to be passed in and simply maps the results to the appropriate fields of a model.**

An example based on an issue I ran into the the other night dealing with a legacy database in a project I'm building:

Assume I have a table called `vehicles` as such:

````
-----------------------------------------------
|  vehicleid  |  year  |  makeid  |  modelid  |
-----------------------------------------------
|     1       |  1999  |    54    |    65     |
|     2       |  2000  |    55    |    72     |
|     ...     |   ...  |    ...   |    ...    |
-----------------------------------------------
````

`makeid` and `modelid` are both references to foreign keys in the `makes` and `models` tables respectively.  In those tables are naming columns (called `makename` and `modelname`).

If I want to build a single JOIN query to pull in a vehicle's year, make, and model names, I can construct the following SQL to do so:

````sql
SELECT vehicles.vehicleid, vehicles.yearid, makes.makename AS make, models.modelname AS model FROM vehicles JOIN make ON vehicles.makeid = makes.makeid JOIN model ON vehicles.modelid = models.modelid
````

There's no way currently to have Granite construct that query -- the JOIN part could be done manually, but by default the SELECT portion can only be constructed to access the fields from a single table (the model's table).

In my proposal, you could have a model as such:

````crystal
class Vehicle < Granite::ORM::Base
  adapter pg

  primary vehicleid : Int32
  field yearid : Int32
  field make : String
  field model : String
end
````

And submit a query as such:

````crystal
vehicles = Vehicle.sql("SELECT vehicles.vehicleid, vehicles.yearid, makes.makename AS make, models.modelname AS model FROM vehicles JOIN make ON vehicles.makeid = makes.makeid JOIN model ON vehicles.modelid = models.modelid")
````

With the resulting `vehicles` object, you could then access the respective fields as you would normally, i.e. 

````crystal
vehicles.each do |v|
  puts "This is a #{v.yearid} #{v.make} #{v.model}"
end
````

I've implemented a monkey-patched version of this in my own application as follows (PG only, for starters) 

````crystal
class Granite::Adapter::Pg < Granite::Adapter::Base
[...]

PR #215, opened 2018-05-23
added query macro, tests, and docs

noahlh:
This PR (#184 & #200) allows you to customize the SQL used in the SELECT portion of the query (or the entire query if necessary) by adding the `query` macro.

The major limitation is that when using a highly-modified `query`, the save/update functionality of the model will likely break.  In the issue @drujensen suggested taking a more complete View Model approach, and I agree with that entirely, but wanted to start with a slightly narrower scope.  So this is "View Model Lite".  Baby steps.  Likewise, `find` may also break depending on how the query is modified.

Also, naming is hard.  I'm not wedded to any of the names used, and particularly with things like Query, Select, etc., things might get complex.

(PS - Please feel free to rip to shreds :) I'm new to this type of contribution so am eager to learn & improve in both style and substance.  Happy to make as many changes as needed.)

### DIFF
--- a/src/adapter/sqlite.cr
+++ b/src/adapter/sqlite.cr
@@ -27,14 +27,19 @@ class Granite::Adapter::Sqlite < Granite::Adapter::Base
     end
   end
 
-  # select performs a query against a table.  The table_name and fields are
-  # configured using the sql_mapping directive in your model.  The clause and
-  # params is the query and params that is passed in via .all() method
-  def select(table_name, fields, clause = "", params = [] of DB::Any, &block)
-    statement = String.build do |stmt|
-      stmt << "SELECT "
-      stmt << fields.map { |name| "#{quote(table_name)}.#{quote(name)}" }.join(", ")
-      stmt << " FROM #{quote(table_name)} #{clause}"
+  # select performs a query against a table.  The query object containes table_name,
+  # fields (configured using the sql_mapping directive in your model), and an optional
+  # raw query string.  The clause and params is the query and params that is passed
+  # in via .all() method
+  def select(query : Granite::Select::Container, clause = "", params = [] of DB::Any, &block)
+    if query.custom.size > 0
+      statement = query.custom
+    else
+      statement = String.build do |stmt|
+        stmt << "SELECT "
+        stmt << query.fields.map { |name| "#{quote(query.table_name)}.#{quote(name)}" }.join(", ")
+        stmt << " FROM #{quote(query.table_name)} #{clause}"
+      end
     end
 
     log statement, params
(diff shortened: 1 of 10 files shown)


src/granite/select.cr
```diff
@@ -0,0 +1,21 @@
+module Granite::Select
+  struct Container
```
robacarp: Could this be called a fieldset?

src/granite/select.cr
```diff
@@ -0,0 +1,21 @@
...
+    property custom
+    getter table_name, fields
```
robacarp: Since you’re target is to support multiple tables, shouldn’t each field have a paired table?

drujensen: @robacarp there may be no tables or fields that are calculated.  When this is a custom query, the table_name will not have any meaning.

What I think may be a bigger issue is all of the transactional methods are still included when a custom query is provided.  These methods should be excluded like `save` and `destroy` or at least throw runtime exceptions if you attempt to use them.  But it would be better to catch these at compile time.

src/granite/select.cr
noahlh: It's not so much a set of fields as it is a (loose) container for the query (or SELECT) statement.  `Granite::Query` is used in your new query builder, so I tried to capture this as simply as possible.  I don't love `Granite::Select::Container` though.  Open to ideas.

src/granite/select.cr
noahlh: @drujensen Agreed.  I think the next step (and I'll write up an issue to discuss this) is to go with your original suggestion in #184 and fully break this out into a ViewModel, which will eliminate the transactional methods (and, hence, no need for compiler check).

src/adapter/sqlite.cr
```diff
@@ -47,14 +52,19 @@ class Granite::Adapter::Sqlite < Granite::Adapter::Base
...
 
   # select_one is used by the find method.
-  def select_one(table_name, fields, field, id, &block)
-    statement = String.build do |stmt|
-      stmt << "SELECT "
-      stmt << fields.map { |name| "#{quote(table_name)}.#{quote(name)}" }.join(", ")
-      stmt << " FROM #{quote(table_name)}"
-      stmt << " WHERE #{quote(field)}=:id LIMIT 1"
+  def select_one(query : Granite::Select::Container, field, id, &block)
+    if query.custom.size > 0
```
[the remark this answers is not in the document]
noahlh: Great - thank you for suggestions.  Done and done.

src/granite/select.cr
robacarp: Sure `Container` is a little vague, but is it going to show up in an Amber devs code? If not, I don't think it's important to be 100% on target for merging this.

noahlh: Makes sense.  No, the only dev interface is the `query` macro, so not a big deal here.

PR #220, opened 2018-05-30
Fix last insert ID race condition

charlie-hadden:
First of all, thanks for the great library!

While working on an application using Granite we noticed a race condition with the IDs being returned for instances. In our case, it was two concurrent API requests to a create API - both records were being returned with the same ID.

The failing spec I've added here is a minimal example to reproduce the issue. We discovered it using postgres but it also appears to occur for mysql. I've included fixes for both adapters.

Strangely, the issue doesn't seem to occur with sqlite any time I have run the specs. I'm not sure if this is coincidence, or if some difference in the driver prevents the fiber from yielding in the same way. For now I've left the adapter untouched, but it may be more correct to wrap it in a `using_connection` block the same way I have with mysql.

Thanks for all of your work on this library so far.

Charlie

### DIFF
--- a/src/granite/transactions.cr
+++ b/src/granite/transactions.cr
@@ -66,14 +66,14 @@ module Granite::Transactions
       end
       begin
         {% if primary_type.id == "Int32" %}
-          @{{primary_name}} = @@adapter.insert(@@table_name, fields, params, lastval: true).to_i32
+          @{{primary_name}} = @@adapter.insert(@@table_name, fields, params, lastval: "{{primary_name}}").to_i32
         {% elsif primary_type.id == "Int64" %}
-          @{{primary_name}} = @@adapter.insert(@@table_name, fields, params, lastval: true)
+          @{{primary_name}} = @@adapter.insert(@@table_name, fields, params, lastval: "{{primary_name}}")
         {% elsif primary_auto == true %}
           {% raise "Failed to define #{@type.name}#save: Primary key must be Int(32|64), or set `auto: false` for natural keys.\n\n  primary #{primary_name} : #{primary_type}, auto: false\n" %}
         {% else %}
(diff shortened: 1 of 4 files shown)


src/granite/transactions.cr
```diff
@@ -66,14 +66,14 @@ module Granite::Transactions
...
+          @{{primary_name}} = @@adapter.insert(@@table_name, fields, params, lastval: "{{primary_name}}").to_i32
         {% elsif primary_type.id == "Int64" %}
-          @{{primary_name}} = @@adapter.insert(@@table_name, fields, params, lastval: true)
+          @{{primary_name}} = @@adapter.insert(@@table_name, fields, params, lastval: "{{primary_name}}")
         {% elsif primary_auto == true %}
           {% raise "Failed to define #{@type.name}#save: Primary key must be Int(32|64), or set `auto: false` for natural keys.\n\n  primary #{primary_name} : #{primary_type}, auto: false\n" %}
         {% else %}
           if @{{primary_name}}
-            @@adapter.insert(@@table_name, fields, params, lastval: false)
+            @@adapter.insert(@@table_name, fields, params, lastval: nil)
```
robacarp: why is this changed from false to nil?

faustinoaq:
@drujensen I think this fix is very nice, maybe we should consider it before releasing v0.12.0 :sweat_smile:

src/granite/transactions.cr
drujensen: It looks like `lastval` is being changed from a `Bool` to a `String?` to pass the name of the primary key.  Maybe we should rename that field?

charlie-hadden: @drujensen that's exactly it. If there's another name you'd rather I use then I would be happy to update it.

PR #221, opened 2018-06-03
Support model.from_json(JSON::Any)

### LINKED ISSUE #170: Error with normal constructor with JSON.mapping
Initializing a new Granite model using Model.new(args), with a JSON.mapping in the model does not work.

```Crystal
     JSON.mapping(
        type_id: Int32,
        type_name: String
     )

      adapter mysql
      primary type_id : Int32, auto: false
      field type_name : String
    end
```

Doing `Item.new(type_id: 1, type_name: "item1")` produces:

```
Error in src/api2.cr:20: missing argument: __temp_42
Overloads are:
 - Item.new(__temp_42 : ::JSON::PullParser)

p Item.new(type_id: 111, type_name: "Rhea")
                     ^~~
```

The error can be corrected by adding `def initialize(@type_id, @type_name) end` but that seems not ideal.

It would be even better if the defining of the fields/primary using granite notation would also allow for `Model.from_json` since they are basically the same definitions but in different formats.

Blacksmoke16:
This PR resolves #170 by including support for model.new(JSON::Any)

I just added another layer of ternary that checks of the value is of the `JSON::Any` type, if it is then use the `JSON::Any` .as_x to cast it to proper type.

### DIFF
--- a/src/granite/fields.cr
+++ b/src/granite/fields.cr
@@ -1,7 +1,7 @@
 require "json"
 
 module Granite::Fields
-  alias Type = JSON::Type | DB::Any
+  alias Type = JSON::Type | DB::Any | JSON::Any
   TIME_FORMAT_REGEX = /\d{4,}-\d{2,}-\d{2,}\s\d{2,}:\d{2,}:\d{2,}/
 
   macro included
@@ -101,7 +101,7 @@ module Granite::Fields
       end
     end
 
-    def set_attributes(args : Hash(String | Symbol, Type))
+    def set_attributes(args : Hash(String | Symbol, Type) | JSON::Any)
       args.each do |k, v|
         cast_to_field(k, v.as(Type))
       end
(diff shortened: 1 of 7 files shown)


Blacksmoke16:
Unfortunately this still does not work.  might be able to create a wrapper function that will instantiate multiple classes and return an array of them to replicate the `Array(Klass).from_json(ARRAY)`

```Crystal
class Foo < Granite::Base
  adapter mysql
  table_name foos

  primary type_id : Int64, auto: false
  field capacity : Float32
end
```

```crystal
  p Array(Foo).new(JSON.parse(%([{"type_id": 1, "capacity": 69.0},{"type_id": 2, "capacity": 5.0},{"type_id": 3, "capacity": 100.0}])))
```

src/granite/fields.cr
```diff
@@ -120,30 +120,30 @@ module Granite::Fields
...
-                @{{PRIMARY[:name]}} = value.as({{PRIMARY[:type]}})
+                @{{PRIMARY[:name]}} = value.is_a?(JSON::Any) ? value.raw.as({{PRIMARY[:type]}}) : value.as({{PRIMARY[:type]}})
               {% end %}
               return
             end
 
             return @{{_name.id}} = nil if value.nil?
             {% if type.id == Int32.id %}
-              @{{_name.id}} = value.is_a?(String) ? value.to_i32(strict: false) : value.is_a?(Int64) ? value.to_i32 : value.as(Int32)
+              @{{_name.id}} = value.is_a?(JSON::Any) ? value.as_i : value.is_a?(String) ? value.to_i32(strict: false) : value.is_a?(Int64) ? value.to_i32 : value.as(Int32)
```
drujensen: man, this is some true ugliness on my part.  Maybe someday we can figure out how to clean this up.

Blacksmoke16:
@robacarp it should be clarified that the `from_json` implemented here is _not_ the same one as Crystals `Array(Model).from_json`

robacarp:
@Blacksmoke16 yep. Since there are two approvals, are you ready for this to merge?

Blacksmoke16:
I probably should add some info to the README.  I'll do that tonight and can merge it tomorrow.

drujensen:
@Blacksmoke16 let me know when you are ready.

Blacksmoke16:
Readme is updated.  Should be good to go @drujensen

robacarp:
@Blacksmoke16 thank you!

PR #228, opened 2018-06-15
Update to crystal 0.25.0

### LINKED ISSUE #226: Get ready for Crystal 0.25.0
With 0.25.0 coming soon, we need to update some things for the new version.

- [x]  Update Dockerfile since `crystal deps` is being removed. https://github.com/crystal-lang/crystal/pull/5544
- [ ]  Remove references of `JSON::Type`. https://github.com/crystal-lang/crystal/pull/5183
- [ ] `File.stat` is renamed to `File.info` https://github.com/crystal-lang/crystal/pull/5584
- [ ]  Read the changelog for anything I missed. https://github.com/bcardiff/crystal/blob/changelog/0.25.0/CHANGELOG.md

faustinoaq:
Minimally addresses issues required to update Granite to Crystal 0.25.0

- [x]  Update Dockerfile since `crystal deps` is being removed. https://github.com/crystal-lang/crystal/pull/5544
- [x]  Remove references of `JSON::Type`. https://github.com/crystal-lang/crystal/pull/5183
- [x] `File.stat` is renamed to `File.info` https://github.com/crystal-lang/crystal/pull/5584
- [x] Update all instances of `Time::Kind` to use the new `Time::Location` instead

Read the changelog for anything I missed. https://github.com/bcardiff/crystal/blob/changelog/0.25.0/CHANGELOG.md

Fixes #226

### DIFF
--- a/src/granite/querying.cr
+++ b/src/granite/querying.cr
@@ -23,7 +23,7 @@ module Granite::Querying
             if @@adapter.class.name == "Granite::Adapter::Sqlite"
               # sqlite3 does not have timestamp type - timestamps are stored as str
               # will break for null timestamps
-              self.\{{name.id}} = Time.parse(result.read(String), "%F %X" )
+              self.\{{name.id}} = Time.parse(result.read(String), Granite::DATETIME_FORMAT)
             else
               self.\{{name.id}} = result.read(Union(\{{type.id}} | Nil))
             end
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,10 @@ RUN apt-get update -qq && apt-get install -y --no-install-recommends libpq-dev l
 
 WORKDIR /app/user
 
-ADD . /app/user
-
+COPY shard.yml shard.lock ./
 RUN shards install
+
+COPY . /app/user
 
 CMD ["crystal", "spec"]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /app/user
 
 ADD . /app/user
 
-RUN crystal deps
+RUN shards install
 
 CMD ["crystal", "spec"]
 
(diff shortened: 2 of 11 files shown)


Blacksmoke16:
@faustinoaq We probably should also address other changes specified in https://github.com/amberframework/granite/issues/226.

src/granite/querying.cr
```diff
@@ -23,7 +23,7 @@ module Granite::Querying
             if @@adapter.class.name == "Granite::Adapter::Sqlite"
               # sqlite3 does not have timestamp type - timestamps are stored as str
               # will break for null timestamps
-              self.\{{name.id}} = Time.parse(result.read(String), "%F %X" )
+              self.\{{name.id}} = Time.parse(result.read(String), Granite::DATETIME_FORMAT)
```
drujensen: nice change to `Granite::DATETIME_FORMAT` 💯

damianham:
why not rescue the exception for existing time strings and try to parse with "%F %X" ?

faustinoaq:
> why not rescue the exception for existing time strings and try to parse with "%F %X" ?

Interesting idea, Would this solve the SQLite issue?

^^ @robacarp @drujensen

drujensen:
I think this might work with one caveat:

We either need to store the time as UTC as a convention or we need to include the timezone in the stored string. You should always store the timezone implicitly or explicitly to get the right results.

I prefer we store the timezone in the string explicitly.  The advantage is you can determine the timezone that was originally used by the user/server when it was originally stored.  If you convert it to UTC before storing, you lose that information.

robacarp:
I agree, it makes sense for Granite to store the timezone as provided in the Time object given. In the event that it makes more sense for an application timezone to always be UTC, the appdev can make that happen easy enough.

PR #231, opened 2018-06-16
Further update Granite to be compatible with Crystal 0.25

robacarp:
The latest granite tag produces this when attempting to build amber projects which used the scaffold generator:

```
instantiating 'PostCommentController#create()'
in src/controllers/post_comment_controller.cr:22: no overload matches 'PostComment.new' with type Hash(String, String | Nil)
Overloads are:
 - Granite::Base.new(args : Hash(Symbol | String, String | JSON::Any))
 - Granite::Base.new(**args : Object)
 - Granite::Base.new()

    post_comment = PostComment.new(post_comment_params.validate!)
```

In trying to get the amber project compatible with crystal 0.25, I'd like to change this to JSON::Any::Type, which allows the generated code to compile.

### DIFF
--- a/src/adapter/mysql.cr
+++ b/src/adapter/mysql.cr
@@ -81,9 +81,9 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
       model_array.each do |model|
         model.set_timestamps
         next unless model.valid?
-        stmt << '('
+        stmt << "("
         stmt << Array.new(fields.size, '?').join(',')
-        params.concat fields.map { |field| model.to_h[field] }
+        params.concat fields.map { |field| model.read_attribute field }
         stmt << "),"
       end
     end.chomp(',')
--- a/src/adapter/mysql.cr
+++ b/src/adapter/mysql.cr
@@ -70,7 +70,6 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
 
   def import(table_name : String, primary_name : String, auto : String, fields, model_array, **options)
     params = [] of DB::Any
-    now = Time.utc_now
 
     statement = String.build do |stmt|
       stmt << "INSERT"
@@ -80,8 +79,7 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
       stmt << ") VALUES "
 
       model_array.each do |model|
-        model.updated_at = now if model.responds_to? :updated_at
-        model.created_at = now if model.responds_to? :created_at
+        model.set_timestamps
         next unless model.valid?
         stmt << '('
         stmt << Array.new(fields.size, '?').join(',')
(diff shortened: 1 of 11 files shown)


eliasjpr:
I have asked in the Crystal channel about `JSON::Any` and `JSON::Any::Type` I think the later makes more sense to keep things compatible. Not sure waiting for a response from the Crystal channel

eliasjpr:
I think we should be using `JSON::Any::Type` everwhere we had `JSON::Type` since both points to a union type

robacarp:
I’d discourage tagging after this is merged until it’s known that no other changes need to be made to Granite to make amber compatible

eliasjpr:
To be safe we should probably use `def initialize(args : Hash(Symbol | String, JSON::Any | JSON::Any::Type))` this will support the new way and the previous way

src/granite/base.cr
```diff
@@ -43,7 +43,7 @@ class Granite::Base
     set_attributes(args.to_h)
   end
 
-  def initialize(args : Hash(Symbol | String, String | JSON::Any))
+  def initialize(args : Hash(Symbol | String, JSON::Any::Type))
```
eliasjpr: Lets define a initialize overload, I think it would be better
```crystal
def initialize(args : Hash(Symbol | String, JSON::Any::Type))
  new(JSON.parse(args))
end

def initialize(args : Hash(Symbol | String, JSON::Any))
end
```

src/granite/base.cr
```diff
@@ -43,7 +43,15 @@ class Granite::Base
     set_attributes(args.to_h)
   end
 
-  def initialize(args : Hash(Symbol | String, String | JSON::Any))
+  def initialize(args : Hash(Symbol | String, String?))
+    set_attributes(args)
+  end
+
+  def initialize(args : Hash(Symbol | String, Granite::Fields::Type))
```
eliasjpr: 👍

src/adapter/mysql.cr
```diff
@@ -81,9 +81,9 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
       model_array.each do |model|
         model.set_timestamps
         next unless model.valid?
-        stmt << '('
+        stmt << "("
         stmt << Array.new(fields.size, '?').join(',')
-        params.concat fields.map { |field| model.to_h[field] }
+        params.concat fields.map { |field| model.read_attribute field }
```
robacarp: cc @Blacksmoke16

eliasjpr:
👍🏻

src/adapter/mysql.cr
Blacksmoke16: neat, i like that better.

PR #232, opened 2018-06-17
Adding transaction raise methods

Thellior:
<h1>What is done</h1>
Added "helper" methods that raises exceptions on errors. The idea is that sometimes you want to be sure that an transaction method has succeeded. If the method did not complete successfully than it will raise an exception.

<h1>Methods</h1>

- Added update method that accepts args that will set the attributes of the model and tries to save it
- Added update! method that accepts args that will set the attribute of the model and tries to save! it
- Added save! method that will call the save method and raises an exception if it failed
- Added create! method that will call the create method and raises an exception if it failed
- Added destroy! method that will call the destroy method and raises an exception if it failed
- Added custom exceptions for the methods so an application developer is able to catch specific errors

### DIFF
--- a/spec/granite/transactions/create_spec.cr
+++ b/spec/granite/transactions/create_spec.cr
@@ -72,5 +72,19 @@ module {{adapter.capitalize.id}}
       end
     end
   end
+
+  describe "{{ adapter.id }} .create!" do
+    it "creates a new object" do
+      parent = Parent.create!(name: "Test Parent")
+      parent.id.should_not be_nil
+      parent.name.should eq("Test Parent")
+    end
+
+    it "does not create an invalid object but raise an exception" do
+      expect_raises(Granite::Transactions::TransactionFailed) do
+        parent = Parent.create!(name: "")
+      end
+    end
+  end
 end
 {% end %}
--- a/spec/granite/transactions/create_spec.cr
+++ b/spec/granite/transactions/create_spec.cr
@@ -81,7 +81,7 @@ module {{adapter.capitalize.id}}
     end
 
     it "does not create an invalid object but raise an exception" do
-      expect_raises(Granite::Transactions::TransactionFailed) do
+      expect_raises(Granite::Transactions::TransactionError) do
         parent = Parent.create!(name: "")
       end
     end
--- a/spec/granite/transactions/create_spec.cr
+++ b/spec/granite/transactions/create_spec.cr
@@ -5,13 +5,13 @@ module {{adapter.capitalize.id}}
   describe "{{ adapter.id }} .create" do
     it "creates a new object" do
       parent = Parent.create(name: "Test Parent")
-      parent.id.should_not be_nil
+      parent.persisted?.should be_true
       parent.name.should eq("Test Parent")
     end
 
     it "does not create an invalid object" do
       parent = Parent.create(name: "")
-      parent.id.should be_nil
+      parent.persisted?.should be_false
     end
 
     it "takes JSON::Any" do
@@ -51,7 +51,7 @@ module {{adapter.capitalize.id}}
     describe "with a custom primary key" do
       it "creates a new object" do
         school = School.create(name: "Test School")
-        school.custom_id.should_not be_nil
+        school.persisted?.should be_true
         school.name.should eq("Test School")
       end
     end
@@ -59,7 +59,7 @@ module {{adapter.capitalize.id}}
     describe "with a modulized model" do
       it "creates a new object" do
         county = Nation::County.create(name: "Test School")
-        county.id.should_not be_nil
+        county.persisted?.should be_true
         county.name.should eq("Test School")
       end
     end
@@ -76,12 +76,12 @@ module {{adapter.capitalize.id}}
   describe "{{ adapter.id }} .create!" do
     it "creates a new object" do
       parent = Parent.create!(name: "Test Parent")
-      parent.id.should_not be_nil
+      parent.persisted?.should be_true
       parent.name.should eq("Test Parent")
     end
 
     it "does not create an invalid object but raise an exception" do
-      expect_raises(Granite::Transactions::TransactionError) do
+      expect_raises(Granite::RecordInvalid, "{{adapter.capitalize.id}}::Parent") do
         parent = Parent.create!(name: "")
       end
     end
--- a/spec/granite/transactions/create_spec.cr
+++ b/spec/granite/transactions/create_spec.cr
@@ -80,8 +80,8 @@ module {{adapter.capitalize.id}}
       parent.name.should eq("Test Parent")
     end
 
-    it "does not create an invalid object but raise an exception" do
-      expect_raises(Granite::RecordInvalid, "{{adapter.capitalize.id}}::Parent") do
+    it "does not save but raise an exception" do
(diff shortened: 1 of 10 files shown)


src/granite/transactions.cr
```diff
@@ -1,4 +1,6 @@
 module Granite::Transactions
+  class TransactionFailed < Exception; end
```
eliasjpr: `TransactionError` instead

src/granite/transactions.cr
```diff
@@ -11,6 +13,18 @@ module Granite::Transactions
...
     end
 
+    def create!(**args)
+      create!(args.to_h)
+    end
+
+    def create!(args : Hash(Symbol | String, DB::Any) | JSON::Any)
+      instance = create(args)
+
+      raise Granite::Transactions::TransactionFailed.new("Could not create #{self.name}") unless instance.errors.empty?
```
eliasjpr: This line is very long also I dont think you need to specify `Granite::Transactions::`

faustinoaq: Yeah, looks like Java :rofl:

Thellior: Ok, i worked on validation helper last time and all methods were one liners so i removed my multiline if statement for this one. But i will change it. I will remove the prefix module names

src/granite/transactions.cr
```diff
@@ -11,6 +13,20 @@ module Granite::Transactions
...
     end
 
+    def create!(**args)
+      create!(args.to_h)
+    end
+
+    def create!(args : Hash(Symbol | String, DB::Any) | JSON::Any)
+      instance = create(args)
+
+      if !instance.errors.empty?
```
robacarp: This would read better as `if instance.errors.any?`

src/granite/transactions.cr
```diff
@@ -1,4 +1,6 @@
 module Granite::Transactions
+  class TransactionError < Exception; end
```
robacarp: Do we really want to create a custom exception class for this? I doubt it, but I'd like to hear from @drujensen on this.

If so, it needs to be named differently. This file is called 'transactions' but it doesn't yet do anything with the SQL concept of a transaction. It's not really a problem because the filename isn't presented to an appdev.

This exception will be handled and used directly by an appdev, and the name is misleading.

src/granite/transactions.cr
```diff
@@ -143,6 +159,11 @@ module Granite::Transactions
       true
     end
 
+
+    def save!
+      save || raise Granite::Transactions::TransactionError.new("Could not save #{self.name}")
```
robacarp: I like this "save || raise" format better than the previous "raise unless save", but as someone else said, the explicit prefix `Granite::Transactions` can be dropped from these.

spec/granite/transactions/create_spec.cr
```diff
@@ -72,5 +72,19 @@ module {{adapter.capitalize.id}}
       end
     end
   end
+
+  describe "{{ adapter.id }} .create!" do
+    it "creates a new object" do
+      parent = Parent.create!(name: "Test Parent")
+      parent.id.should_not be_nil
```
robacarp: These would read better as `parent.persisted.should be_true`

src/granite/transactions.cr
drujensen: >> Do we really want to create a custom exception class for this? I doubt it, but I'd like to hear from @drujensen on this.

@robacarp yeah, I think we do.  We have `Granite::Error` class to handle errors so I think having our own exception makes sense.  I would rename this to `Granite::Exception`.  Not sure I would place it here in this file.  Maybe in its own file `/src/granite/exception.cr`?

robacarp: I went [looking for the way ActiveRecord does this](http://api.rubyonrails.org/classes/ActiveRecord/Base.html#class-ActiveRecord::Base-label-Exceptions) and here's what I found.

- ActiveRecordError - Generic error class and superclass of all other errors raised by Active Record.
- RecordInvalid - raised by ActiveRecord::Base#save! and ActiveRecord::Base.create! when the record is invalid.
- RecordNotFound - No record responded to the ActiveRecord::Base.find method. Either the row with the given ID doesn't exist or the row didn't meet the additional restrictions. 
- StatementInvalid - The database server rejected the SQL statement. The precise error is added in the message.

After thinking on it, as an application developer, when I write the save! method, I want to be able to capture the specific exception that is going to be raised by save! and not other random exceptions.

I retract my statement about removing this custom exception. Instead I think that it should be a `Granite::InvalidRecord` exception. It's more helpful and explicit, and allows use of the save! / create! methods without opening up to a category of bugs about catching the wrong exception.

robacarp: ( Rails `destroy!` raises a  ActiveRecord::RecordNotDestroyed )

src/granite/exceptions.cr
```diff
@@ -0,0 +1,13 @@
+module Granite
+  class RecordInvalid < ::Exception
```
robacarp: Would this be more grammatically correct as "InvalidRecord"?

spec/granite/transactions/destroy_spec.cr
```diff
@@ -55,5 +55,37 @@ module {{adapter.capitalize.id}}
...
+      parent.name = "Test Parent"
+      parent.save
+
+      id = parent.id
+      parent.destroy
+      found = Parent.find id
+      found.should be_nil
+    end
+
+    it "does not destroy an invalid object but raise an exception" do
```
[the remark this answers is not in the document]
Thellior: I agree on that but i could not find an way to raise an database exception to verify this behaviour. Granite will generate the following query "DELETE FROM "parents" WHERE "id"=$1". If the record does not exist it will still run the query and succeed. So i could not raise an "DB::Error" so i was verify the behaviour with an "Granite::Callbacks::Abort" error. I verified that the delete query never runned. I will add an test that verifies that the record is still there

src/granite/exceptions.cr
Thellior: I followed the Rails convention you mentioned in the comment. So i introduced RecordNotDestroyed and RecordInvalid into the stack. Do you want me to change it to InvalidRecord instead?

spec/granite/transactions/destroy_spec.cr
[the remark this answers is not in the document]
robacarp: Great explanation. To be clear, I don't have any problem with the way you're triggering the error condition, I was just commenting on the description text of the test itself. :]

src/granite/exceptions.cr
robacarp: Generally with Amber I'm not as interested in following in Rails' footsteps as I am in producing something that feels natural. Sometimes that means features similar to Rails, sometimes not.

Thellior:
After discussion with @robacarp on gitter we decided that we are going to change "RecordInvalid" to "RecordNotSaved". We are going to add an model property on the exception so people who are catching the specific exception are having access to the instance.

eliasjpr:
@Thellior great work on this! 🥇

Thellior:
Can we merge this?

faustinoaq:
@drujensen Should we release granite v0.12.2 or v0.13.0? :sweat_smile:

Issue #180, opened 2018-04-10
Clarify usage of database_url string


Blacksmoke16:
1.  Add to readme that you can use `Granite::ORM.settings.database_url = "mysql://username:password@hostname:3306/database"` in code instead of ENV variable or the yaml file if so desired.

2.  Make a note it must be set **before** any models are declared (in the case of a single file).

robacarp:
Someone mentioned in the gitter and I agree, I don’t think Granite should be responsible for reading config files or environment variables anymore. Granite as a library should only care to receive its connection information via its config interface.

Blacksmoke16:
Wouldn't hurt to have it setup more like AR was in Ruby, where you can define different databases based on ENV, but general config was all in the yaml file.

robacarp:
It seems like reading the ENV/config are outside the bounds of this library though, and should be handled by Amber (or whatever application is using Granite).

elorest:
We do have conf files within amber for this. :)

Blacksmoke16:
I been thinking of this, here are my thoughts.

Granite should decide upon one format for supplying a connection, such as `Granite::ORM.settings.database_url =  xxx`, or like Granite::ORM.Settings.new(host, port, db, ...) or what have you.  Granite should accept this format for defining which db should be used for the queries.  

Whatever application/framework you are using can have their own method of storing the information required(such as in environment, or a yaml file, or a string etc).  Wherever/how the parameters come to be supplied to Granite is outside the scope of Granite.

drujensen:
Just to give some background, Granite was originally built in Kemalyst and the configuration was read from the `config/database.yml`.  When we migrated to Amber, we decided to include the database configuration in the environment (which totally makes sense) but I left the older style for people that was still using it.  

We can deprecate the `config/database.yml` and reading the environment stuff  and only support setting the `database_url` since its been over a year.

Blacksmoke16:
@robacarp @drujensen @faustinoaq 

I'm thinking this will be my next PR, after some of my current ones are merged/closed so that I dont get into branch soup.

My plan for this is to deprecate the ENV and .yml config options in favor of Granite::ORM.settings.database_url = xxx`.  First by removing their implementations, but keeping the logic that accepts them in the base.cr adapter.  If the `Granite::ORM.settings.database_url = xxx` is nil (meaning they used one of the other methods) throw an exception telling them to switch.  Probably keep this exception there for a while then remove it.

[...]

PR #236, opened 2018-06-18
Adds dev dep on envy for .env loading/parsing

robacarp:
Currently the docs say "source the .env file to add the variables to your shell". This makes that step unnecessary because envy does that automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,8 +762,7 @@ $ docker-compose down
 If you'd like to test without docker you can do so by following the instructions below:
 
 1. Install dependencies with `$ crystal deps`
-2. Update .env to use appropriate ENV variables, or create appropriate databases.
-3. Setup databases:
+2. Create appropriate databases:
 
 #### PostgreSQL
 
@@ -785,5 +784,4 @@ CREATE DATABASE granite_db;
 GRANT ALL PRIVILEGES ON granite_db.* TO 'granite'@'localhost' WITH GRANT OPTION;
 ```
 
-4. Export `.env` with `$ source .env`
-5. `$ crystal spec`
+3. `$ crystal spec`
--- a/README.md
+++ b/README.md
@@ -761,7 +761,7 @@ $ docker-compose down
 
 If you'd like to test without docker you can do so by following the instructions below:
 
-1. Install dependencies with `$ crystal deps`
+1. Install dependencies with `$ shards install`
 2. Create appropriate databases:
 
 #### PostgreSQL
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -762,8 +762,7 @@ $ docker-compose down
 If you'd like to test without docker you can do so by following the instructions below:
 
 1. Install dependencies with `$ crystal deps`
```
Blacksmoke16: This doesn't exist anymore.  Prob should update that as well.

Blacksmoke16:
We were talking a bit about this in #180.  The issue is that as of now there isn't a better way to handle supplying the connection strings for running the tests for granite itself (since it needs all 3), which this change affects.  

That issue kinda ballooned out of scope, but i will be making a PR this weekend that will at least throw a warning if someone is using ENV variables for their app, as we want to deprecate reading ENV variables in favor of just doing `Granite.settings.database_url = xxx`.

faustinoaq:
> That issue kinda ballooned out of scope, but i will be making a PR this weekend that will at least throw a warning if someone is using ENV variables/config file for their app, as we want to deprecate reading ENV variables/config files in favor of just doing Granite.settings.database_url = xxx.

@Blacksmoke16 So, What happens if I need to change my `database_url` on production, Do I need to recompile my app? :sweat_smile: 

Should I do this :point_down: then?

```cr
Granite.settings.database_url = ENV["DATABASE_URL"] if ENV["DATABASE_URL"]?
```

Blacksmoke16:
Well the idea is that you can still use ENV variables like `Granite.settings.database_url = ENV["DATABASE_URL"]` as you have.  But it shouldn't be Granite's responsibility to look for config files or read ENV variables directly.

eliasjpr:
@Blacksmoke16 I completely agree about it should not be Granite responsibility to look for environment I have raised this concern in the past. For that regard we should remove the `ENV["DATABASE_URL"]` from source code. :man_shrugging:

drujensen:
Just to provide history and perspective on this.  

Granite was following the same standard that Rails uses for reading the DATABASE_URL environment variable if it exists.  I did this to make it easy to use but I am the first to admit that DATABASE_URL has bit me in the butt on several occasions in Rails projects where I wasn't aware the ENV variable was set and it would override the `config/databsase.yml`. 

I agreed that we want to remove this but I preferred deprecating it instead of breaking old sites (specifically mine) ;-)  I have moved or shutdown these sites so I'm good with removing it now.  :-D

drujensen:
@robacarp Is this PR still valid?

robacarp:
@drujensen Yes and no, but I think until the adapter thing is resolved it can be closed.

PR #240, opened 2018-06-23
Add ameba

Thellior:
<h1> What is done</h1>
Adding ameba to the stack and enforced the code style changes coming with it


src/granite/table.cr
```diff
@@ -53,6 +53,10 @@ module Granite::Table
       @@primary_name
     end
 
+    def self.primary_type
+      @@primary_type
+    end
```
c910335: There is no such class variable.

src/granite/query/assemblers/postgresql.cr
```diff
@@ -60,7 +60,7 @@ module Granite::Query::Assembler
 
     def count : Executor::Value(Model, Int64)
       where = build_where
-      order = build_order(use_default_order = false)
+      order = build_order(false)
```
robacarp: should this be `use_default_order: false`?

docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '2'
 services:
   spec:
     build: .
-    command: 'bash -c "cd /app/user && sleep 10 && crystal spec"'
+    command: 'bash -c "cd /app/user && sleep 10 && bin/ameba && crystal spec"'
```
c910335: Also add `crystal tool format --check`?

src/granite/query/assemblers/postgresql.cr
Thellior: It is a local variable assignment that was never used in that function anymore. So it was redundant that's why it was complaining about the variable not being used

Thellior:
@robacarp I agree on you on the # rubocop:disable bla:bla everyone in ruby projects. We as the amber team should enforce that we do not allow that kind of statements inside the code. If you for example have an error stating that the method length is to long it probably really is. We should fix the method length instead of disabling the line. If we as an team fix this kind of issues than the static analysis could be really powerfull. What does the rest of the people think about this? @amberframework/contributors

src/granite/table.cr
Thellior: There was a macro variable assignment to primary_type = PRIMARY[:type]. It was not used anywhere so i made the method. The class variable was not declared and i will add it. Thanks for noticing

docker-compose.yml
Thellior: I added it to the docker compose file. I tried to run it in the command line on my own machine but it did not change any file. I tried to read to cli help but it just stated "crystal tool format". Also checking crystal documentation but could not find it. Am i doing something wrong or are all files already formatted correctly. If i change something to be really off that it still does nothing for me. Could you provide me more information regarding to the tool format functionality?

faustinoaq:
@Thellior We're already using ameba to check amber code, we even have a `.ameba.yml` file, see: https://github.com/amberframework/amber/blob/master/.ameba.yml

SO, I think this is fine :sweat_smile:

docker-compose.yml
c910335: Here is a good guide for [Writing Shards](https://crystal-lang.org/docs/guides/writing_shards.html#coding-style) in the official documentation and it mentioned `crystal tool format --check`.
Also you can see more formatter options in `crystal tool format --help`.

Thellior:
@faustinoaq Ofcourse i agree but i also think @robacarp has a valid point. If you are coming from Rails projects we often see allot of disables in the code instead of fixing the real issues there. So it's good that we address this matter and enforce a solution for it

src/adapter/mysql.cr
```diff
@@ -50,7 +50,7 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
       stmt << "INSERT INTO #{quote(table_name)} ("
       stmt << fields.map { |name| "#{quote(name)}" }.join(", ")
       stmt << ") VALUES ("
-      stmt << fields.map { |name| "?" }.join(", ")
+      stmt << fields.map { |_name| "?" }.join(", ")
```
veelenga: Why not just remove the block var?

```
fields.map { "?" }.join(", ")
```

src/adapter/sqlite.cr
```diff
@@ -52,7 +52,7 @@ class Granite::Adapter::Sqlite < Granite::Adapter::Base
       stmt << "INSERT INTO #{quote(table_name)} ("
       stmt << fields.map { |name| "#{quote(name)}" }.join(", ")
       stmt << ") VALUES ("
-      stmt << fields.map { |name| "?" }.join(", ")
+      stmt << fields.map { |_name| "?" }.join(", ")
```
veelenga: ditto

src/granite/query/assemblers/postgresql.cr
veelenga: In term of backward compatibility, replacing it with `false` is correct.

veelenga:
Hey there. Glad to find this thread. 

> Every project I’ve used rubocop on was littered with rubocop comments everywhere, which defeats the purpose and makes the source worse than if it weren’t there to begin with. Amber has already started down this dark road.

@robacarp I agree with that. I would suggest not to use inline comments everywhere. Actually, those directives are designed for the very specific cases:

1. preventing false negativeness (issue in Ameba)
1. design mistakes in third-party libraries
1. rare code noise/fix/hack we have to do and probably will fix it later

As you can see these directives are intended to be used in places similar to where `TODO` word is used. Once the issue is fixed (on Ameba's side, in a thirparty or in a code), Ameba will force you to remove the directive via [UnneededDisableDirective](https://veelenga.github.io/ameba/Ameba/Rule/Lint/UnneededDisableDirective.html) rule.

> What do you think is a good way to prevent this behavior?

1. In my opinion every project has to follow predefined style/lint rules across the full codebase. This is where Ameba should be useful. It is easy to define a list of rules for your project and if you disagree with something, you just disable it in `.ameba.yml` file for a whole project.

2. In coming release, Ameba will allow to run only a group of rules. For example, if you wouldn't want to follow any style, you would be able to just check yourself for Lint issues:

```
$ ameba --only Lint       # use only Lint rules
$ ameba --only Lint,Style # use only Lint and Style rules
```

3. Lint rules report code noise and don't enforce any style. However, if you disagree with those suggestions, you should not use Ameba at all.

Blacksmoke16:
I could go either way on this.  TBH it's prob a good thing.  If things get more complicated down the line we can worry about it then.

I like how most of those changes were from me....I guess i had a long day of camel casing PHP variables that day at work ha.

faustinoaq:
@veelenga BTW, I just realize crystal has some [`has_x?`](https://crystal-lang.org/api/0.25.0/Hash.html#has_key?(key)-instance-method) methods on stdlib? 

That's the reason why I added an `.ameba.yml` file to `Disallows tautological predicate names` :sweat_smile:

veelenga:
@faustinoaq crystal repo even violates [Coding style](https://crystal-lang.org/docs/conventions/coding_style.html) in a couple of places.

However, i'm going to disable `has_x?` and a couple of other Style rules by default.  Rules which irritate people more than provide help...

faustinoaq:
> crystal repo even violates Coding style sometimes.

Then we should use ameba on crystal repo as well :wink: @asterite @RX14 :sweat_smile:

asterite:
Coding style is formatter. If formatter accepts the code then it's good (and currently that's the case).

faustinoaq:
@asterite Well, code linting is also useful, by example ameba detects **some** unused arguments and vars, bad named variables, etc.

> Then we should use ameba on crystal repo as well 

@veelenga NVM, you already opened a PR here https://github.com/crystal-lang/crystal/pull/6055 :sweat_smile:

robacarp:
Thanks everyone for chiming in. I like the tone and goal here and agree with @Blacksmoke16 that we can address issues if/when they come up. Thanks @veelenga for ameba.

PR #241, opened 2018-06-23
Update database docker

Thellior:
<h1>What is done</h1>

- Changed mysql version from 5.6 to 5.7 (8.0 gives an error)
- Changed postgres version from 9.5 to 10.4


robacarp:
That’s a big upgrade for Postgres. I doubt we are doing anything that PG changed during the upgrades, but I wonder if we should start to consider testing Granite against multiple versions of the databases?

Thellior:
@robacarp I think that it's a great idea. I'm going to read some documents about how it's best achieved with dockers and i will come with my idea that we can approve :)

Blacksmoke16:
should just have to add another service to the docker-compose.

for example:
```YAML
  pg-10.4:
    image: postgres:10.4
  pg-9.5:
    image: postgres:9.5
  mysq-5.6:
    image: mysql:5.6
    environment:
      MYSQL_ROOT_PASSWORD: pass
      MYSQL_DATABASE: test
      MYSQL_USER: user
      MYSQL_PASSWORD: pass
  mysql:
    image: mysql:5.7
    environment:
      MYSQL_ROOT_PASSWORD: pass
      MYSQL_DATABASE: test
      MYSQL_USER: user
      MYSQL_PASSWORD: pass
```

PR #242, opened 2018-06-23
Refactor Granite::Settings to support multiple DB strings and remove ENV/config file reading

Blacksmoke16:
This PR is the first step towards the removal of Granite's ability to reading ENV and config files directly, as per #180 and #236.

This change logs a warning when either env or config file is used to provide the database_string, while still allowing the connection to work.  In a future update, we could raise an exception and force them to use `Granite.settings.database_url = xxx`.  

One thing i did notice that is slightly annoying is that the message gets logged for each model the user has, since a new instance of the adapter is getting created.


Blacksmoke16:
@robacarp How would that work?  One thing i realized was that the config file allowed you to define each adapter connection strings in one place and it would pick the one it needs for the specific model based on the adapter macro.

Which unless we refactored settings to like `Granite.settings.database_urls["mysql"] = xxx` you couldn't do with just one `Granite.settings.database_url`

Blacksmoke16:
@robacarp @eliasjpr If you would could you review this again/give your thoughts?

I refactored the settings `database_url` property to be a class property hash, so that urls for each adapter can be supplied.  I also addressed @eliasjpr suggestion of using `Granite::Settings`

I think this is a pretty good solution but WDYT?

robacarp:
I don't see the reason to make a breaking change here by switching to `Granite::Settings`. It'll require a minor-version update to Amber, and anyone maintaining an app will have to manually update code in at least two places, `config/granite.cr` and `config/spec_helper.cr`. It's a lot of churn for a style preference. (Obviously I disagree with the preference as well, my preference is that Granite.settings.database_url is fine)

src/adapter/base.cr
```diff
@@ -8,22 +8,10 @@ abstract class Granite::Adapter::Base
   property database : DB::Database
 
   def initialize(adapter : String)
-    if url = ENV["DATABASE_URL"]? || Granite.settings.database_url || replace_env_vars(settings(adapter)["database"].to_s)
+    if url = Granite::Settings.database_url[adapter]
```
robacarp: I think this is a good solution to the problem at hand, but I also think it can be made backwards compatible with a little more effort.

Is it possible to retain the _setter_ `Granite::Settings.database_url = 'postgres://u:p@h:5432/db'` syntax and make this happen behind the scenes? That would:

- not require a breaking-change release
- allow people to specify multiple adapter connection strings if needed (though only up to 3, so no connections to multiple postgres instances)

Blacksmoke16: I think that is a good solution.  Few things to note:
* If I provide three postgres strings, the last one would be the one that is used since they would just reset the previous ones?
* We would have to either rename some of adapter classes to match the connection string adapter, i.e. `pg -> postgres and sqlite -> sqlite3`, or have conditional logic to handle the change in keys

eliasjpr: @robacarp I disagree about making this backward compatible instead why not make the `Granite::Settings` can happen at a later PR where there are more breaking changes for a given release. This means that we would have to keep track of the backward compatibility code move any settings property to both places (more duplication) to later remove the dead code.

eliasjpr:
@robacarp moving to `Granite::Settings` is a refactor that needs to happen regardless, sooner or later. And to me is not a style preference, the current `Granite.settings.database_url` violates a couple OO principles, and while it is a **breaking change**, the required changes are in mere 2 places in any Amber app. I do not see this as a lot of churn for something that is done properly, we introduce breaking changes quite regularly. 

I rather have this in sooner than later, but not necessarily in the current PR

Blacksmoke16:
Updated things to support current `Granite.settings.database_url` which also supports setting the connection string for each adapter behind the scenes.  Only one connection string is allowed per adapter as additional ones would just override the old one.

eliasjpr:
@Blacksmoke16 ideally the `Granite::Settings.database_url` should just take a single connection string and it should be name ``Granite::Settings.connection_string`  this is more descriptive and on target to the purpose IMO.  Settings could have another method called `Granite::Settings.database_connection` this can define a connection by accepting a Hash or NamedTuple this can then support loading db connections from YAML files as Rails does. 

```crystal
{
host: config[Rails.env]["host"]
database: config[Rails.env]["database"]
username: config[Rails.env]["username"]
password:  config[Rails.env]["password"]
...etc..
}
```

Again I would keep this in a separate PR.

spec/spec_helper.cr
```diff
@@ -1,3 +1,7 @@
+Granite.settings.database_url = ENV["MYSQL_DATABASE_URL"]
```
eliasjpr: How does this work without overwriting the previous connection? :confused:

Blacksmoke16:
@eliasjpr I could go either way on the naming of the method.  They both are correct and fine imo.

However, im not sure I agree with the `database_connection` method.  This seems like it's getting back into what we are trying to get out of.  

Also, it'll be for a later PR.  But this PR solves the biggest issue of running tests w/o using ENV/config files.

src/granite/settings.cr
```diff
@@ -2,13 +2,25 @@ require "logger"
 
 module Granite
   class Settings
-    property database_url : String? = nil
-    property logger : Logger
+    property logger = Logger.new STDOUT
```
eliasjpr: :+1:

src/granite/settings.cr
```diff
@@ -2,13 +2,25 @@ require "logger"
...
+    getter adapters = {} of String => String
 
     def initialize
-      @logger = Logger.new STDOUT
       @logger.progname = "Granite"
     end
+
+    def database_url=(database_url : String)
```
eliasjpr: I see, while this is cool, I don't think we should do this, is confusion, reasons why is this is an odd behavior for an assignment method `=` this would be much better for an append method `<<`. And again 1 instance of Granite should manage 1 connection IMO.

spec/spec_helper.cr
Blacksmoke16: _magic_

https://github.com/Blacksmoke16/granite-orm/blob/database_url-deprecations/src/granite/settings.cr#L12-L23

the database_url= method is just a wrapper that in the backgrounds sets the given URL as the value with the adapter being the key.  Since each of those uses a different adapter, it saves all three and doesn't override anything.

src/granite/settings.cr
eliasjpr: Also what if someone plug in a MemSQL adapter to granite and wants to use it, this will not allow it since the method is generalized. :)

spec/spec_helper.cr
eliasjpr: Too much magic for me :( I'm boring :sweat_smile:

src/granite/settings.cr
Blacksmoke16: In most cases only 1 connection_url would be used, Granite itself needs all 3 for the specs, since the specs run for each adapter.

It would error saying that it's an invalid adapter.

eliasjpr: It seems that the spec should have an instance of Granite per Adapter, not the way is currently set up. So instead of defining a single spec helper with all 3 connections I will define a spec helper per adapter and require it per the corresponding adapter.

eliasjpr: > EDIT: Granite in the past supported multiple connections, but in the form of the database.yml config file. So it's not like this is anything new.

Thank you for the info, I didn't know this detail.

Blacksmoke16: Yea https://github.com/amberframework/granite/blob/master/config/database.yml

So how that worked was, code in Adapter::Base parsed that out and inserted the database_urls from ENV.  This way, when a new instance of an adapter was created from the `adapter` macro, it just picked the correct one.

eliasjpr:
@Blacksmoke16 I will wait on @robacarp and @drujensen weight in.

robacarp:
@Blacksmoke16 thanks for your work here. This ended up being kind of a land mine instead of the quick cleanup it seemed like 😒 

@eliasjpr and I spent some time talking about this and the goals that this borders on. 

Short term goals:

- Get the yaml parsing / environment var parsing out of granite
- Have tests that continue to work across all adapters
- Be able to specific connection parameters easily, perhaps with even more flexibility
- Provide room for forwards compatibility for long term goals

Long terms goals, in no particular order:

- Read slave support for select queries (requires tuning in the adapters to fetch the appropriate connection)
- Connections to multiple databases, even of the same adapter (requires some restructuring of the Granite::Base adapter DSL)

-----------

### RFC

After some spitballing, I'd like to propose this syntax:

```crystal
# Simple connection url, with adapter specified as protocol in a NamedTuple literal:
Granite.connections << { url: "postgres://etc:etc@etc:etc/etc" }

# Allow multiple adapters to be specified now, and connections are looked up according to adapter, to help with testing
Granite.connections << { url: "sqlite://file.sqlite" }
Granite.connections << { url: "mysql://databases@databases" }
```

Later, when developer time allows, the NamedTuple allows extra parameters to be added:

```crystal
# read slave connections
Granite.connections << { url: "postgres://etc:etc", read_only: true }

# Named connections for use in specifying in adapters, or possibly even sharded data
Granite.connections << { url: "database://", name: :user_database }
```

--------------

#### Reasoning

- Granite.connections `<<` instead of `=`: makes it obvious that `connections` (plural) is a collection of connection strings.
- NamedTuple allows: a terse syntax that is easy to read and write, and provides a data structure capable of handling more data than a simple string, including alternate connection information styles (username, host, password instead of url), specifying read-slaves, etc.

Thoughts?

drujensen:
@robacarp @eliasjpr When you have some time, would love to discuss this. Not following the multiple `connections`.  I think we should find a way to use inheritance to support connecting to different databases.
```
class Post << MysqlBase
end

class Comment << PgBase
end

class MysqlBase << Granite::Base
  adapter mysql
  database_url = Amber.secrets.mysql_database_url
end
...
```

This will allow connecting to any database and any flavor or connecting to the same flavor twice.  I don't see how the `connections` collection will support that.

robacarp:
@drujensen I could get behind recommending inheritance, but that means cementing a big refactor to Granite::Base - last I checked the finished macro doesn’t work well with inheritance at all. 

My thinking with connection list is that it’ll work the same as it does today, with the settings.database_url, just a little easier to step into letting granite connect to more than one database at once. Primarily for testing, but also to make room for other features of connections. 

@blacksmoke16 I’m neither here nor there about the naming. Pg and Postgres are basically synonyms in my vocabulary, and I don’t mind it one way or another

drujensen:
@robacarp I think inheritance is working except for the `inheritance` macro for validation.  I will look into it since I think inheritance is pretty important for cleaner code and right now, I find it very annoying.

I would like to hold off on this change since the collection idea doesn't make sense to me.  I can't figure out how a connection is associated with which model unless you change it to a hash and then somehow lookup the connection string using the db type.  Even then, if you have two different connections to a mysql db, you would need someway of determining which one the model should use.

eliasjpr:
> Even then, if you have two different connections to a MySQL DB, you would need some way of determining which one the model should use.

This is where most of my confusion resides. If we would like to add multiple connections feature then I think 2 things should happen:

1. Provide a class method `use_connection :some_name` to specify the connection name to use per model basis. 
2. I also do like the idea presented by @robacarp `Granite.connections << { name: "somename", url: ""}`this allows a lot of flexibility for setting up multiple connections. 

I do not agree with `database_url=` create a hash is the background with connections. I expect this method call to override a connection.

Also, after talking with @robacarp I think it is best to frop the `Settings` scope since there is only 2 settings at the moment, and simply add a `Granite.database_url=` this will simplify the code from `Granite.settings.database_url`, which I am not fond. 

let me know what you guys think. :) 

@Blacksmoke16 you have done great work with this PR and other PRs I have seen. Thank you! :raised_hands:

eliasjpr:
> @robacarp: @drujensen I could get behind recommending inheritance, but that means cementing a big refactor to Granite::Base - last I checked the finished macro doesn’t work well with heritage at all.

We can use modules for that matter which I think is more explicit

```crystal 
class User < Granite::Model
  include PGMasterConnection
end

module PGMasterConnection
  use_connection :postgres_master
end
```
I think it has the same effect and it stays away from the inheritance

robacarp:
After sleeping on it, I'm going to backtrack on inheritance. Inheritance seems like too valuable of an architectural tool for Granite to enforce for something as simple as connections. It could work, but it also creates an explicit link from connections to model hierarchy. That's probably fine in most cases, but when it's not it's going to be really difficult to unwind.

Regarding your question of how does a model choose a connection, I don't think the model should _need_ to choose anything, same as how it works now -- minimize boilerplate. The adapter compatible connection should be taken.

Here's some spitballing code to make my vision a little more clear:

####Framework changes

```crystal
module Granite
  @@connections = [] of Connection

  def self.connections<<(url : String)
    @@connections << Connection.new url: url
  end

  def self.connections
    @@connections
  end
end

class Granite::Connection
  def initialize(@url) : Adapter::Base.class
    @adapter = case @url
    when .starts_with? "postgres://"
      Adapter::Postgres
    when .starts_with? "mysql://"
      Adapter::Mysql
    when .starts_with? "sqlite://"
      Adapter::Sqlite
    else
      raise "Invalid connection url"
    end
  end
end

module Granite::Table
  macro adapter(name)
    @@adapter = Granite::Adapter::{{name.id.capitalize}}.new

    def self.adapter
      @@adapter
    end
  end
end

abstract class Granite::Adapter::Base
  def initialize(adapter : String)
    @connection = Granite.connections.select do |connection|
      connection.adapter == self.class
    end.first

    unless @database = DB.open @connection
      raise "No viable connection provided to Granite"
    end
  end
end
```

#### Application usage:
```crystal
# in some config/setup routine
Granite.connections << { url: "postgres://localhost/database" }

# a Model, unchanged
class MyModel < Granite::Base
  adapter pg
end
```

eliasjpr:
How would you choose from different PG connections with this approach? Also, if you pass a connection name instead it simplyfies this case statement. I don't understand why we need to have this logic branch in the source code just to make the test work :-1: the tests should be refactor to run each Adapter independently from one another then this logic can be remove. IMHO :sweat_smile: 

```crystal 
class Granite::Connection
  def initialize(@url) : Adapter::Base.class
    @adapter = case @url
    when .starts_with? "postgres://"
      Adapter::Postgres
    when .starts_with? "mysql://"
      Adapter::Mysql
    when .starts_with? "sqlite://"
      Adapter::Sqlite
    else
      raise "Invalid connection url"
    end
  end
end
```

robacarp:
The point is not just to make the tests work, that's unrelated. I'd prefer to have this code present even with a change to the way tests are run.

eliasjpr:
My bad I read this code wrong. I see what it is doing after looking at the source code. So it is picking the adapter based on the connection. and that completely makes sense.

Blacksmoke16:
@drujensen If inheritance is working as you say, has anyone thought about MTI/STI support?

drujensen:
@Blacksmoke16 There is an issue with inheritance right now but yes, we should at some point add support.

drujensen:
@robacarp If you have two mysql databases, we will need a way to determine which one to choose per model.

drujensen:
@Blacksmoke16 Thanks for providing this.  

I am going to close this PR.  I agree that there is too much magic in the assignment of the `database_url`.  That is not a behavior that would be expected and will cause too much confusion.

@Blacksmoke16 @robacarp @eliasjpr I think we should look at adapter registration to handle this.  Take a look at #254 for my proposed solution.

PR #247, opened 2018-06-27
fix conflicting use of query

### LINKED ISSUE #246: unable to run an arbitrary query
Previously this code worked fine:
```
sections = [] of String
    Post.query("select distinct section from posts") do |rs|
      rs.each do
        sections.push(rs.read(String))
      end
    end
```
With version 0.12.1 this no longer compiles and produces this error output
```
instantiating 'PostController#index()'
in src/modules/post/post_controller.cr:20: macro didn't expand to a valid program, it expanded to:

================================================================================
--------------------------------------------------------------------------------
   1.     @@select.custom = "select distinct section from posts"
   2. 
   3.     def self.select
   4.       @@select.custom
   5.     end
   6.   
--------------------------------------------------------------------------------
Syntax error in expanded macro: query:3: can't define def inside def

    def self.select
    ^

================================================================================

    Post.query("select distinct section from posts") do |rs|
         ^~~~~
```

drujensen:
renamed the macro to `select_statement` so it won't conflict with the pass-through `query` method.  fixed #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ post = Post.all("ORDER BY posts.name DESC LIMIT 1").first
 
 #### Customizing SELECT
 
-The `query` macro allows you to customize the entire query, including the SELECT portion.  This shouldn't be necessary in most cases, but allows you to craft more complex (i.e. cross-table) queries if needed:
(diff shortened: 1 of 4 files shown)


drujensen:
@damianham 👍 right. I will add one for each pass-through.

drujensen:
@damianham @robacarp added tests for the passthroughs.

PR #249, opened 2018-06-27
Test with multiple databases

Thellior:
<h1>What is done</h1>

- Added postgres_image to the .env file with default value 10.4
- Added mysql_image to the .env file with default value 5.7
- Added dynamic value assignment on docker-compose
- Added env values to travis so it will run several builds

screen shot 2018-06-27 at 21 18 06

<h1>Questions</h1>
Currently we are testing with 

```crystal
module GraniteExample
  ADAPTERS = ["pg", "mysql", "sqlite"]
end
```

This means that in every database image it will run pg mysql sqlite. The idea of this is that we test each specific database and have failing tests based on the database. Currently if we are testing postgres:9.7 for example it will grab the default mysql image to test. So we are than testing pg mysql sqlite. I would like to refactor the tests to only test one kind of database if you guys agree. <b>What is your opinion about this?</b>


.travis.yml
```diff
@@ -1,10 +1,17 @@
...
+  - POSTGRES_IMAGE=postgres:9.6
+  - POSTGRES_IMAGE=postgres:10.4
 services:
   - docker
 
 before_install:
   - docker-compose build spec
 
 script:
+  - source .env
```
robacarp: does this need to be changed so that the travis database username matches?

.travis.yml
```diff
@@ -1,10 +1,17 @@
 language: generic
 
+env:
+  - MYSQL_IMAGE=mysql:5.6
+  - MYSQL_IMAGE=mysql:5.7
+  - POSTGRES_IMAGE=postgres:9.5
+  - POSTGRES_IMAGE=postgres:9.6
+  - POSTGRES_IMAGE=postgres:10.4
```
robacarp: ... won't these overwrite each other?

.env
```diff
@@ -1,3 +1,5 @@
 export PG_DATABASE_URL=postgres://granite:password@localhost:5432/granite_db
 export MYSQL_DATABASE_URL=mysql://granite:password@localhost:3306/granite_db
```
robacarp: do these lines need to be changed so that they're using the correct variables for travis's containers?

.env
```diff
@@ -1,3 +1,5 @@
...
 export SQLITE_DATABASE_URL=sqlite3:./config/granite.db
+export POSTGRES_IMAGE="${POSTGRES_IMAGE:=postgres:10.4}"
+export MYSQL_IMAGE="${MYSQL_IMAGE:=mysql:5.7}"
```
robacarp: It seems like these variables don't need to be in here since they're in travis.yml

.travis.yml
Thellior: I thought that in the first case as well but travis sees that there are multiple of the same variable so it will run a container for every variable. See attached image. <a href="https://docs.travis-ci.com/user/speeding-up-the-build/">Speeding up the build</a>

.env
Thellior: No they don't have to change. They will be the same for every database container that is build

.env
Thellior: Travis.yml will only load on the travis service. If you want to run docker locally it will raise an error that these variables are not in your env. If you source .env the file now it will include the default values into your env so you can use docker locally as well

.env
robacarp: I cant tell if we miscommunicated or not, so forgive me if not. 

What I mean is this. I believe that the password for Travis's database container is 'password', but surely the username for Travis's database isn't 'granite'?

.travis.yml
robacarp: Wow, what a strange syntax to produce that behavior. 

image

.env
robacarp: Got it, I see your goal. Thanks for the clarification.

I don't _know_ but I suspect this won't work with the crystal .env loaders (eg #236). That's not necessarily an objection, just something that should be made clear. My preference is to not need to `source .env` files on my dev machine for anything.

Blacksmoke16: Do we know how many people develop using the non docker version?  I'd be curious to know.

Thellior: Actually i use it because i have specific databases installed

Thellior: Actually i use it because i have specific databases installed

Thellior:
Shall i rewrite all tests so we only test one database at a time??

drujensen:
@Thellior Take a look at #257. We will be able to support 2 mysql db's at the same time i.e. 5.6 adapter and a 5.7 adapter

Blacksmoke16:
Adapter registration is merged now.  What's the status of this?

Thellior:
Currently I’m on holiday so I cannot change anything but this is currently blocked because I’m not sure which direction of multiple database testing we like to go

Blacksmoke16:
cc @robacarp @drujensen

robacarp:
As much as I like not needing to run tests in docker to get the whole suite exercised, I think the current method of macro-iterating the available adapters is going to prevent the project from adequately testing against many versions of many databases efficiently.

I'd be open to removing this pattern:
```
{% for adapter in GraniteExample::ADAPTERS %}
  module {{adapter.capitalize.id}}
    describe "{{ adapter.id }} test" do
```

if we can come up with a better way to indicate that a specific test _isn't_ going to run against a specific adapter. Up until recently sqlite had a bunch of extra tests around dates and times because it handled that very differently. I don't doubt for one second that we have more of that kind of special case testing and that amount of code is more likely to grow than it is to shrink.

Thellior:
https://github.com/amberframework/granite/pull/270

PR #253, opened 2018-07-15
Support JSON::Serializable

Blacksmoke16:
This PR, fixes #253 brings about native support for JSON::Serializable.  This is super exciting and probably my favorite PR so far.  The benefits of this are immense, especially when working with JSON APIs.

```Crystal
class Foo < Granite::Base
  adapter mysql
  table_name foos

  field name : String
  field age : Int32
end
```
#### Single Object
`foo = Foo.from_json(%({"name": "Granite", "age": 10})) `
```Crystal 
<Foo:0x55c77d901ea0
 @_current_callback=nil,
 @age=10,
 @created_at=nil,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @name="Granite",
 @new_record=true,
 @updated_at=nil>
```
`foo.to_json`
```JSON
{
  "name": "Granite",
  "age": 10
}
```
#### Array of Objects

`foo = Array(Foo).from_json(%([{"name": "Granite1", "age": 10},{"name": "Granite2", "age": 50}]))`
```Crystal
[#<Foo:0x55f66bdeaea0
  @_current_callback=nil,
  @age=10,
  @created_at=nil,
  @destroyed=false,
  @errors=[],
  @id=nil,
  @name="Granite1",
  @new_record=true,
  @updated_at=nil>,
 #<Foo:0x55f66bdeae10
  @_current_callback=nil,
  @age=50,
  @created_at=nil,
  @destroyed=false,
  @errors=[],
  @id=nil,
  @name="Granite2",
  @new_record=true,
  @updated_at=nil>]
```
`foo.to_json`
```JSON
[
  {
    "name": "Granite1",
    "age": 10
  },
  {
    "name": "Granite2",
    "age": 50
  }
]
```
It works pretty much out of the box, but with one slight breaking change.  By default, as you may have noticed above, when doing `to_json` it will *not* add nil values into the object, i.e. the `id`.  If the user wants that, they would have to do:
```Crystal
@[JSON::Serializable::Options(emit_nulls: true)]
class Foo < Granite::Base
  adapter mysql
  table_name foos

  field name : String
  field age : Int32
end
```
`foo = Foo.from_json(%({"name": "Granite", "age": 10})) `
```Crystal 
<Foo:0x55c77d901ea0
 @_current_callback=nil,
 @age=10,
 @created_at=nil,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @name="Granite",
 @new_record=true,
 @updated_at=nil>
```
`foo.to_json`
```JSON
{
  "name": "Granite",
  "age": 10,
  "id": null,
  "created_at": null,
  "updated_at": null
}
```

#### after_initialize
Another cool feature this brings is the `after_initialize` method. This method gets called after `from_json` is done parsing the given json.  This allows you to set other fields that are not in the JSON directly or that require some more logic. 
```Crystal
class Foo < Granite::Base
  adapter mysql
  table_name foos

  field name : String
  field age : Int32

  def after_initialize
      @age = 0
  end
end
```
`foo = Foo.from_json(%({"name": "Granite1"}))`
```Crystal 
<Foo:0x55c77d901ea0
 @_current_callback=nil,
 @age=0,
 @created_at=nil,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @name="Granite",
 @new_record=true,
 @updated_at=nil>
```

#### JSON::Serializable::Unmapped

```
 If the JSON::Serializable::Unmapped module is included, unknown properties in the JSON document will be stored in a Hash(String, JSON::Any). On serialization, any keys inside json_unmapped will be serialized appended to the current json object.
```

```Crystal
class Foo < Granite::Base
  include JSON::Serializable::Unmapped

  adapter mysql
  table_name foos

  field name : String
  field age : Int32
end
```
`foo = Foo.from_json(%({"name": "Granite1", "age": 12, "foobar": true}))`
```Crystal 
#<Foo:0x55efba40ff00
 @_current_callback=nil,
 @age=12,
 @created_at=nil,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @json_unmapped={"foobar" => true},
 @name="Granite1",
 @new_record=true,
 @updated_at=nil>
```
`foo.to_json`
```JSON
{
  "name": "Granite",
  "age": 12,
  "foobar": true
}
```

#### on_to_json

Allows the user to tap into the `to_json` and do additional manipulations before returning it.

```Crystal
class Foo < Granite::Base
  adapter mysql
  table_name foos

  field name : String
  field age : Int32

  def on_to_json(json : JSON::Builder)
      json.field("years_young", @age)
  end
end
```
`foo = Foo.from_json(%({"name": "Granite1", "age": 12}))`
```Crystal 
#<Foo:0x55eb12bd2f00
 @_current_callback=nil,
 @age=12,
 @created_at=nil,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @name="Granite1",
 @new_record=true,
 @updated_at=nil>
```
`foo.to_json`
```JSON
{
  "name": "Granite",
  "age": 12,
  "years_young": 12
}
```

### DIFF
--- a/src/granite/error.cr
+++ b/src/granite/error.cr
@@ -5,6 +5,10 @@ class Granite::Error
   end
 
   def to_s
-    (@field == :base) ? @message : "#{@field.to_s.capitalize} #{message}"
+    if @field == :base
+      @message
+    else
+      "#{@field.to_s.capitalize} #{message}"
+    end
   end
 end
--- a/src/granite/error.cr
+++ b/src/granite/error.cr
@@ -5,10 +5,6 @@ class Granite::Error
   end
 
   def to_s
-    if @field == :base
-      @message
-    else
-      "#{@field.to_s.capitalize} #{message}"
-    end
+    (@field == :base) ? @message : "#{@field.to_s.capitalize} #{message}"
   end
 end
--- a/src/granite/fields.cr
+++ b/src/granite/fields.cr
@@ -39,6 +39,9 @@ module Granite::Fields
     {% for name, options in FIELDS %}
       {% type = options[:type] %}
       {% suffixes = options[:raise_on_nil] ? ["?", ""] : ["", "!"] %}
+      {% if options[:json_options] %}
+         @[JSON::Field({{**options[:json_options]}})]
+      {% end %}
       property{{suffixes[0].id}} {{name.id}} : Union({{type.id}} | Nil)
       def {{name.id}}{{suffixes[1].id}}
         raise {{@type.name.stringify}} + "#" + {{name.stringify}} + " cannot be nil" if @{{name.id}}.nil?
--- a/src/granite/fields.cr
+++ b/src/granite/fields.cr
@@ -1,7 +1,7 @@
 require "json"
 
 module Granite::Fields
-  alias Type = DB::Any | JSON::Any
+  alias Type = DB::Any
   TIME_FORMAT_REGEX = /\d{4,}-\d{2,}-\d{2,}\s\d{2,}:\d{2,}:\d{2,}/
 
   macro included
@@ -81,22 +81,6 @@ module Granite::Fields
       return fields
     end
 
-    def to_json(json : JSON::Builder)
-      json.object do
-        {% for name, options in FIELDS %}
-          {% type = options[:type] %}
-          %field, %value = "{{name.id}}", {{name.id}}
-          {% if type.id == Time.id %}
-            json.field %field, %value.try(&.to_s(Granite::DATETIME_FORMAT))
-          {% elsif type.id == Slice.id %}
-            json.field %field, %value.id.try(&.to_s(""))
-          {% else %}
-            json.field %field, %value
-          {% end %}
-        {% end %}
-      end
-    end
-
     def read_attribute(attribute_name : Symbol | String) : DB::Any
       {% begin %}
         case attribute_name.to_s
@@ -113,10 +97,6 @@ module Granite::Fields
       args.each do |k, v|
         cast_to_field(k, v.as(Type))
       end
-    end
-
-    def set_attributes(attributes : JSON::Any)
-      set_attributes(attributes.as_h)
     end
 
     def set_attributes(**args)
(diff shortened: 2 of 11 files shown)


Blacksmoke16:
Latest commit adds support for the `@[JSON::Field()]` property annotations 

Usage is like `field name : String, json_options: {ignore: true}`

Blacksmoke16:
👍 Yea no kidding, much better.  Don't merge yet, think i want to include some more details in readme on at least the `json_options` hash on fields macro.

drujensen:
@Blacksmoke16 Is there any side affects using the native version?  I recall attempting this before but the constructor needed required supporting the JSON::Puller or something like that.

Blacksmoke16:
This is a bit difference than the JSON.mapping version.  This was just released with Crystal 0.25.0.  

Using this it isn't possible to give like `Model.create(JSON::ANY)`, would have to do `Model.from_json(json_string)`.

All the tests passed and all the Serializable extra stuff seems to work fine so unless something comes up i think we're good.

drujensen:
aaah, I was wondering how this was possible without the constructor.  I will checkout 0.25 release notes for more info.  This is all excellent news!   I prefer the `.from_json` so that is fine.

Blacksmoke16:
https://crystal-lang.org/api/0.25.1/JSON/Serializable.html

Is a good overview of it.

src/granite/error.cr
```diff
@@ -5,10 +5,6 @@ class Granite::Error
   end
 
   def to_s
-    if @field == :base
-      @message
-    else
-      "#{@field.to_s.capitalize} #{message}"
-    end
+    (@field == :base) ? @message : "#{@field.to_s.capitalize} #{message}"
```
robacarp: Can we leave this as it was? I don't see the benefit to the more subtle syntax.

src/granite/fields.cr
```diff
@@ -39,6 +39,9 @@ module Granite::Fields
     {% for name, options in FIELDS %}
       {% type = options[:type] %}
       {% suffixes = options[:raise_on_nil] ? ["?", ""] : ["", "!"] %}
+      {% if options[:json_options] %}
+         @[JSON::Field({{**options[:json_options]}})]
+      {% end %}
```
robacarp: clever, I like it.

Blacksmoke16:
There is no dependence on `after_initialize`.  It is just an undocumented feature that can be used.  

It could be removed and the basic implementation would still be fine.  It is just a method that gets called after the `from_json` method is finished.

src/granite/error.cr
Blacksmoke16: I can change it back if you want.  I just have a thing for ternary, easier to read and cleaner imo.

Blacksmoke16:
Updated README on JSON support.  Can see it at https://github.com/Blacksmoke16/granite/tree/JSON-Serializable-support#json-support.

Otherwise this is good to merge unless one of you finds an issue.

src/granite/error.cr
robacarp: I guess this is the sort of thing that a style guide helps communicate. My preference is to avoid ternaries wherever possible. 

More than anything, this is an unrelated change to the rest of the pull. I don’t feel really strong about changing this, but I’d like to avoid someone else changing it back simply because they prefer it to be the other way. It just creates more potential for bugs to sneak in and opens the door for endless bike shedding.

drujensen:
@robacarp can we merge this?

src/granite/error.cr
Blacksmoke16: Fair enough.

Issue #223, opened 2018-06-08
Create Granite documentation


faustinoaq:
Hi @drujensen I think Granite's README is growing very fast, maybe we should maintain a separate directory (or repo) for docs like Crecto and Jennifer do, WDYT?

BTW, I already created a GitBook page for Granite here: ~~https://amberframework.gitbook.io/granite~~ (linked to this repo and using Granite's README)

Update: new link :tada: https://docs.amberframework.org/granite/

Blacksmoke16:
At the very least a table of contents would help.

PR #255, opened 2018-07-17
Support YAML::Serializable

Blacksmoke16:
Surprise!  Is branched off of #253 so needs that first.

I'll need to edit the README again, probably use same examples as the JSON, but say it could also be YAML.

EDIT:  Since we didnt have this feature before i updated the labels.


drujensen:
@Blacksmoke16 wow, very cool!

Blacksmoke16:
An update on this, am going to skip adding docs for this for now.

Plan on dealing with #223 by making a directory to split-up our README into one topic per file and will write up docs for YAML then.

robacarp:
@Blacksmoke16 looks like this will need some manual intervention to bring up to date with the json stuff.

Blacksmoke16:
rip.  I'll take a look when i get home.

spec/granite_spec.cr
```diff
@@ -67,7 +67,76 @@ require "./spec_helper"
...
-            collection.should eq(%([{"name":"todo 1","priority":1},{"name":"todo 2","priority":2},{"name":"todo 3","priority":3}]))
+            collection.should eq %([{"name":"todo 1","priority":1},{"name":"todo 2","priority":2},{"name":"todo 3","priority":3}])
+          end
+        end
+      end
+
+      describe "YAML" do
+        context ".from_yaml" do
+          it "can create an object from YAML" do
+            yaml_str = %({"name": "yaml::anyReview","upvotes": 2, "sentiment": 1.23, "interest": 4.56, "published": true})
```
robacarp: should this be more classical yaml instead of basically json?

robacarp:
Thanks @Blacksmoke16

Issue #254, opened 2018-07-16
Support adapter registration


drujensen:
We should look at supporting registration of an adapter and then using that adapter by name in the model.  Right now we are hard coding the adapter names to `pg`, `mysql` and `sqlite` but I think we can fix this by allowing you to register your own named adapter and then using that named adapter in the model.

Registering a new adapter:
```
adapter = Granite::Adapter::Pg.new(
 name: "my_pg_adapter", 
 url: "postgres://localhost/database"
)
Granite.adapters << adapter
```
And then using it:
```
class MyModel < Granite::Base
  adapter my_pg_adapter
end
```

This will allow for other adapters in the future and support connecting to two of the same type of database as well.

Blacksmoke16:
This is pretty much what @robacarp and @eliasjpr were talking about in my old PR.

See https://github.com/amberframework/granite/pull/242#issuecomment-400520029 and https://github.com/amberframework/granite/pull/242#issuecomment-400711014

drujensen:
Correct, except this will support future adapters and handles multiple connections to the same type of database using the adapters registered name.

PR #257, opened 2018-07-19
First pass at adapter registration support

Blacksmoke16:
First pass on #254.  Tests pass so assuming this works.  Still some things to do tho :p Feedback appreciated :)

TODOs:
* Write tests for this
* Possibly cleanup `Granite.adapters.adapters` syntax
* Update docs

### DIFF
--- a/spec/adapter/base_spec.cr
+++ b/spec/adapter/base_spec.cr
@@ -1,32 +1,32 @@
-require "../spec_helper"
+# require "../spec_helper"
 
-describe Granite::Adapter::Base do
-  describe "#env" do
-    context "no ${} exists" do
-      it "should return itself" do
-        url = "dummy://user:pswd@host:3333/database"
-        Granite::Adapter::Base.env(url).should eq url
-      end
-    end
+# describe Granite::Adapter::Base do
+#   describe "#env" do
+#     context "no ${} exists" do
+#       it "should return itself" do
+#         url = "dummy://user:pswd@host:3333/database"
+#         Granite::Adapter::Base.env(url).should eq url
+#       end
+#     end
 
-    context "one ${} exists" do
-      it "should replace it with environment variable" do
-        ENV["DATABASE"] = "test"
-        url = "dummy://user:pswd@host:3333/${DATABASE}"
-        Granite::Adapter::Base.env(url).should eq "dummy://user:pswd@host:3333/test"
-      end
-    end
+#     context "one ${} exists" do
+#       it "should replace it with environment variable" do
+#         ENV["DATABASE"] = "test"
+#         url = "dummy://user:pswd@host:3333/${DATABASE}"
+#         Granite::Adapter::Base.env(url).should eq "dummy://user:pswd@host:3333/test"
+#       end
+#     end
 
-    context "multiple ${} exists" do
-      it "should replace each with correct environment variable" do
-        ENV["USER"] = "user"
-        ENV["PSWD"] = "pswd"
-        ENV["HOST"] = "host"
-        ENV["PORT"] = "3333"
-        ENV["DATABASE"] = "test"
-        url = "dummy://${USER}:${PSWD}@${HOST}:${PORT}/${DATABASE}"
-        Granite::Adapter::Base.env(url).should eq "dummy://user:pswd@host:3333/test"
-      end
-    end
-  end
-end
+#     context "multiple ${} exists" do
+#       it "should replace each with correct environment variable" do
+#         ENV["USER"] = "user"
+#         ENV["PSWD"] = "pswd"
+#         ENV["HOST"] = "host"
+#         ENV["PORT"] = "3333"
+#         ENV["DATABASE"] = "test"
+#         url = "dummy://${USER}:${PSWD}@${HOST}:${PORT}/${DATABASE}"
+#         Granite::Adapter::Base.env(url).should eq "dummy://user:pswd@host:3333/test"
+#       end
+#     end
+#   end
+# end
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -5,30 +5,16 @@ require "db"
 # objects to perform actions against a specific database.  Each adapter needs
 # to implement these methods.
 abstract class Granite::Adapter::Base
-  property database : DB::Database
+  property name : String
+  property url : String
 
-  def initialize(adapter : String)
-    if url = ENV["DATABASE_URL"]? || Granite.settings.database_url || replace_env_vars(settings(adapter)["database"].to_s)
-      @database = DB.open(url)
-    else
-      raise "database url needs to be set in the config/database.yml or DATABASE_URL environment variable"
-    end
-  end
-
-  DATABASE_YML = "config/database.yml"
-
-  def settings(adapter : String)
-    if File.exists?(DATABASE_YML) &&
-       (yaml = YAML.parse(File.read DATABASE_YML)) &&
-       (settings = yaml[adapter])
-      settings
-    else
-      return {"database": ""}
-    end
+  def initialize(connection_hash : NamedTuple(url: String, name: String))
+    @name = connection_hash[:name]
+    @url = connection_hash[:url]
   end
 
   def open(&block)
-    yield @database
+    yield DB.open(@url)
   end
 
   def log(query : String, params = [] of String) : Nil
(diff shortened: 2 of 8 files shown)


Blacksmoke16:
@drujensen @robacarp How do you think I should best test this?  In theory I shouldn't have to actually test connecting to diff databases as that already happens for each adapter macro.  

Probably just add some connections and ensure that it finds right one?

drujensen:
Maybe we can add multiple database testing #249 with this effort and leverage different adapters to the same db type.  wdyt?

src/granite/settings.cr
```diff
@@ -2,16 +2,15 @@ require "logger"
 
 module Granite
   class Settings
-    property database_url : String? = nil
-    property logger : Logger
+    class_getter adapters = [] of Granite::Adapter::Base
+    class_property logger = Logger.new STDOUT
```
drujensen: I wonder if we should change this to `::Logger.new(nil)`.  This potentially could be a bottleneck if all logging is sent to a single IO class instance.  @robacarp wdyt?

src/granite/table.cr
```diff
@@ -9,11 +9,7 @@ module Granite::Table
   # specify the database adapter you will be using for this model.
   # mysql, pg, sqlite, etc.
   macro adapter(name)
-    @@adapter = Granite::Adapter::{{name.id.capitalize}}.new("{{name.id}}")
-
-    def self.adapter
-      @@adapter
-    end
+    class_getter adapter : Granite::Adapter::Base = Granite::Settings.adapters.find { |a| a.name == {{name.stringify}} } || raise "No registered adapter with the name '{{name.id}}'"
```
drujensen: nice!

src/granite/settings.cr
Blacksmoke16: Does having it at the class level matter?  Before, `@logger` was defaulted to `Logger.new STDOUT` as well.  Was just keeping that behavior.

Blacksmoke16:
@drujensen That might be the best option IMO.  The more I think about this, i'm thinking that it gets tested by default.  Otherwise none of the 3 adapters we provide would connect and all the tests would fail.

Testing on a different versions would be a good addition.  Deff work pursuing.

Blacksmoke16:
Alright @drujensen @robacarp @eliasjpr This is ready for review now.

spec/adapter/base_spec.cr
```diff
@@ -1,32 +1,40 @@
...
+        mysql_adatper.url.should eq ENV["MYSQL_DATABASE_URL"]
+      end
+      if pg_adatper = Granite::Settings.adapters.find { |adapter| adapter.name == "pg" }
+        pg_adatper.url.should eq ENV["PG_DATABASE_URL"]
+      end
+      if sqlite_adapter = Granite::Settings.adapters.find { |adapter| adapter.name == "sqlite" }
+        sqlite_adapter.url.should eq ENV["SQLITE_DATABASE_URL"]
+      end
+      if foo_adapter = Granite::Settings.adapters.find { |adapter| adapter.name == "FooBar123" }
+        foo_adapter.url.should eq "FAKE_DB"
```
robacarp: these tests can fail with a false positive if the find returns nil. Is there an easy way to write them so that they'll throw when find returns nil?

spec/adapter/base_spec.cr
```diff
@@ -1,32 +1,40 @@
...
       end
     end
 
-    context "one ${} exists" do
-      it "should replace it with environment variable" do
-        ENV["DATABASE"] = "test"
-        url = "dummy://user:pswd@host:3333/${DATABASE}"
-        Granite::Adapter::Base.env(url).should eq "dummy://user:pswd@host:3333/test"
+    it "should disallow adapters multiple adapters with the same name" do
```
robacarp: should this be "should disallow multiple adapters with the same name"?

spec/adapter/base_spec.cr
```diff
@@ -1,32 +1,40 @@
...
+end
+
 describe Granite::Adapter::Base do
-  describe "#env" do
-    context "no ${} exists" do
-      it "should return itself" do
-        url = "dummy://user:pswd@host:3333/database"
-        Granite::Adapter::Base.env(url).should eq url
+  describe "registration" do
+    it "should allow adapters to be registered" do
```
robacarp: it "should allow adapters to be saved and looked up"

spec/adapter/base_spec.cr
```diff
@@ -1,32 +1,40 @@
...
+  adapter FooBar123
+end
+
 describe Granite::Adapter::Base do
-  describe "#env" do
-    context "no ${} exists" do
-      it "should return itself" do
-        url = "dummy://user:pswd@host:3333/database"
-        Granite::Adapter::Base.env(url).should eq url
+  describe "registration" do
```
robacarp: Since the tests in this file mess with the global scope of tests, it has some potential to create problems with other tests when run in different orders. Is it possible to use some test interface to run these tests without depending on the global scope of Granite::Settings?

src/adapter/base.cr
```diff
@@ -5,34 +5,21 @@ require "db"
...
-  def settings(adapter : String)
-    if File.exists?(DATABASE_YML) &&
-       (yaml = YAML.parse(File.read DATABASE_YML)) &&
-       (settings = yaml[adapter])
-      settings
-    else
-      return {"database": ""}
-    end
+  def initialize(connection_hash : NamedTuple(url: String, name: String))
+    raise "Adapter with name '#{connection_hash[:name]}' has already been registered." if Granite::Settings.adapters.any? { |adapter| adapter.name == connection_hash[:name] }
```
robacarp: This check logic feels out of place here, because it feels like this specific instance of an adapter shouldn't have knowledge of the rest of the adapters. Is there another place it could live where it's not crossing domain boundaries?

src/granite/settings.cr
```diff
@@ -2,16 +2,11 @@ require "logger"
 
 module Granite
   class Settings
-    property database_url : String? = nil
-    property logger : Logger
+    class_getter adapters = [] of Granite::Adapter::Base
```
robacarp: Why should these be changed from properties of a class to class properties? This feels like it will set Granite back a few steps in composability.

src/granite/settings.cr
robacarp: In practice this will almost always be initialized to STDOUT, or a logger pointed at STDOUT. But I think the default should be nil. It's frustrating when including a library suddenly generates a bunch of output in a program and requires additional configuration to silence it. For example, imagine including Granite in a console application which reads sqlite files -- Granite would pollute the user interface by default. The logger facility is a great tool for use of granite, but silence feels like a nice default.

spec/adapter/base_spec.cr
Blacksmoke16: Could just do the find off adapters array and `not_nil!` it?

spec/adapter/base_spec.cr
Blacksmoke16: Possibly?  Really i could probably refactor that to use a one of the predefined classes and just test it based off of adapters already registered.  

i.e.

```Crystal
class Foo
  adapter mysql
end

adapter = Foo.adapter
adapter.name.should eq "mysq"
adapter.url.should eq ENV['MYSQL_DATABASE_URL"]
```

Would that be better?

src/adapter/base.cr
Blacksmoke16: Not sure where else to put it.  Can't override the `<<` method in the Settings class so this was only other place that kinda worked since it has access to name of adapter being instantiated.

src/granite/settings.cr
Blacksmoke16: I'm not sure which is the preferred way of doing this.  I assumed ```@@settings ||= Settings.new` was kind of a hack and this was the better method?

If not i can change it back if you'd like.

Blacksmoke16:
Refactored a bit:

* Syntax is now `Granite::Adapters << adapter`
  * The `Adapters' <<` class method now hold the duplicate name logic
* Changed the settings back to how it was before
  * Defaulted logger to nil

Blacksmoke16:
For how easy it would be to do might not be a bad idea.

drujensen:
@Blacksmoke16 this is excellent work.  Yes, I am for making the models as simple as possible. I think it would need to raise an exception if there are 2 or more adapters and no adapter is specified, right?

Blacksmoke16:
More i'm looking into this, the main issue is we are using the `adapter` macro to set the adapter for a given model.  But if that is omitted what would set the adapter?

robacarp:
@blacksmoke16 hmm good question. Maybe we could implement the adapter method on the base class to find the default adapter and raise if there’s an ambiguity, and then have the adapter macro define a new method which explicitly retrieves the specified adapter.

drujensen:
Let’s do this in a separate PR.

PR #259, opened 2018-07-20
Fix wording on dep install

drinkmorewaters:
from crystal dep to shards install


drinkmorewaters:
You're welcome :) Can't really code, so I'm helping out in other ways.

PR #261, opened 2018-07-21
Optionally add nodoc

### LINKED ISSUE #251: Add `# :nodoc:` to some properties/methods
Quite a few Granite methods/constants pollute the generated docs.

For example i dont think `adapter` `table_name` `validate` `CALLBACKS` `CONTENT_FIELDS` etc should be included in the docs for classes using Granite.  All they do is pollute and quite frankly take over the generated docs for any class inheriting from `Granite::Base`

To clarify this would be docs generated via `crystal docs`

WDYT?

EDIT:  TIL we have an update method.

Blacksmoke16:
This PR resolves #251.

It provides the ability to skip generation of Granite methods, macros, constants etc when using `crystal docs`.  It will only _not_ include the docs when the ENV var `DISABLE_GRANITE_DOCS=true`, if it is not set, or set to false it will include the Granite stuff.

The reasoning behind this; before this change on a just primary key model:

image

There is so much here that it almost makes the docs for the user's project unusable.

The same model skipping Granite stuff:

image

As you may notice this also adds the ability to define a comment to use on the getter/setter of each property with the following syntax: 

`field name : String, comment: "# The name of person"`

or in this case:

`primary id : Int64, comment: "# The primary key"`

### DIFF
--- a/src/granite/table.cr
+++ b/src/granite/table.cr
@@ -1,6 +1,6 @@
 # Adds a :nodoc: to granite methods/constants if `DISABLE_GRANTE_DOCS` ENV var is true
-macro noDoc(stmt)
-  {% if env("DISABLE_GRANTE_DOCS") == "true" %}
+macro disable_granite_docs?(stmt)
+  {% unless env("DISABLE_GRANITE_DOCS") == "false" %}
     # :nodoc:
     {{stmt.id}}
   {% else %}
@@ -11,8 +11,8 @@ end
 module Granite::Table
   macro included
     macro inherited
-      noDoc SETTINGS = {} of Nil => Nil
-      noDoc PRIMARY = {name: id, type: Int64, auto: true}
+      disable_granite_docs? SETTINGS = {} of Nil => Nil
+      disable_granite_docs? PRIMARY = {name: id, type: Int64, auto: true}
     end
   end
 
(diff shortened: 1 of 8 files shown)


src/granite/table.cr
```diff
@@ -1,8 +1,18 @@
+# Adds a :nodoc: to granite methods/constants if `DISABLE_GRANTE_DOCS` ENV var is true
+macro noDoc(stmt)
+  {% if env("DISABLE_GRANTE_DOCS") == "true" %}
+    # :nodoc:
+    {{stmt.id}}
+  {% else %}
+    {{stmt.id}}
+  {% end %}
+end
```
robacarp: Clever. I like that the logic is centralized.

`noDoc` feels like a command though, which would prevent anything from being documented. Also it's snakeCased, which doesn't feel very Crystal.

Naming isn't my forte, but is there another name that would work better here?

src/granite/table.cr
```diff
@@ -1,8 +1,18 @@
+# Adds a :nodoc: to granite methods/constants if `DISABLE_GRANTE_DOCS` ENV var is true
+macro noDoc(stmt)
+  {% if env("DISABLE_GRANTE_DOCS") == "true" %}
```
robacarp: The purpose here is to make the docs more terse and relevant by default, right? Should this logic default to nodoc then?

src/granite/table.cr
Blacksmoke16: Yea...i'll think of something better

src/granite/table.cr
Blacksmoke16: Depends on how you look at it.  Was trying to allow you to disable the docs if you wanted while retaining current behavior.

Blacksmoke16:
@robacarp how about `disable_granite_docs?`, now defaults to off unless the env variable is false

PR #262, opened 2018-07-23
Revert `.is_a?(JSON::Any)`

Blacksmoke16:
Missed this logic when adding JSON::Serializable support.

### DIFF
--- a/src/granite/fields.cr
+++ b/src/granite/fields.cr
@@ -122,7 +122,7 @@ module Granite::Fields
           when "{{_name.id}}"
             if "{{_name.id}}" == "{{PRIMARY[:name]}}"
               {% if !PRIMARY[:auto] %}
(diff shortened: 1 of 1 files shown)


src/granite/fields.cr
```diff
@@ -122,7 +122,7 @@ module Granite::Fields
           when "{{_name.id}}"
             if "{{_name.id}}" == "{{PRIMARY[:name]}}"
               {% if !PRIMARY[:auto] %}
-                @{{PRIMARY[:name]}} = value.is_a?(JSON::Any) ? value.raw.as({{PRIMARY[:type]}}) : value.as({{PRIMARY[:type]}})
```
drujensen: Can you provide a better explanation?  Why do the other types not need it?

Blacksmoke16: This was added in #221 to parse the primary key value from `JSON::Any`.

Since we are using `JSON::Serializable` now, we don't define our own `.from_json` thus `JSON::Any` never reaches this anymore.  

This should have been changed back in #253 with the removal of the rest of the `JSON::Any` typings, but i missed this one.

PR #264, opened 2018-08-16
Has one relationship

zaidakram:
My first ever contribution to this wonderful project. And thanks to all the contributors for the awesome stuff that is Amber.
 
This PR adds `has_one` relationship to the models. Adds two simple methods which are documented and tested.

**NOTE:** It does not support through associations yet.

### DIFF
--- a/docs/relationships.md
+++ b/docs/relationships.md
@@ -139,7 +139,6 @@ CREATE TABLE posts (
 CREATE INDEX 'user_id_idx' ON posts (user_id);
 ```
 
-
 ## Many to Many
 
 Instead of using a hidden many-to-many table, Granite recommends always creating a model for your join tables.  For example, let's say you have many `users` that belong to many `rooms`. We recommend adding a new model called `participants` to represent the many-to-many relationship.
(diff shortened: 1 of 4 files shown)


docs/relationships.md
```diff
@@ -60,6 +139,7 @@ CREATE TABLE posts (
 CREATE INDEX 'user_id_idx' ON posts (user_id);
 ```
 
+
```
Blacksmoke16: Can we remove this extra line?

zaidakram: Done!

src/granite/associations.cr
```diff
@@ -36,6 +36,24 @@ module Granite::Associations
...
+    # retrieve the associated object
+    def {{model_name.id}}
+      {{model_name.id.camelcase}}.find_by({{foreign_key.id}}: self.id)
+    end
+
+    # set the associated object to given relationship
+    def {{model_name.id}}=(children)
+      children.{{foreign_key.id}} = self.id
+    end
+  end
```
robacarp: The comments here are probably redundant. The code is pretty straightforward.

zaidakram: Just removed them. You are right. Code is pretty self-explanatory.

PR #270, opened 2018-08-21
Test multiple databases

Thellior:
<h1>What is done</h1>

- Refactored the test suite to only test one database at a time.
- Added CURRENT_ADAPTER to the env file
- Travis is testing database types separately now
- Travis is now testing seperate database version parallel.

<h1>Notes</h1>
On Travis website it show 10+ minutes run time. This is because it will run one container for each matrix item. The matrix items run in parallel. The 10+ minutes is all container time together. The whole suits runs in 3+ minute

 Ran for 3 min 36 sec
 Total time 10 min 7 sec

<h1>References</h1>

- https://github.com/amberframework/granite/pull/249
- https://docs.travis-ci.com/user/speeding-up-the-build/


robacarp:
This is nice, and it should make testing new versions of databases much easier.

image

Thellior:
@robacarp I also moved some models back into the models file. I need to fix that for the validation helpers as well there is a issue open for that matter. I already discussed the issue with you. When this is merged i will continue to work on that :)

Blacksmoke16:
👍 I'll take a look this weekend.

robacarp:
@Thellior is there a way to make the docker-compose have a default, which runs the tests against something when the `-f` parameter isn't used?

Blacksmoke16:
Probably by adding a `CURRENT_ADAPATER` env variable to the parent `docker-compose.yml`

Thellior:
@robacarp @Blacksmoke16 I only capable of setting the current_adapter to sqlite. If i'm going to change it to pg or mysql we need to specify the depends_on option and the pg or mysql image. If i do so we receive a error that we cannot extend the other files because you cannot extend depends_on.

```shell
ERROR: Cannot extend service 'spec' in /home/thellior/Git/amber/granite/docker-compose.yml: services with 'depends_on' cannot be extended
```

If you guys want a different option like pg i have to investigate how to achieve that with the multiple files.
Let me know what you guys think

robacarp:
@Thellior I'm fine with setting the default to sqlite, especially now that the sqlite driver has essentially the same code and feature set as the rest of the drivers do.

spec/spec_helper.cr
```diff
@@ -7,10 +6,8 @@ Granite::Adapters << Granite::Adapter::Mysql.new({name: "mysql", url: ENV["MYSQL
 Granite::Adapters << Granite::Adapter::Pg.new({name: "pg", url: ENV["PG_DATABASE_URL"]})
```
drujensen: This isn't going to work since we establish a connection when this get's instantiated.  We should lazy load the connection.  @Thellior Can you fix that?

Issue #271, opened 2018-08-25
New database connections are opened on each query


yumoose:
After adapter registration support (#257) was added, each time Granite tries to open a connection to the database, a new database is opened, rather than the reusing the already opened database in the adapter.

```crystal
# src/adapter/base.cr

def open(&block)
  yield DB.open(@url)
end
```

This breaks being able to set connection pool configuration as described in the [Crystal docs](https://crystal-lang.org/docs/database/connection_pool.html), and is leading to a massive number of `too many connections for role/user` errors thrown by the database. This is severe for limited tier databases (eg. Heroku,) where the number of open connections allowed remains quite low, where errors arise after just a few queries.

Perhaps the adapter should keep a single reference to the database to make successive queries to, I'm just unsure as to when the database should be closed as adapters are not yet unregistered by the looks of it, and I haven't seen any tear-down logic for application shutdown, not that I have dug around (though this seems like another, less-pressing issue anyway).

drujensen:
@yumoose You are correct.  I should have caught this since I had changed this to maintain a single database before.  I will look into fixing this issue.  Thanks for reporting it.

rishavs:
Probably related to this. I am also getting too many open connections error on running a simple select query a couple of times. Also the queries are taking 1+sec while they generally take 150 ms to complete.

`````
GET / - Unhandled exception:
 (DB::ConnectionRefused)
  from lib/pg/src/pg/connection.cr:16:9 in 'initialize'
  from lib/pg/src/pg/connection.cr:7:5 in 'new'
  from lib/pg/src/pg/driver.cr:3:5 in 'build_connection'
  from lib/db/src/db/database.cr:51:9 in '->'
  from lib/db/src/db/pool.cr:255:3 in 'build_resource'
  from lib/db/src/db/pool.cr:17:34 in 'initialize'
  from lib/db/src/db/pool.cr:15:5 in 'new:initial_pool_size:max_pool_size:max_idle_pool_size:checkout_timeout:retry_attempts:retry_delay'
  from lib/db/src/db/database.cr:50:7 in 'initialize'
  from lib/db/src/db/database.cr:43:5 in 'new'
  from lib/db/src/db.cr:139:5 in 'build_database'
  from lib/db/src/db.cr:135:5 in 'build_database'
  from lib/db/src/db.cr:103:5 in 'open'
  from lib/granite/src/granite/querying.cr:17:11 in 'raw_all'
[...]

PR #272, opened 2018-08-25
Reuse open database connections in adapters

yumoose:
Totally open to feedback on this approach @drujensen, but it looks to share the open database across multiple queries fine.

Looks to resolve issue #271.


drujensen:
checkout #274

Issue #243, opened 2018-06-23
Implement ==(other) and <=>(other) on Base


robacarp:
It'd be handy both for Granite specs and for application devs to be able to take an Array(MyModel) and call `#sort` on it, or be able to ask if two instances of a model are `==` each other.

drujensen:
@robacarp I think this requires some knowledge of the business logic to determine if an object is less than or greater than another object.  I don't think we should try to implement this feature.  For example, you might use the `name` field for a `User` object but you would use the `zip code` for an `Address` object.

PR #276, opened 2018-08-26
Issue #243, implement == and <=> on base.

astupidnerd:
This seemed like a good first issue to tackle ( #243 ).

This implements == and <=> by comparing the fields defined in `FIELDS`.

One limitation my implementation has is that it will not compare fields whose type does not implement `<=>` itself. I think this decision makes sense (how else would you compare those fields?), but it probably should be mentioned somewhere in the docs.

That being said, if a field's type does not have a `<=>` method, it will just _safely_ skip checking that field.

I have also added in a new set of unit tests for this feature.

I'm still fairly new to Crystal (and Amber/Granite), so please let me know if I did anything unusual and/or if there is anything that could be done in an better way.

### DIFF
--- a/spec/granite/comparisons/comparisons_spec.cr
+++ b/spec/granite/comparisons/comparisons_spec.cr
@@ -68,48 +68,35 @@ module {{adapter.capitalize.id}}
         review_ids.should eq(sorted)
       end
 
-      it "sorts records by other fields if the ids are the same" do
+      it "raises an exception if sorting with nil primary keys" do
         Review.clear
         reviews = [] of Review
         100.times do |num|
           review = new_review({downvotes: num})
           reviews << review
         end
-        review_downvotes = reviews.map{|review| review.downvotes.as(Int32)}.sort
-        sorted = reviews.shuffle.sort.map(&.downvotes)
-        review_downvotes.should eq(sorted)
+        expect_raises(Exception) do
+          sorted = reviews.shuffle.sort
+        end
       end
 
     end
 
     describe "<=>" do
-      it "compares records by id before other attributes" do
+      it "compares records with primary keys" do
         r1 = create_review({downvotes: 50})
         r2 = create_review({downvotes: 10})
-        (r1 <=> r2 < 0).should be_truthy
+        (r1 <=> r2 < 0).should_not eq(0)
       end
 
-      it "compares other fields if the ids are the same" do
-        r1 = new_review({downvotes: 50})
-        r2 = new_review({downvotes: 10})
-        (r1 <=> r2 > 0).should be_truthy
+      it "raises an exception if comparing a record without a primary key" do
+        r1 = new_review
+        r2 = create_review
+        expect_raises(Exception) do
+          r1 <=> r2
+        end
       end
 
-      it "compares Bool fields" do
-        r1 = create_review({published: true})
-        r2 = create_review({published: false})
-        (r1 <=> r2).should_not eq(0)
-      end
-
-      it "compares belongs_to field ids" do
-        teacher_1 = Teacher.create(name: "Test Teacher")
-        teacher_2 = Teacher.create(name: "Test Teacher")
-        klass_1 = Klass.new(name: "Test Class")
-        klass_1.teacher = teacher_1
-        klass_2 = Klass.new(name: "Test Class")
-        klass_2.teacher = teacher_2
-        (klass_1 <=> klass_2).should_not eq(0)
-      end
     end
   end
 end
--- a/spec/granite/comparisons/comparisons_spec.cr
+++ b/spec/granite/comparisons/comparisons_spec.cr
@@ -19,66 +19,98 @@ module {{adapter.capitalize.id}}
     review
   end
 
-  describe "{{ adapter.id }} ==" do
-    it "should correctly compare two identical reviews" do
-      r1 = new_review
-      r2 = new_review
-      (r1 == r2).should be_truthy
+  describe "{{ adapter.id }} Granite::Base" do
+
+    describe "==" do
+      it "correctly compares two identical reviews" do
+        r1 = new_review
+        r2 = new_review
+        (r1 == r2).should be_truthy
+      end
+
+      it "correctly compares two identical records with different ids" do
+        r1 = create_review
+        r2 = create_review
+        (r1 == r2).should_not be_truthy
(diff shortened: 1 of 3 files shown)


spec/granite/comparisons/comparisons_spec.cr
```diff
@@ -0,0 +1,116 @@
...
+        review_downvotes.should eq(sorted)
+      end
+
+    end
+
+    describe "<=>" do
+      it "compares records by id before other attributes" do
+        r1 = create_review({downvotes: 50})
+        r2 = create_review({downvotes: 10})
+        (r1 <=> r2 < 0).should be_truthy
```
robacarp: I'm not sure I can think about all the ways a record could be greater than or less than another record. For example, you have here a record which is created later, and therefore has a greater id. So one field is greater (id) and the other field is lesser (downvotes)...I don't think it's safe to say that they are greater or lesser than each other.

spec/granite/comparisons/comparisons_spec.cr
```diff
@@ -0,0 +1,116 @@
+require "../../spec_helper"
+
+{% for adapter in GraniteExample::ADAPTERS %}
+module {{adapter.capitalize.id}}
```
robacarp: I really appreciate you following suite with the tests here, @astupidnerd! 

There is a pending pull which simplifies this syntax greatly and runs tests across different adapters in docker/CI instead of requiring devs to have all the adapters installed and configured to run tests.

src/granite/comparator.cr
```diff
@@ -0,0 +1,35 @@
...
+    def_equals_and_hash {{ *FIELDS.keys }}
+
+    # Compares this {{@type}} to another {{@type}} by comparing their fields in this order: {{ FIELDS.keys.join(", ").id }}.
+    #
+    # Each individual field will be compared using the <=> (if available).
+    #
+    # Returns 0 if the two objects are equal, a negative number if this object is considered less than other, or a positive number otherwise.
+    def <=>(other : {{@type}})
+      {% for field, options in FIELDS %}
+        if self.{{field}} && !other.{{field}}
```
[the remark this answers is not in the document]
astupidnerd: Yeah, that makes sense to me. I'll remove the field comparisons and just check the primary keys.

I was struggling with what to do when there are two unsaved records (without an id) that have different values set for some fields.

Should those be considered equal?

For example:

```crystal
a = Review.new({downvotes: 50})
b = Review.new({downvotes: 40})
a <=> b #=> 0?
```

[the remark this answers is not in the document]
astupidnerd: Oh yeah, according to this issue I opened up with the crystal repo (https://github.com/crystal-lang/crystal/issues/6608), `Array#sort` will soon accommodate checking for `nil` being returned from the `<=>` operator, indicating that two objects are not comparable.

I think having `<=>` return `nil` when the primary keys are `nil` would make a lot of sense. What do you think?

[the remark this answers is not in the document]
robacarp: @astupidnerd I think that's a great idea, thanks for looking into that

spec/granite/comparisons/comparisons_spec.cr
robacarp: @astupidnerd yes. There are some weird corner cases when people use uuid's as their primary key, but I think that can be ignored. If people are doing that, they're unlikely to be comparing the records without also implementing their own <=>

astupidnerd:
@robacarp I removed checks against all other fields and only compared the primary keys. In the case where a primary key is `nil`, or does not implement the `<=>` itself, it raises an exception.

Once https://github.com/crystal-lang/crystal/issues/6608 is integrated, we should change the method to return `nil` instead.

Feedback is always appreciated; please let me know if there is anything else that should change.

spec/granite/comparisons/comparisons_spec.cr
astupidnerd: I'm happy to update the tests if need be. :)

src/granite/comparator.cr
```diff
@@ -0,0 +1,32 @@
...
+      pk_field = PRIMARY[:name].id
+    %}
+
+    {% if pk_type_node.has_method?("<=>") %}
+    # Compares this `{{@type}}` to another `{{@type}}` by comparing their primary keys (`{{pk_field}}`).
+    #
+    # Returns 0 if the two primary keys are equal, a negative number if this object's primary key is considered less than other's, or a positive number otherwise.
+    {% else %}
+    # Raises an exception; the `{{pk_field}}` field does not implement the `<=>` operator.
+    {% end %}
```
robacarp: I think the code below is pretty straightforward and doesn't need clarifying comments, so this can be removed.

src/granite/comparator.cr
```diff
@@ -0,0 +1,32 @@
+# Adds equals, hash, and comparison methods to Granite::Base objects.
+module Granite::Comparator
+  macro __process_comparator
+
+    def_equals_and_hash {{ *FIELDS.keys }}
```
c910335: Both `Comparable` and `#def_equals_and_hash` define `#==`, so I think this line is redundant.

src/granite/comparator.cr
```diff
@@ -0,0 +1,32 @@
...
+    # Compares this `{{@type}}` to another `{{@type}}` by comparing their primary keys (`{{pk_field}}`).
+    #
+    # Returns 0 if the two primary keys are equal, a negative number if this object's primary key is considered less than other's, or a positive number otherwise.
+    {% else %}
+    # Raises an exception; the `{{pk_field}}` field does not implement the `<=>` operator.
+    {% end %}
+    def <=>(other : {{@type}})
+      raise "Cannot compare two {{@type}} objects if either's `{{pk_field}}` is `nil`." if self.{{pk_field}}.nil? || other.{{pk_field}}.nil?
+      {% if pk_type_node.has_method?("<=>") %}
+        self.{{pk_field}}.as({{pk_type}}) <=> other.{{pk_field}}.as({{pk_type}})
```
drujensen: Comparing primary keys doesn't seem like it would provide a good determination if something is greater than or less than another.  This would be something that would need the business context to know how to perform this.  I agree with @c910335 that we need to allow the users to override this.

I'm going to close this PR since I believe it needs some rethinking.  Thanks for your contribution.

PR #277, opened 2018-08-29
Dj/lazy load connection

drujensen:
This pr changes the way the connection is created.  Before, the connection would be created when the Adapter is created.  This change will create the connection the first time a database connection is attempted.  Subsequent calls will use the same connection.

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -7,12 +7,14 @@ require "db"
 abstract class Granite::Adapter::Base
   getter name : String
   getter url : String
-  private property database : DB::Database
 
   def initialize(connection_hash : NamedTuple(url: String, name: String))
     @name = connection_hash[:name]
     @url = connection_hash[:url]
-    @database = DB.open(@url)
+  end
+
+  def database
+    @database ||= DB.open(@url)
   end
 
   def open(&block)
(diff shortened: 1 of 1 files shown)


src/adapter/base.cr
```diff
@@ -7,12 +7,15 @@ require "db"
...
-  private property database : DB::Database
+  private property _database : DB::Database?
 
   def initialize(connection_hash : NamedTuple(url: String, name: String))
     @name = connection_hash[:name]
     @url = connection_hash[:url]
-    @database = DB.open(@url)
+  end
+
+  def database : DB::Database
```
eliasjpr: Should this be a private method as well?

eliasjpr: Also I think the database part should live in a module maybe, just maybe.

```
module Adapter
  module Connection 
    private property _database : DB::Database? = DB.open(@url)
  end 
end
```

drujensen: > Should this be a private method as well?
yeah, i'm not sure.  Getting access to the underlying database could be handy but its not something we should document or encourage. I left it public but not against making it private.

> Also I think the database part should live in a module maybe, just maybe.
it could but I would avoid the name `Connection` because that might be confusing.  The database object itself has connections, so `Connection::Database.connections` is weird.

Issue #129, opened 2018-01-27
Fix pluralization issues with relationships


drujensen:
Look into integration of an inflector: https://github.com/luckyframework/lucky_inflector

EDIT:
I think I agree with @shayneoneill that this is one feature in rails that creates a lot of pain.

I think a better solution is to default to singular but allow for overriding the name so:
```
belongs_to :user
has_many :car
```

This would use the `Car` class and default to `car_id`.   

If someone wanted to use the plural form:
```
belongs_to :user
has_many :cars, class_name: Car
```

You would be required to specify the class name.  We should also support the `foreign_key` to override the default.

shayneoneill:
Is this something you really want? I always considered this an anti-feature in rails, an unneccesary form of magic that got in the way of readability and seemed to confuse ides

drujensen:
@shayneoneill PR #278 removes the pluralization of the table name and the has_many relationship.  It always felt like a hack to me and removing it feels great!  Take a look at the PR and provide any feedback.  Thanks!

shayneoneill:
perfect.

PR #278, opened 2018-09-04
Dj/129 has many

drujensen:
This PR addresses issue #129 by removing the automatic pluralization (adding `s`) to the table name.  I think the industry agrees this is a bad practice.  If you want to pluralize the name, you need to specify the `table_name` explicitly.   

It also changes the `has_many` relation to remove stripping of the `s` and add support to allow you to use a type declaration as follows:
```crystal
  has_many :post

  # pluralization requires providing the class name
  has_many posts : Post

  # or you can provide class name as a parameter
  has_many :posts, class_name: Post

```
This eliminates the need for adding a inflector to the macro level.  This is a breaking change and moves away from the Rails pluralization.  You can still pluralize if you want but you need to explicitly specify the type if you do.

This PR also allows you to specify a custom `foreign_key`:
```crystal
  has_many :posts, class_name: Post, foreign_key: :custom_id
```

This PR also refactors the `has_one` to use the double splat for options so they can be provided in any order.

### DIFF
--- a/src/granite/associations.cr
+++ b/src/granite/associations.cr
@@ -52,26 +52,20 @@ module Granite::Associations
   end
 
   macro has_many(model)
-    {% if model.is_a? TypeDeclaration %}
-      def {{model.var}}
-          Granite::AssociationCollection(self, {{model.type}}).new(self)
-      end
-    {% else %}
-      def {{model.id}}
-          Granite::AssociationCollection(self, {{model.id.camelcase}}).new(self)
-      end
-    {% end %}
+    def {{model.id}}
+      Granite::AssociationCollection(self, {{model.id.camelcase}}).new(self)
+    end
   end
 
-  macro has_many(model, through)
-    {% if model.is_a? TypeDeclaration %}
-      def {{model.var}}
-          Granite::AssociationCollection(self, {{model.type}}).new(self, {{through}})
-      end
-    {% else %}
-      def {{model.id}}
-          Granite::AssociationCollection(self, {{model.id.camelcase}}).new(self, {{through}})
-      end
-    {% end %}
+  macro has_many(model, model_name)
+    def {{model.id}}
+      Granite::AssociationCollection(self, {{model_name.id}}).new(self)
+    end
+  end
+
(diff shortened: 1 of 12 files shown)


c910335:
I think the syntax is not intuitive. `cats` is obviously a collection of `Cat` but not one `Cat`.
It would be better for me to support the syntax in #129.

```crystal
has_many cats, class: Cat
```

src/granite/associations.cr
```diff
@@ -52,26 +52,20 @@ module Granite::Associations
...
-    {% if model.is_a? TypeDeclaration %}
-      def {{model.var}}
-          Granite::AssociationCollection(self, {{model.type}}).new(self, {{through}})
-      end
-    {% else %}
-      def {{model.id}}
-          Granite::AssociationCollection(self, {{model.id.camelcase}}).new(self, {{through}})
-      end
-    {% end %}
+  macro has_many(model, model_name)
```
robacarp: `model` is no longer really descriptive here, is it more of the name of the getter?

drujensen:
@robacarp I like the type declaration as well but I agree with @c910335  that `has_many cats : Cat` is confusing since it really is a collection and not a single instance.

Please re-review this as I have removed the type declaration in favor of `class_name`.   `class` is a reserved word.  Also Rails uses `class_name`. 

I have also implemented `foreign_key` so now you have can multiple `has_many` relationships to the same model using different foreign keys.

drujensen:
@c910335 I'm thinking of adding back the type declaration `has_many cats : Cat`.  The has_many implies its a collection and this reads well to me:  `has many cats of type Cat`.  WDYT?  Is it too confusing?  We can support both syntaxes.

drujensen:
I"m reworking the `belongs_to` to use splats for more flexibility when defining it.

drujensen:
I added support for both syntaxes.  Let me know what you think.

drujensen:
ping!

drujensen:
I'm merging this if there is no feedback.  Let me know if we need any changes.

Issue #234, opened 2018-06-18
validation helper specs need to be brought into modularized style


robacarp:

Thellior:
Do you mean that instead of an validator_helper_spec we should create spec files like validate_is_nil_spec.cr validate_is_blank_spec.cr?

robacarp:
@Thellior good question. If the spec files are long enough, that's a great way to make them more readable.

These are the changes I'd like to see:

-  the presence of test models in the spec files (spec/spec_models.cr has _plenty_ and they should be used whenever possible)
- not using the [GraniteExample::Adapters pattern](https://github.com/amberframework/granite/blob/master/spec/granite/querying/find_by_spec.cr#L3-L5) to enforce that specs are run on all adapters.

Blacksmoke16:
Reason i put them with the spec is they all contain validation helpers on them.  I didn't want other thinking they could be used in other normal tests and wonder why they fail if one of the validation helpers constraints isnt met.  

I'm not opposed to moving them all into the `spec_models.cr` file and naming them `ValidationHelperxxxx` however.

drujensen:
I believe this has been done.  closing this.

PR #283, opened 2018-09-28
Refactored the validator helpers and added exclusion validator

Thellior:
<h1>What is done</h1>

- Refactored validator helpers to own folder. Every validator category is separated in his/her own file. Test models are no in spec_models and they are inside the Validators module. Specs are also separated in validator category.
- Added exclusion validator. It will validate if the current value is not inside the reserved values


Thellior:
@robacarp I think this fixes your issue https://github.com/amberframework/granite/issues/234

PR #284, opened 2018-10-06
Add limit and offset to Builder/Assembler with some fixes

c910335:
I found that the query builder which was introduced in #132 is very powerful when I was building [Shale](https://github.com/c910335/shale).
But I think it needs some improvement.

1. Fix typo "include" => "extend" in `src/granite/query/builder_methods.cr`.
2. Fix the weird whitespaces in the log.
Thanks @Thellior for finding this bug.
3. little refactoring
4. Introduce limit and offset.
They are very useful for pagination.


drujensen:
@robacarp are you ok with merging this?  I have time to do a patch release today and would like to include these if you approve.

robacarp:
@drujensen sure, if you like. I think there's room  for more refactor along the comments I made, but it's functional as is, I think

src/granite/query/assemblers/base.cr
```diff
@@ -26,6 +26,12 @@ module Granite::Query::Assembler
       [Model.fields].flatten.join ", "
     end
 
+    def build_sql
+      clauses = [] of String?
+      yield clauses
+      clauses.compact!.join " "
+    end
```
c910335: @robacarp 
What about this?

PR #285, opened 2018-10-09
reverse query builder

drujensen:
Because SQL is fairly standard and the only differences are related to
param replacement, this reverses the base class to contain the main code
and moves the differences to the driver specific class.

### DIFF
--- a/src/granite/query/assemblers/base.cr
+++ b/src/granite/query/assemblers/base.cr
@@ -1,14 +1,17 @@
 module Granite::Query::Assembler
   abstract class Base(Model)
+    @where : String?
+    @order : String?
+    @limit : String?
+    @offset : String?
+    @group_by : String?
+
     def initialize(@query : Builder(Model))
       @numbered_parameters = [] of DB::Any
       @aggregate_fields = [] of String
     end
 
-    def add_parameter(value : DB::Any) : String
-      @numbered_parameters << value
-      "$#{@numbered_parameters.size}"
-    end
+    abstract def add_parameter(value : DB::Any) : String
 
     def numbered_parameters
       @numbered_parameters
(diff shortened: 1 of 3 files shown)


drujensen:
@robacarp @c910335 wdyt?

robacarp:
@drujensen Great idea! I like it.

There are a bunch of places where the SQL standard is insufficient and databases do weird things, but I think this model will allow for super calls, etc to override that when needed. More importantly it'll provide the query syntax to all supported drivers in the short term... which I love!

c910335:
[Granite::Query::Builder#assembler](https://github.com/amberframework/granite/blob/bbe603fb10fb85ff4df51b3edce5093cfe4d89b6/src/granite/query/builder.cr#L35)
Also enable them here?

drujensen:
> [Granite::Query::Builder#assembler](https://github.com/amberframework/granite/blob/bbe603fb10fb85ff4df51b3edce5093cfe4d89b6/src/granite/query/builder.cr#L35)
> Also enable them here?

@robacarp @c910335  Any recommendations on how to do this?

drujensen:
I tried something like this:
```
    case Model.adapter.class
    when Granite::Adapter::Pg
      Assembler::Postgresql(Model).new self
    when Granite::Adapter::Mysql
      Assembler::Mysql(Model).new self
    when Granite::Adapter::Sqlite
      Assembler::Sqlite(Model).new self
    else
      raise "Query builder not supported for #{adapter.class}"
    end
```
but the `Model.adapter` class_getter is not found and not sure why yet.

robacarp:
@drujensen what about modifying the adapter macro to define the method instead?

c910335:
https://github.com/amberframework/granite/blob/13d5fc050c71e6109a84857def11542e4f638ebe/spec/granite/query/spec_helper.cr#L5-L17

Because there is no `self.adapter`.

drujensen:
@c910335 @robacarp I like that the test works against a generic model but I think we will need to test the differences between the db's.  Do you agree that I should add the adapter to the model?

drujensen:
> @drujensen what about modifying the adapter macro to define the method instead?

Instead of the case statement?  Can you give me an example of what you are envisioning?

drujensen:
FYI, I added support for operators in the where clause.  It still supports parameter list as well:
```
 builder.where(:name, :like, "bob*").where(:age, :gteq, 24).where(occupation: "developer")
```

src/granite/query/assemblers/base.cr
```diff
@@ -32,8 +35,156 @@ module Granite::Query::Assembler
...
+        s << order
+        s << limit
+        s << offset
+      end
+
+      Executor::List(Model).new sql, numbered_parameters
+    end
+
+    def sql_operator(operator : Symbol) : String
+      case operator
```
c910335: Use `NamedTuple` instead of `case`?

```crystal
OPERATORS[operator]? || operator.to_s
```

src/granite/query/assemblers/sqlite.cr
```diff
@@ -0,0 +1,8 @@
+module Granite::Query::Assembler
+  class Sqlite(Model) < Base(Model)
+    def add_parameter(value : DB::Any) : String
+      @numbered_parameters << value
+      "?"
+    end
```
robacarp: The semantics of this are a bit strange (I'm pretty sure that's my fault, not yours). This method shifts a value onto a list, and then returns a different value which ends up in the query. Is there a better way that can be arranged?

robacarp:
Interesting dsl for operators, and an easy implementation too. I like that.

For the macro suggestion earlier, forget that. I don't know what I was thinking.

Instead, could [this code](https://github.com/amberframework/granite/blob/master/src/granite/query/builder_methods.cr#L7-L13) can be modified to create a Builder that's already specified for the adapter type?

drujensen:
@robacarp @c910335 thanks for your last review.  Can you go over this again?  I didn't want the query builder to have a dependency on the Adapter so I added an enum type for the different supported DB's that needs to be specified when created.  Is there a better way to do this?

Also, I'm not completely happy that the Builder and Assembler are both dependent on each other but couldn't find a good way to break it.  

Finally, I think we should consider merging the `query` and `querying` modules and adding these methods automatically.

Let me know what you think.

src/granite/query/assemblers/sqlite.cr
drujensen: Yeah, maybe we just rename this method?  something like "replace_the_parameter_in_the_query_and_store_the_value_for_later_use_to_prevent_sql_injection`  ;-)

src/granite/query/builder.cr
```diff
@@ -16,41 +16,57 @@
...
+    case @db_type
+    when DbType::Pg
+      Assembler::Pg(Model).new self
+    when DbType::Mysql
+      Assembler::Mysql(Model).new self
+    when DbType::Sqlite
+      Assembler::Sqlite(Model).new self
+    else
+      raise "Database type not supported"
+    end
```
drujensen: should we memoize this?  Seems like it will create a new one every time its called.

src/granite/query/builder_methods.cr
```diff
@@ -6,7 +6,18 @@
...
+              when Granite::Adapter::Sqlite
+                DbType::Sqlite
+              else
+                raise "Adapter not supported #{Model.adapter.class}"
+              end
+
+    Builder(self).new(db_type)
   end
 
   delegate where, count, order, offset, limit, first, to: __builder
```
drujensen: @robacarp when I try to `extend Granite::Query::BuilderMethods`, I get an error for the `first` method since it's defined in the `querying.cr` module.  It looks like `count` is also defined in both.  

What is the plan for integration?  Did we want to move all the current `querying` functionality over?  Should we look at adding the lazy loaded `Collection` into the executor?

c910335:
> Also, I'm not completely happy that the Builder and Assembler are both dependent on each other but couldn't find a good way to break it.
> Finally, I think we should consider merging the query and querying modules and adding these methods automatically.

The new BuilderMethods+Builder+Assembler+Executor design plays a similar role as Querying+Adapter currently.
And I think the former is the direction that we should go.
Thus, merging the latter into the former is what I want to do.
It will definitely take a lot of time though.

drujensen:
@c910335 are you good with me merging this?  We can iterate on moving functionality over starting with `first` and `count`

PR #286, opened 2018-10-14
Support PG Array types

Blacksmoke16:
Adds support for PG array types

Was surprisingly easy this time around....


Blacksmoke16:
Specs are failing on crystal `26.1` ill revert version back to `26.0` and look into that separately

PR #289, opened 2018-10-14
Annotation support

Blacksmoke16:
This PR adds the ability to supply third-party annotations to be applied to each field's property declaration.

I needed this for https://github.com/Blacksmoke16/CrSerializer so that a code update didn't have to be made for each third party library that wanted to be able to define annotations on our property.

This PR also:
- Cleans up the `primary` macro to use double spat
- Adds a shell script to run specs for each adapter
- Fixes readme, left justifies everything
- Move the JSON/YAML serialization modules out from macro finished, into macro inherited so to not override model specific includes.

Having this also kinda makes me want to remove:
```Crystal
      # Apply JSON/YAML serialization option annotations
      {% if options[:json_options] %}
         @[JSON::Field({{**options[:json_options]}})]
      {% end %}
      {% if options[:yaml_options] %}
         @[YAML::Field({{**options[:yaml_options]}})]
      {% end %}
```
but not sure if its worth the breaking change.  But it would remove some logic that we have to maintain, and move it to up to the user supplying them.

WDYT?


docs/getting_started.md
```diff
@@ -144,3 +144,20 @@ The `field` and `primary` macros have a comment option that will specify the doc
...
+
+Annotations can be a powerful method of adding property specific features with minimal amounts of code.  Since Granite utilizes the `property` keyword for its fields, annotations are able to be applied easily.  This is accomplished by using the `annotations` option on a field, similar to the `comment` option above.  It is used in the form of `annotations: ["annotation1", "annotation2"]`, for example:
+
+```Crystal
+class Foo < Granite::Base
+  adapter mysql
+  table_name foos
+
+  field name : String
+  field password : String, annotations: ["@[Foo::Options(option1: true)]", "@[Bar::Settings(other_option: 7)]"]
```
drujensen: @Blacksmoke16 Is it possible to apply the annotations to the field?
```
@[Foo::Options(option1: true)]
field password : String
```

spec/run_all_specs.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/bash
+
+source ../.env
+
+echo "Testing PG"
+
+docker-compose -f docker/docker-compose.pg.yml build spec
+docker-compose -f docker/docker-compose.pg.yml run spec
```
drujensen: I think you can use `--rm` to remove the containers after running it.

spec/run_all_specs.sh
```diff
@@ -0,0 +1,24 @@
...
+docker-compose -f docker/docker-compose.mysql.yml run spec
+
+echo "Testing sqlite"
+
+docker-compose -f docker/docker-compose.sqlite.yml build spec
+docker-compose -f docker/docker-compose.sqlite.yml run spec
+
+echo "Done testing...stopping/removing images"
+
+docker-compose -f docker/docker-compose.sqlite.yml down
```
drujensen: I think you can remove these if you use the `--rm` flag in the run.

src/granite/table.cr
```diff
@@ -27,32 +27,11 @@ module Granite::Table
     {% SETTINGS[:table_name] = name.id %}
   end
 
-  # specify the primary key column and type
-  macro primary(decl)
+  macro primary(decl, **options)
```
drujensen: 💯

docs/getting_started.md
Blacksmoke16: I don't think so unfortunately.  I'll try it again but i think it messes with it since its a macro and not the actual property.

spec/run_all_specs.sh
Blacksmoke16: I checked this out, `--rm` removes the `spec` container only, and leaves the corresponding db container.  Doing `down` at the end stops the db container, removes the spec and db container and removes the network created for the run.

Blacksmoke16: @drujensen We possibly could replace it with a `docker system prune` to `Remove all unused containers, networks, images (both dangling and unreferenced)`

However it could possibly delete stopped containers other than Granite ones.

WDYT?

drujensen: @Blacksmoke16 yeah, I do this all the time.  :+1:

docs/getting_started.md
drujensen: bummer

spec/run_all_specs.sh
drujensen: I think its fine the way it is.  going to merge this.

docs/getting_started.md
Blacksmoke16: Kinda late, but I revisited this.  It _is_ possible to apply annotations to properties created via a macro.  However it will _not_ work with Granite since our `field` macro just adds data to a mutable constant and not actually defines the property.

PR #292, opened 2018-10-16
update query documentation

drujensen:
This adds the new Query Builder methods to the documentation.

### DIFF
--- a/docs/crud.md
+++ b/docs/crud.md
@@ -82,7 +82,7 @@ if posts
 end
 ```
 
-See [querying](./querying.md) for more details on using `all`
+See [querying](./querying.md#all) for more details on using `all`
 
 ## Update
 
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -6,7 +6,7 @@ The query macro and where clause combine to give you full control over your quer
 
 Where is using a QueryBuilder that allows you to chain where clauses together to build up a complete query.
 ```crystal
-posts = Post.where(published: true, author_id: User.first.id)
+posts = Post.where(published: true, author_id: User.first!.id)
 ```
 
 It supports different operators:
@@ -15,27 +15,6 @@ Post.where(:created_at, :gt, Time.now - 7.days)
 ```
 
 Supported operators are :eq, :gteq, :lteq, :neq, :gt, :lt, :nlt, :ngt, :ltgt, :in, :nin, :like, :nlike
-
-
-This allows you create snippets of clauses that can be easily chained together similar to scopes in Rails.
-```crystal
-class Post < Granite::Base
-...
-  def live
-    where(published: true)
-  end
-
-  def draft
-    where(published: false)
-  end
-
-  def last_week
-    where(:created_at, :gt, Time.now - 7.days)
-  end
-end
-
-posts = Post.live.last_week
-```
 
 ## Order
 
@@ -46,7 +25,7 @@ Post.order(:created_at)
 
 Direction
 ```crystal
-Post.order(updated_at: Sort::Descending)
+Post.order(updated_at: :desc)
 ```
 
 Multiple fields
@@ -56,7 +35,7 @@ Post.order([:created_at, :title])
 
 With direction
 ```crystal
-Post.order(created_at: Sort::Descending, title: Sort::Ascending)
+Post.order(created_at: :desc, title: :asc)
 ```
 
 ## Limit
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -1,21 +1,21 @@
 ## Documentation
 
-[Getting Started](docs/getting_started.md)
+[Getting Started](./getting_started.md)
 
-[CRUD](docs/crud.md)
+[CRUD](./crud.md)
 
-[Querying](docs/querying.md)
+[Querying](./querying.md)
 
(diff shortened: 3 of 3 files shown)


docs/readme.md
```diff
@@ -0,0 +1,21 @@
...
+
+[Callbacks](docs/callbacks.md)
+
+[Migrations](docs/migrations.md)
+
+[Imports](docs/imports.md)
+
+[JSON Support](docs/json_support.md)
+
+[YAML Support](docs/yaml_support.md)
```
Blacksmoke16: All these will need to have `docs/` removed since they are in same directly as the files now.

docs/querying.md
```diff
@@ -2,10 +2,83 @@
 
 The query macro and where clause combine to give you full control over your query.
 
+## Where
+
+Where is using a QueryBuilder that allows you to chain where clauses together to build up a complete query.
+```crystal
+posts = Post.where(published: true, author_id: User.first.id)
```
Blacksmoke16: For sake of working examples, probably should use `User.first!.id`, otherwise this would most likely throw the `no method id on (Int32 | Nil)` error

docs/querying.md
```diff
@@ -2,10 +2,83 @@
...
+...
+  def live
+    where(published: true)
+  end
+
+  def draft
+    where(published: false)
+  end
+
+  def last_week
```
Blacksmoke16: In these scopes (not really related to this PR) is there way to declare them as not_nil similar to the `field!` macro?  Or do nil values have to be handled on the user's end?

docs/crud.md
```diff
@@ -77,6 +82,7 @@ if posts
 end
 ```
 
+See [querying](./querying.md) for more details on using `all`
```
Blacksmoke16: Can we make this one link directly to the `all` section of querying?

docs/querying.md
```diff
@@ -2,10 +2,83 @@
...
+```crystal
+posts = Post.where(published: true, author_id: User.first.id)
+```
+
+It supports different operators:
+```crystal
+Post.where(:created_at, :gt, Time.now - 7.days)
+```
+
+Supported operators are :eq, :gteq, :lteq, :neq, :gt, :lt, :nlt, :ngt, :ltgt, :in, :nin, :like, :nlike
```
robacarp: @drujensen this is outside the scope of this pull, but I was just thinking. Since these are symbols, could they also be more literal operators: `:==` `:>=` etc?

docs/querying.md
```diff
@@ -2,10 +2,83 @@
...
+  def draft
+    where(published: false)
+  end
+
+  def last_week
+    where(:created_at, :gt, Time.now - 7.days)
+  end
+end
+
+posts = Post.live.last_week
```
robacarp: wait, does this work?

docs/querying.md
```diff
@@ -2,10 +2,83 @@
...
+## Order
+
+Order is using the QueryBuilder and supports providing an ORDER BY clause:
+```crystal
+Post.order(:created_at)
+```
+
+Direction
+```crystal
+Post.order(updated_at: Sort::Descending)
```
robacarp: I thinks [this can just be `order(updated_at: :desc)`](https://github.com/amberframework/granite/blob/master/src/granite/query/builder.cr#L92-L94), though it gets converted to a Sort::Descending internally

docs/querying.md
drujensen: Funny you should ask.  I tried that but they are not valid symbols.  I kinda followed arel's syntax https://www.rubydoc.info/github/rails/arel/master/Arel/Predications

docs/querying.md
drujensen: Interesting question!  I think the `where` would return `Array(Model)` not `Array(Model | Nil)` so there is no need for `!` in this case, right?

docs/querying.md
drujensen: no, it doesn't. :-(  It returns the QueryBuilder, not this Object so it doesn't chain like I expected.  Is there anyway to get this to work?  Maybe with proc's?

drujensen: I will remove this example for now

docs/querying.md
Blacksmoke16: Right, which in some cases could be empty.  But that would just be an empty `Array(Model)` so nil shouldnt be an issue i'd think.

docs/querying.md
robacarp: Rails suffers from the same problem, and I think they solve it by creating a separate scope builder which is capable of dereferencing further named scope calls. 

Even if you delegate back to the model (somehow) from the QueryBuilder, it'd have to pass the builder instance into the method to be able to chain. Both of these where's start from scratch with a fresh QueryBuilder:

```crystal
def named_scope_1
   where(column: value)
end

def named_scope_2
  where(column: value)
end
```

Potentially, if [instead of returning `self`](https://github.com/amberframework/granite/blob/master/src/granite/query/builder.cr#L61), each builder method returned a dup/clone, then I could see a possibility of something like this:

```crystal
macro named_scope(name)
  def {{name}}_scope_chain(context)
    yield context
  end
end

named_scope :last_week do |c|
  c.where(date: 1.week.ago)
end
```

PR #293, opened 2018-10-17
Better UUID support

### LINKED ISSUE #210: Provide primary auto: UUID option that leverages Postgresql native support
`abstract def insert(table_name, fields, params, lastval) : Int64`

`insert` is limited to return `Int64`

Blacksmoke16:
Resolves #210 

Allows the primary macro's `auto` to take a `:uuid` option.  This will generate a secure UUID on model save.  

This is compliant with all three adapters, with MySQL and SQLite using `char(36)` datatype for migrator, while Postgres uses its native `UUID` type.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -118,21 +118,20 @@ end
 
 ### UUIDs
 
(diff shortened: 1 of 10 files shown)


drujensen:
This is great!  I am tempted to change the Amber generators to use a UUID for primary instead of an auto incrementing number.

PR #297, opened 2018-10-19
Add support for AND and OR clause

drujensen:
This adds two new methods: `and` and `or`.  

```crystal
User.where(age: 24).and(gender: "male").and(:name, :like, "joe*").or(:name, :like, "fred*")
```

The where method was modified to generate the where SQL with appropriate clauses.  Let me know if you think there is a better way to do this.

I will add specs soon.

### DIFF
--- a/src/granite/query/assemblers/base.cr
+++ b/src/granite/query/assemblers/base.cr
@@ -38,18 +38,12 @@ module Granite::Query::Assembler
     def where
       return @where if @where
 
-      first = true
-      clauses = [] of String
+      clauses = ["WHERE"]
 
       @query.where_fields.each do |expression|
         add_aggregate_field expression[:field]
 
-        if first
-          first = false
-          clauses << "WHERE"
-        else
-          clauses << expression[:join].to_s.upcase
-        end
+        clauses << expression[:join].to_s.upcase unless clauses.size == 1
 
         if expression[:value].nil?
           clauses << "#{expression[:field]} IS NULL"
@@ -58,7 +52,7 @@ module Granite::Query::Assembler
         end
       end
 
-      return nil if clauses.none?
+      return nil if clauses.size == 1
 
       @where = clauses.join(" ")
     end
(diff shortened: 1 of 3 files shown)


src/granite/query/assemblers/base.cr
```diff
@@ -38,19 +38,29 @@ module Granite::Query::Assembler
...
-          "#{expression[:field]} #{sql_operator(expression[:operator])} #{add_parameter expression[:value]}"
+          clauses << "#{expression[:field]} #{sql_operator(expression[:operator])} #{add_parameter expression[:value]}"
         end
       end
 
       return nil if clauses.none?
 
-      @where = "WHERE #{clauses.join " AND "}"
+      @where = clauses.join(" ")
     end
```
drujensen: There is probably a better, less empirical way to do this.  Please use github's  new `suggestion` feature to provide a solution.

src/granite/query/assemblers/base.cr
```diff
@@ -38,19 +38,29 @@ module Granite::Query::Assembler
...
+
+      @query.where_fields.each do |expression|
         add_aggregate_field expression[:field]
 
+        if first
+          first = false
+          clauses << "WHERE"
+        else
+          clauses << expression[:join].to_s.upcase
+        end
```
Blacksmoke16: Could this be cleaned up by like:
```Crystal
clauses << clauses.includes?("WHERE") ? expression[:join].to_s.upcase : "WHERE"
```

Blacksmoke16: Or why not just initialize the clauses array as ` = ["WHERE"]`?

robacarp:
This is tricky. `or` is important, but so is order of operations. `.where(A).or(B)` is relatively straightforward, but what about `.where(A).or(B).and(C)`? Is it `(A or B) and C` Or `A or (B and C)`? 

Regardless of which it is, how do I specify the other combination? Composition is necessary with multiple Boolean operators.

PR #299, opened 2018-10-21
Fix Assembler#count

c910335:
`Assembler#count` appends `ORDER BY id DESC` to sql unintentionally since #284.
Sorry for making this bug and here is the fix.


Blacksmoke16:
Prob wouldn't be a bad idea to add a spec to this/other query builder stuff to verify it generates correct SQL?

c910335:
@Blacksmoke16 I have also fixed the spec.

PR #301, opened 2018-10-26
Only define `updated/created_at` properties if they are wanted

Blacksmoke16:
Adds specs for the `json_options` and `yaml_options` features

Also only includes `@updated_at` and `@created_at` properties if they were added via the `field` or `timestamps` macros.

This _could_ break people's apps who were using those without declaring them, maybe?  But i doubt it.

### DIFF
--- a/spec/granite_spec.cr
+++ b/spec/granite_spec.cr
@@ -2,7 +2,7 @@ require "./spec_helper"
 
 describe "Granite::Base" do
   describe "JSON" do
-    describe "#from_json" do
+    describe ".from_json" do
       it "can create an object from json" do
         json_str = %({"name": "json::anyReview","upvotes": 2, "sentiment": 1.23, "interest": 4.56, "published": true})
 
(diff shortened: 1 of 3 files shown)


spec/granite_spec.cr
```diff
@@ -2,7 +2,7 @@ require "./spec_helper"
 
 describe "Granite::Base" do
   describe "JSON" do
-    context ".from_json" do
+    describe "#from_json" do
```
c910335: This should be `.from_json` because it is a class method but not instance method.

spec/granite_spec.cr
```diff
@@ -68,10 +68,22 @@ describe "Granite::Base" do
...
+
+      it "should serialize correctly" do
+        model.to_json.should eq %({"task_name":"The Task"})
+      end
+    end
   end
 
   describe "YAML" do
-    context ".from_yaml" do
+    describe "#from_yaml" do
```
c910335: ditto

spec/granite_spec.cr
Blacksmoke16: Yes, yes it should.

PR #304, opened 2018-10-28
Belongs_to refactor

Blacksmoke16:
This PR refactors the `belongs_to` macro to be in the new world of double spat args.

It also adds json/yaml_option supports to the `belongs_to` macro to hide/change FK key name etc on serialization.  

NOTE: Specs on this will fail until #301 is merged since `updated_at` and `created_at` are serialized even though they were never declared (using the field or timestamps macros).


drujensen:
@Blacksmoke16 good stuff! thanks.

PR #305, opened 2018-10-29
Include clause when using custom select statement

hfjallemark:


drujensen:
@hfjallemark Thanks for this PR.  Can you do the same for mysql and sqlite adapters?

Blacksmoke16:
@drujensen is there a reason why `#select` isnt in the base class?  Looks like it would be the same for all adapters?

drujensen:
@Blacksmoke16 not that i'm aware of.  We could look at refactoring some of the other methods as well.

hfjallemark:
I moved `#select` into the base class instead. I tried to override `#select` in the pg adapter to run `ensure_clause_template` but wasn't able to call `super` with the `&block` parameter (I'm pretty new to Crystal) so move `ensure_clause_template` into the base class and overrid it in the pg adapter instead. Feel free to change if you want!

Had a quick look at the other methods but noticed some small differences between them so didn't touch them in this pr.

PR #306, opened 2018-11-05
Not nil field refactor

### LINKED ISSUE #298: Not Nil fields refactor
Having the `field` vs `field!` to determine if a column is not nil doesn't feel crystal like to me.  

IMO, wouldn't it be more clear to base that off of the type of the field?  For example:
```Crystal
class Test < Granite::Base
  adapter pg
  
  field name : String # Not nil
  field is_attending : Bool? # Nilable
end
```

This is essentially what the `field!` macro does, but then you could end up with using the bang version of field with a type of `String`.  At a glance this makes it feel like it should be not nil, based on typing behavior elsewhere in Crystal.

Blacksmoke16:
So currently this is mostly a working prototype.  I still want to go back and maybe clean up some things, but wanted to get this out here to start getting feedback.

Resolves #298 

For the most part this worked out quite nicely, with the biggest benefit of more compile time checks when working with nullable fields.

This PR also comes with the added bonus of default values on a field.

For example i'll be using this class, and assuming that it returns a record from the database.
```crystal
class Example < Granite::Base
  adapter pg
  table_name "test"

  field name : String = "Joe"
  field user_id : Int64
  field test_id : Int32?
end

model = Example.first!

# With this PR
pp model.user_id = nil
Error in src/app.cr:50: no overload matches 'Example#user_id=' with type Nil
Overloads are:
 - Example#user_id=(user_id : Int64)

# Before this PR
pp model.user_id = nil # => nil
```

This PR retains the bang column accessor, which will raise on nil, useful for use with the PK.


src/granite/migrator.cr
```diff
@@ -48,7 +48,10 @@ module Granite::Migrator
       end
 
       def create
-        resolve = ->(key : String) {
+        resolve = ->(key : String, auto : Bool) {
+          key = key.match(/\((\w+)\,/).not_nil![1] if key.includes?("Union(")
+          key = key.chomp(" | ::Nil") if key.includes?("Nil")
+          key = "AUTO_#{key}" if auto
```
Blacksmoke16: Need to come up with a better way to handle this.  I'll think of something.

This also brings the question should the Granite migrator be responsible for setting `NOT NULL` on columns that are not nilable?  Would be added functionality to the migrator, but would require some work, best for another PR later on.

spec/spec_models.cr
```diff
@@ -9,11 +9,10 @@ require "uuid"
   {% adapter_literal = env("CURRENT_ADAPTER").id %}
 
   class Parent < Granite::Base
-    primary id : Int64
```
drujensen: I'm curious, why did you remove these?

Blacksmoke16:
More I think about it the more I'm not happy with this implementation.  For what this PR is doing I think i would rather work on migrating granite away from the mutable constants/macro magic to use annotations.

PR #307, opened 2018-11-21
Association improvements

Blacksmoke16:
* Allows custom primary keys on `belongs_to` and `has_one` associations. 
* Adds bang, raise on nil, methods to `belongs_to` and `has_one` associations. 
* Makes the `has_many` association use whatever primary key is defined on the model, vs always assuming `id`

### DIFF
--- a/spec/spec_models.cr
+++ b/spec/spec_models.cr
@@ -106,6 +106,8 @@ require "uuid"
     adapter {{ adapter_literal }}
     table_name services
 
+    has_many :couriers, class_name: Courier, foreign_key: "service_id"
+
     primary owner_id : Int64, auto: false
     field! name : String
   end
--- a/spec/spec_models.cr
+++ b/spec/spec_models.cr
@@ -83,6 +83,33 @@ require "uuid"
     table_name users
   end
 
+  class Character < Granite::Base
+    adapter {{ adapter_literal }}
+    table_name characters
+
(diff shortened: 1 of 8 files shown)


spec/spec_models.cr
```diff
@@ -83,6 +83,35 @@ require "uuid"
...
+  end
+
+  class Courier < Granite::Base
+    adapter {{ adapter_literal }}
+    table_name couriers
+
+    primary courier_id : Int32, auto: false
+    field! issuer_id : Int32
+
+    belongs_to service : CourierService, primary_key: "owner_id"
```
drujensen: @Blacksmoke16 just to confirm, can you use a Symbol instead of a String for the value?
```crystal
belongs_to service : CourierService, primary_key: :owner_id
```

drujensen: Looks like `mysql` is failing.  :-(

Blacksmoke16: Yea, ill take a look.

Should be able to, just sets it and calls `.id` on them.

Blacksmoke16: @drujensen I figured it out, see 2e80d10.

I'm wondering why this didn't come up with like `Book` or any other model that used a `Int32` or `Int64` PK with `auto: false`.

I guess no other spec had it setup like I did this one time. :man_shrugging:

PR #308, opened 2018-11-23
Aggregation functions

Blacksmoke16:
Add aggregate methods.  `sum, min, max, and avg`.

### DIFF
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -53,10 +53,8 @@ Post.offset(100).limit(50)
 ```
 ## Aggregates
 
-**NOTE:** The return type of each aggregate function can differ depending on the database adapter being used.  For example the `sum ` method:
+The return type of each aggregate function can differ depending on what method and what adapter is being used.  Review the [spec file](https://github.com/amberframework/granite/blob/master/spec/granite/querying/aggregates_spec.cr) for examples.
 
-* `MySQL` will return a `Float64` even if it is an integer.
-* `Postgres` and `Sqlite` will return them a `Int64` or `Int32` depending on the size of the integer.
 
 ### Min
 
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -51,6 +51,48 @@ Offset is using the QueryBuilder and provides the ability to offset the results.
 ```crystal
 Post.offset(100).limit(50)
 ```
+## Aggregates
+
+**NOTE:** The return type of each aggregate function can differ depending on the database adapter being used.  For example the `sum ` method:
+
+* `MySQL` will return a `Float64` even if it is an integer.
+* `Postgres` and `Sqlite` will return them a `Int64` or `Int32` depending on the size of the integer.
+
+### Min
+
+Min is using the QueryBuilder and will return the smallest value in the given column:
+
+```crytsal
+Post.min(:author_id, Int32)
+```
+### Max
+
+Max is using the QueryBuilder and will return the largest value in the given column:
+
+```crytsal
+Post.max(:author_id, Int32)
+```
+### Sum
+
+Sum is using the QueryBuilder and will return the sum of all the values in the given column:
+
+```crytsal
+Post.sum(:author_id, Int32)
+```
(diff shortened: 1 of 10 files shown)


Blacksmoke16:
Few updates on this.  There was firstly an issue with the migrator and pg where the `Float64` was being created as a `REAL` column type.  This was causing fields that were set to type `Float64` to return as `Float32`.

The other issue im coming across is that MySQL shard will default to `Int64` and `Float64` on math, including `SUM`.  I made an issue to see if anyone has any clarification/insight.  

https://github.com/crystal-lang/crystal-mysql/issues/67

docs/querying.md
```diff
@@ -51,6 +51,48 @@ Offset is using the QueryBuilder and provides the ability to offset the results.
 ```crystal
 Post.offset(100).limit(50)
 ```
+## Aggregates
+
+**NOTE:** The return type of each aggregate function can differ depending on the database adapter being used.  For example the `sum ` method:
+
+* `MySQL` will return a `Float64` even if it is an integer.
+* `Postgres` and `Sqlite` will return them a `Int64` or `Int32` depending on the size of the integer.
```
robacarp: Is there a lot of difference between adapters here? Would this documentation be worth specifying explicitly with the example of the #sum method? If so, Perhaps a table showing the return types for each method?

src/granite/query/builder.cr
```diff
@@ -153,6 +153,16 @@ class Granite::Query::Builder(Model)
     assembler.count
   end
 
+  def aggregate(field : String | Symbol, t)
+    assembler.aggregate(field, t).run
+  end
```
robacarp: I'm not sure why this method is here...why not just make the method defined in the for loop below call the assembler directly?

Blacksmoke16: :man_shrugging: Might have just assumed i need it?  I'll take a look tomorrow.

docs/querying.md
Blacksmoke16: Probably would be a good idea. :+1:

Blacksmoke16:
@robacarp It's because the type of the value coming from db will be DB::Any.  If you do not cast it you'd get errors like `undefined method 'to_i' for Bool (compile-time type is (Bool | Float32 | Float64 | Int32 | Int64 | Slice(UInt8) | String | Time | Nil))`.

Although it might be easier to just return the full union and let user deal with it.  Doesn't really matter who does it.  The expected behavior as it is now would make it seem like it would cast/convert the value to the type you give.  Would that be something we'd want/should do?

robacarp:
@Blacksmoke16 After sleeping on it, I think it's reasonable to ask the developer to specify the return type.

docs/querying.md
Blacksmoke16: Might be easier to just point the user at the specs for them to look at.  The table would be quite complex.  I don't know how to easily display the data without making a _HUGE_ table or a bunch of smaller ones, for something that can be figured out pretty quickly looking at the spec.

src/granite/query/executors/value.cr
```diff
@@ -14,7 +14,9 @@ module Granite::Query::Executor
 
       Model.adapter.open do |db|
         db.query_one? @sql, @args do |record_set|
-          return record_set.read.as Scalar
+          val = record_set.read
+          return val.to_f if val.is_a?(PG::Numeric)
```
drujensen: unfortunately, `PG::Numeric` will not exist if you don't include the pg driver.

Blacksmoke16: :confused: Ah man...

robacarp: There are ways around that. Granite can define a bare class so the const is around. [mosquito does that with granite::base](https://github.com/robacarp/mosquito/blob/master/src/external_classes.cr) and it's worked well.

Blacksmoke16: That could work.  I'll have to revisit this and try that.

PR #310, opened 2018-12-14
Add timezone support

icyleaf:
Add timezone support to set and get value with `Time` type, but the row in database is still UTC, because [sqlite](https://github.com/crystal-lang/crystal-sqlite3/blob/master/src/sqlite3.cr#L8) and [mysql](https://github.com/crystal-lang/crystal-mysql/blob/master/src/mysql.cr#L16) shards defined a constant  `TIME_ZONE` use `Time::Location::UTC`

Add `default_timezone` property to `Granite::Settings`. setter accepts two types of value: `Time::Location` and `String` but getter returns a `Time::Location` type.

Example:

```crystal
Grante.settings.default_timezone = Time::Location.local
# Or
Grante.settings.default_timezone = "Asia/Shanghai"
```

### DIFF
--- a/spec/granite/fields/casting_spec.cr
+++ b/spec/granite/fields/casting_spec.cr
@@ -1,7 +1,16 @@
+require "../../spec_helper"
+
 describe "#casting_to_fields" do
   it "casts string to int" do
     model = Review.new({"downvotes" => "32"})
     model.downvotes.should eq 32
+  end
+
+  it "casts time with timezone" do
+    Granite.settings.default_timezone = "Asia/Shanghai"
+    created_at = Time.parse("2018-12-12 00:00:00 +00:00", "%F %T %:z", Time::Location::UTC)
+    model = Review.new({"created_at" => created_at})
+    model.created_at.should eq Time.parse("2018-12-12 08:00:00+0800", Granite::DATETIME_FORMAT, Granite.settings.default_timezone)
   end
 
   it "generates an error if casting fails" do
--- a/src/granite.cr
+++ b/src/granite.cr
@@ -2,6 +2,7 @@ require "yaml"
 require "db"
 
 module Granite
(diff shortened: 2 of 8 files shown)


spec/granite/fields/casting_spec.cr
```diff
@@ -1,9 +1,22 @@
...
 
+  it "casts time with timezone" do
+    timezone = Granite.settings.default_timezone
+    Granite.settings.default_timezone = Time::Location.load("Asia/Shanghai")
+    created_at = Time.parse("2018-12-12 00:00:00 +00:00", "%F %T %:z", Time::Location::UTC)
+    model = Review.new({"created_at" => created_at})
+    model.created_at.should eq Time.parse("2018-12-11 16:00:00 +08:00", "%F %T %:z", Granite.settings.default_timezone)
+
+    # Restore previous timezone
+    Granite.settings.default_timezone = timezone
```
drujensen: Instead of restoring here, should we reset this in a `before_each`?  Not sure this will ever be an issue but might make it a little more stable if this failed to restore it.

src/granite.cr
```diff
@@ -2,6 +2,7 @@ require "yaml"
 require "db"
 
 module Granite
+  TIME_ZONE = "UTC"
```
drujensen: I recommend either using vscode crystal plugin or vim plugin to automatically format the code.  or run `crystal tool format .` to fix the tests.

drujensen:
@icyleaf thanks for the contribution! 🥇

src/granite.cr
icyleaf: Updated

spec/granite/fields/casting_spec.cr
icyleaf: I force push commit and add `before_each`

PR #311, opened 2019-01-06
Add colors to logged queries

Blacksmoke16:
Adds color and execution time to logged queries.

Totally steals the humanize duration from @robacarp's mosquito lib :wink: 

Can ignore everything up to the execution time, everything else is because im running in docker and mosquito.

SELECT:
image

INSERT:
image

With Numbers:
image

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -31,8 +31,8 @@ abstract class Granite::Adapter::Base
     yield database
   end
 
-  def log(query : String, elapsed_time : Benchmark::BM::Tms, params = [] of String) : Nil
-    Granite.settings.logger.info colorize query, params, elapsed_time
+  def log(query : String, elapsed_time : Time::Span, params = [] of String) : Nil
+    (logger = Granite.settings.logger) ? logger.info colorize query, params, elapsed_time.total_seconds : nil
   end
 
   # remove all rows from a table and reset the counter on the id.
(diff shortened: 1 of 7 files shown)


eliasjpr:
What about making strings `green` and numbers light blue? :sweat_smile:

Blacksmoke16:
At this moment keywords are light blue.  I could make strings in the WHERE portion be green, otherwise everything would be green due to the select columns and stuff are all strings.

eliasjpr:
I find it hard to read red over black bg. But this looks great, and always wanted timing for queries. I wish the community improves the SQL drivers to be faster, right now I find them slow.

Blacksmoke16:
Made strings warped in single quotes, be a light green.  Also changed the red to be slightly brighter.  

image

drujensen:
@Blacksmoke16 Does the benchmark slow the performance in any way?  Should we allow someone to disable it?

Blacksmoke16:
@drujensen Actually not as much as you would think, however colorize does.  I did change `Benchmark.measure` to `Time.measure` as it was a bit faster for what we needed.

Some benchmarks on selecting first row from a table.
```
colorize + benchmark  14.17k ( 70.57µs) (± 1.68%)  11041 B/op 
benchmark  16.07k ( 62.23µs) (± 1.74%)  4065 B/op
no benchmark/colorize  16.11k ( 62.07µs) (± 1.89%)  4065 B/op
colorize  14.13k ( 70.77µs) (± 1.78%)  10833 B/op
```
However i refactored the logger settings.  Now, if a logger is not defined by the user, the log method (and colorize) will not be executed.

Benchmark with colorize + benchmark after this change
```
No logger  16.73k ( 59.79µs) (± 1.76%)  3680 B/op  fastest
Logger.new nil  13.82k ( 72.36µs) (± 3.25%)  11041 B/op  fastest
```

PR #312, opened 2019-01-06
Support default values

Blacksmoke16:
Adds the ability to define default values.  Changes `#quote` to use `String.build` vs `#gsub` and `+`.  

```
         quote   9.23M (108.35ns) (± 3.16%)  256 B/op   1.60× slower
string builder   14.8M ( 67.58ns) (± 1.12%)  208 B/op        fastest
```

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -82,12 +82,15 @@ abstract class Granite::Adapter::Base
   macro inherited
     # quotes table and column names
     def quote(name : String) : String
-      char = QUOTING_CHAR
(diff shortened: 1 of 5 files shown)


src/adapter/base.cr
```diff
@@ -82,12 +82,15 @@ abstract class Granite::Adapter::Base
   macro inherited
     # quotes table and column names
     def quote(name : String) : String
-      char = QUOTING_CHAR
-      char + name.gsub(char, "#{char}#{char}") + char
+      String.build do |str|
```
drujensen: 💯 good idea!

PR #313, opened 2019-01-06
Support array IN queries with query builder

Blacksmoke16:
Allows for the query builder to handle params that are arrays.  It auto casts them to an `:in` query.

```
Courier.where(date_completed: nil, status: ["outstanding", "in_progress"]).select

SELECT contract_id, status, date_completed FROM couriers WHERE date_completed IS NULL AND status IN ('outstanding','in_progress') ORDER BY contract_id DESC
```

I had some trouble using placeholder vars in this context.  Both `["('outstanding','in_progress')"]` and `[["outstanding", "in_progress"]]` caused a syntax error.  To get around this i just build out the IN statement and plop in the value into the query.  Any ideas on how to better handle this?

### DIFF
--- a/spec/granite/query/assemblers/mysql_spec.cr
+++ b/spec/granite/query/assemblers/mysql_spec.cr
@@ -33,6 +33,26 @@ require "../spec_helper"
         assembler.where
         assembler.numbered_parameters.should eq [] of Granite::Fields::Type
       end
+
+      it "property defines IN query with numbers" do
+        sql = "SELECT #{query_fields} FROM table WHERE date_completed IS NULL AND id IN (1,2) ORDER BY id DESC"
+        query = builder.where(date_completed: nil, id: [1, 2])
+        query.raw_sql.should match ignore_whitespace sql
+
+        assembler = query.assembler
+        assembler.where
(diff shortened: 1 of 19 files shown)


spec/granite/query/assemblers/mysql_spec.cr
```diff
@@ -23,6 +23,16 @@ require "../spec_helper"
         assembler.where
         assembler.numbered_parameters.should eq ["bob", "23"]
       end
+
+      it "property defines IN query" do
```
eliasjpr: In the case of Ints, Floats and bools would it have single quotes? I ask because I don't see tests for those cases.

Blacksmoke16: Ah crap.  No they wouldn't.  I'll get that fixed up.

Blacksmoke16: @eliasjpr Fixed.  This did expose an issue with `Float32` types.  I'll look into it later and make a separate PR for that.  Not sure if its a Granite issue or DB/adapter shard issue yet.

PR #315, opened 2019-01-08
Logger optimization

Blacksmoke16:
@Rx14 brought to my attention a better way to implement the feature where logging doesnt run if the logger's io is nil.

https://github.com/amberframework/granite/compare/master...Blacksmoke16:logger-optimization?expand=1#diff-094e647002b88e121bcba1392f8416a8R34

Since we are calling the `colorize` method inside the block of `Granite.settings.logger.debug`, it will return before executing `colorize` if the IO is set to nil.  Because of this, I was able to reset the logger's type to `Logger`, and revert the nil checks i was doing.  Also because of this https://github.com/amberframework/amber/pull/1027 can be reverted once this is released.

### DIFF
--- a/spec/granite_spec.cr
+++ b/spec/granite_spec.cr
@@ -1,4 +1,5 @@
 require "./spec_helper"
+require "logger"
 
 describe Granite::Base do
   it "can instaniate a model with default values" do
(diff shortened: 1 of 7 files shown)


spec/granite_spec.cr
```diff
@@ -8,6 +9,30 @@ describe Granite::Base do
...
+    end
+
+    describe "when logger is set to nil" do
+      it "should not be logged" do
+        Granite.settings.logger = Logger.new nil
+
+        a = 0
+        Granite.settings.logger.info { a = 1 }
+        a.should eq 0
+      end
```
robacarp: this test is essentially testing the functionality of Logger, I'm not sure it's needed.

Blacksmoke16: Probably not, but I'd rather keep it.  Completes the idea of `if you have a logger defined it logs, otherwise it doesn't`.

src/granite/settings.cr
```diff
@@ -2,7 +2,7 @@ require "logger"
 
 module Granite
   class Settings
-    property logger : Logger? = nil
+    property logger : Logger = Logger.new nil
```
drujensen: 👍

PR #317, opened 2019-01-13
Make migrator.create behave similarly to migrator.drop

coderobe:
Currently, migrator.drop will `DROP IF EXISTS` to avoid erroring when there's nothing to be dropped - similarly, create shouldn't error when there's nothing to create

### DIFF
--- a/src/granite/migrator.cr
+++ b/src/granite/migrator.cr
@@ -53,7 +53,7 @@ module Granite::Migrator
         }
 
         stmt = String.build do |s|
-          s.puts "CREATE TABLE #{ @quoted_table_name }("
+          s.puts "CREATE TABLE IF NOT EXISTS #{ @quoted_table_name }("
 
           # primary key
           k = {{adapter}}.quote("{{primary_name}}")


robacarp:
Hey @coderobe, thanks for your pull. The trick is, the table may exist but also be the wrong table. With the current code, it'll fail and produce an error when the table exists. By adding `IF NOT EXISTS` the action will still succeed, but the state is indeterminate. What do you think?

coderobe:
Hmm, i think it's up to the developer to either migrate or drop outdated tables; Note that the behavior of drop_and_create doesn't change either. With this patch, you can make sure the database is initialized before using it without special dances. I don't think opening a db with a different schema and then trying to create existing tables is an action this should bother to support.

robacarp:
looks like a legit test failure:

```text
  1) Granite::Migrator::Base #create_sql generates correct SQL with sqlite adapter
     Failure/Error: {% if env("CURRENT_ADAPTER") == "pg" %}
       Expected: "CREATE TABLE \"reviews\"(\n\"id\" INTEGER NOT NULL PRIMARY KEY,\n\"name\" VARCHAR(255)\n,\n\"downvotes\" INTEGER\n,\n\"upvotes\" INTEGER\n,\n\"sentiment\" FLOAT\n,\n\"interest\" REAL\n,\n\"published\" BOOL\n,\n\"created_at\" VARCHAR\n) ;\n"
            got: "CREATE TABLE IF NOT EXISTS \"reviews\"(\n\"id\" INTEGER NOT NULL PRIMARY KEY,\n\"name\" VARCHAR(255)\n,\n\"downvotes\" INTEGER\n,\n\"upvotes\" INTEGER\n,\n\"sentiment\" FLOAT\n,\n\"interest\" REAL\n,\n\"published\" BOOL\n,\n\"created_at\" VARCHAR\n) ;\n"
```

coderobe:
oh well, that test needs to also be updated i suppose

drujensen:
@coderobe Closing this.  Please reopen if you decide to fix the tests.

PR #322, opened 2019-02-22
Make Granite::Base abstract

Blacksmoke16:
Add volume to docker-compose
Bumps crystal version in dockerfile (to get specs to pass)

Since `Granite::Base` will never be instantiated on its own it should be abstract.  This will prevent https://github.com/crystal-lang/crystal/issues/7473 from happening when having multiple possible models as a return type.

Also adds a volume to the `docker-compose` file so that you wont have to rebuild the `spec` container after making changes for those changes to be reflected when running `spec`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM crystallang/crystal:0.27.0
+FROM crystallang/crystal:0.27.2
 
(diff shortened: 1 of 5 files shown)


Blacksmoke16:
Apparently the volume breaks CI.  :shrug:

PR #323, opened 2019-02-22
Pass annotations from a `belongs_to` to the resulting field

Blacksmoke16:


robacarp:
Can you provide some context? I think this is a good idea but I'm not sure what's going on

Blacksmoke16:
Yea sorry @robacarp. Since Granite uses mutable constants to define its properties at a later time (instead of the `field` macro actually creating the property), applying annotations on top of the `field` def doesn't apply the annotations to that field's property.  As a work around in #289 i made it so the `field` macro supports a `annotations` property that takes an array of quoted annotations that will be applied to the property that gets created later on.  This allows for the user to apply any arbitrary annotations they want to the underlying property that gets created.  Such as to support third party shards that use annotations to function (my serialization shard for example).

Since the `belongs_to` association macro uses the `field` macro to create that FK field, there would be no way to pass annotations that should be applied to that FK field.  This PR just allows the user to specify the annotations that should be created in the `field` macro.  For example (using my serialization/validation shard):

Register a custom validator to determine a user with the given id exists:
```crystal
register_assertion Assert::UserExists, [:id]

class UserExistsAssertion(ActualValueType) < CrSerializer::Assertions::Assertion
  @message : String = "Invalid {{field}}.  A 'User' with the provided ID does not exist."

  def initialize(
    field : String, # => Name of the instance variable
    message : String?, # => Message to display if the assertion fails
    @actual : ActualValueType, # => value of the instance variable
    @id : Int64?
  )
    super field, message # => Calls the parent's initializer to set field name and message
  end

  def valid? : Bool
    !!Blog::Models::User.find @id
  end
end
```

Use the new validator on the FK field
```crystal
    belongs_to :user, annotations: [
      "@[Assert::NotNil]",
      "@[Assert::UserExists(id: id)]"
    ]
```

So that when you go to POST a new article for example, it would 400 if you didn't specify the `user_id` or if the id given doesn't map to an actual user.
```JSON
{
  "code": 400,
  "message": "Validation tests failed",
  "errors": [
    "'user_id' should not be null",
    "Invalid user_id.  A 'User' with the provided ID does not exist."
  ]
}
```

robacarp:
Thanks for the background. Makes sense.

PR #324, opened 2019-02-23
Do not update created_at on update

Blacksmoke16:
This PR makes it so that when updating a record (that has a valid ID and `new_record = false`), with a `nil` `created_at`, it should not update the `created_at` field in the DB.

I noticed this (and the PRs before this) while working on my blog/tutorial post.  Where I was working on updating an object from the JSON PUT body like:
```json
{
  "id": 12,
  "user_id": 7,
  "title": "My Blog",
  "body": "The Blog!",
  "updated_at": "2019-02-23T15:32:03Z"
}
```
However, since `created_at` was not included, it was getting set to `nil`, which upon saving the record, was overriding the `created_at` column to be null.  Which, even if it _was_ set, it should not be included in the `UPDATE` query.

I also changed the `new_record` getter to a property.  The use case for this is so that external shards could add extensions to instantiate an object from something, and tell granite that it should be treated as an existing model.  Whereas before it was only directly set when fetching directly from the database or saving it for the first time.

### DIFF
--- a/src/granite/transactions.cr
+++ b/src/granite/transactions.cr
@@ -125,8 +125,14 @@ module Granite::Transactions
 
     private def __update
       set_timestamps mode: :update
-      fields = self.class.content_fields
+      fields = self.class.content_fields.dup
       params = content_values + [@{{primary_name}}]
+
+      # Do not update created_at on update
+      if created_at_index = fields.index("created_at")
+        fields.delete_at created_at_index
+        params.delete_at created_at_index
+      end
 
       begin
         @@adapter.update @@table_name, @@primary_name, fields, params
@@ -239,7 +245,7 @@ module Granite::Transactions
   # Returns true if this object hasn't been saved yet.
   @[JSON::Field(ignore: true)]
   @[YAML::Field(ignore: true)]
-  getter? new_record : Bool = true
+  property? new_record : Bool = true
 
   # Returns true if this object has been destroyed.
   @[JSON::Field(ignore: true)]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM crystallang/crystal:0.27.0
+FROM crystallang/crystal:0.27.2
 
 ARG sqlite_version=3110000
 ARG sqlite_version_year=2016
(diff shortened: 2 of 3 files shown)


src/granite/transactions.cr
```diff
@@ -125,9 +125,15 @@ module Granite::Transactions
...
     private def __update
       set_timestamps mode: :update
-      fields = self.class.content_fields
+      fields = self.class.content_fields.dup
       params = content_values + [@{{primary_name}}]
 
+      # Do not update created_at on update
+      if created_at_index = fields.index("created_at")
+        fields.delete_at created_at_index
+        params.delete_at created_at_index
```
robacarp: There is opportunity for this code to crash when fields and params fall out of sync. The resulting bug will be harder to find because a seemingly random index in params will have been deleted.

Could this conditional be replaced with two `.delete("created_at")` calls instead?

src/granite/transactions.cr
```diff
@@ -239,7 +245,7 @@ module Granite::Transactions
   # Returns true if this object hasn't been saved yet.
   @[JSON::Field(ignore: true)]
   @[YAML::Field(ignore: true)]
-  getter? new_record : Bool = true
+  property? new_record : Bool = true
```
robacarp: 👍

src/granite/transactions.cr
Blacksmoke16: I'm open to ideas.  But currently its a bit tricky since `fields` is just an array of column names, and `params` is just an array of values.  It would be much easier if it was like a hash of `column_name => value`.  

That might not be a bad idea, but would deff increase the scope of the PR.

src/granite/transactions.cr
Blacksmoke16: Mmm, I think the new test I added in `update_spec` covers it since it's using the setter, which is the only new thing changing it would do.  I'm not sure what else you could test, especially for just a bool property.

src/granite/transactions.cr
robacarp: Oh geeze, I didn't realize that. It's not worth unwinding here.

robacarp:
Thanks @Blacksmoke16

PR #325, opened 2019-02-23
Make primary_type return the class type vs string

Blacksmoke16:
Makes the `primary_type` method return the actual `.class` type of the primary key, versus a string.  Having the actual type is much more useful.

### DIFF
--- a/spec/granite/table/table_spec.cr
+++ b/spec/granite/table/table_spec.cr
@@ -1,7 +1,7 @@
 require "../../spec_helper"
 
 describe Granite::Table do
-  describe "#table_name" do
+  describe ".table_name" do
     it "sets the table name to name specified" do
       CustomSongThread.table_name.should eq "custom_table_name"
     end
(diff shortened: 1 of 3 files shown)


spec/granite/table/table_spec.cr
```diff
@@ -1,7 +1,7 @@
 require "../../spec_helper"
 
 describe Granite::Table do
-  describe "#table_name" do
+  describe ".table_name" do
```
robacarp: This `#` syntax is common in many languages, and I think is preferred. It's the syntax the crystal docs use.

src/granite/table.cr
```diff
@@ -44,7 +44,7 @@ module Granite::Table
     @@table_name = "{{table_name}}"
     @@primary_name = "{{primary_name}}"
     @@primary_auto = "{{primary_auto}}"
-    @@primary_type = "{{primary_type}}"
+    @@primary_type = {{primary_type}}
```
robacarp: I'm okay with this, and I agree it is a better paradigm. It is also a breaking change to the API.

spec/granite/table/table_spec.cr
Blacksmoke16: I was going off of https://crystal-lang.org/reference/guides/testing.html#anatomy-of-a-spec-file:

```
For unit tests, it is recommended to follow the conventions for method names: 
Outer describe is the name of the class, inner describe targets methods. 
Instance methods are prefixed with `#`, class methods with `.`.
```

src/granite/table.cr
Blacksmoke16: We could add a new method to get the actual type and leave the string one untouched?  If we wanted to avoid the breaking change.  However, I'm skeptical if it would be worth it?

spec/granite/table/table_spec.cr
robacarp: Bah, I'm moving too fast today. Sorry. Class methods should start with a `.` yes.

src/granite/table.cr
robacarp: I have no problem incrementing the version, just want to me sure it's marked as a pull which is a breaking change.

Issue #225, opened 2018-06-08
Timezones aren't persisted to the database (or rendered in hashes, json, etc)


robacarp:
```crystal
write = Model.new
write.user_id = 3_i64
write.timestamp_field = Time.now
write.save

puts "field before database: #{write.timestamp_field}"
puts "hash field before database: #{write.to_h["timestamp_field"]}"

puts "="*40

read = Model.find!(write.id)
puts "field after database: #{read.timestamp_field}"
puts "hash field after database: #{read.to_h["timestamp_field"]}"

puts "="*40

delta = Time.now - read.timestamp_field.not_nil!
puts "Time delta: #{delta}"

# field before database: 2018-06-08 17:07:34 -06:00
# hash field before database: 2018-06-08 17:07:34
# ========================================
# field after database: 2018-06-08 17:07:34 UTC
# hash field after database: 2018-06-08 17:07:34
# ========================================
# Time delta: 06:00:00.673049000
```

At a minimum timezones should be persisted to the database, but they should also be emitted into json in the right string format for javascript.

I'm not sure that Time objects should be stringified when a model is converted to a hash at all.

robacarp:
Adding a little note here about something to consider when implementing this: converting to UTC will solve a _bunch_ of the use cases here, but not all of them. http://www.creativedeletion.com/2015/03/19/persisting_future_datetimes.html

Blacksmoke16:
@robacarp I was looking into this a bit more.  From what i can tell (at least related to PG), is that PG doesnt support timezone specific datetime with timezone columns.  For example using the `TIMESTAMP` data type:
```Crystal
t = Test.new
t.datetime = Time.now(Time::Location.load("America/New_York"))
pp t
t.save

pp Test.first!
```
Which prints (with extraneous fields removed):
```
#<Test:0x280df00
@datetime=2018-10-21 14:28:54.133870000 -04:00 America/New_York,
@id=nil>
#<Test:0x280de60
@datetime=2018-10-21 14:28:54.133000000 UTC,
@id=21_i64>
```
PG will take the given date and assume and store it as is, but when returning it, no TZ data is given so crystal assumes its UTC.

There is the `TIMESTAMPTZ` data type that will take the time obj with tz, convert it to UTC, and store the UTC version of it in the database.

[...]

PR #326, opened 2019-02-25
Extract Migrator#create and #drop sql into separate methods

### LINKED ISSUE #319: Add `Granite::Migrator#create_sql` and `#drop_sql`
### tl;dr

`Granite::Migrator` generates SQL on `#drop` and `#create`, and runs it imminently. The possibility of having the migrator export the generated SQL would be nice for use in migration tools like `micrate`. The generating functionality is already in place, implementing this would just be moving the generating code into `create`- and `drop`- `_sql` methods (names debatable), and using them in place of the current generating code.

---

I'm using Granite for building a program (i.e. no amber), I'm also quite new to SQL, which is one of the reasons I enjoy Granite so much. 

Following the Granite [docs on migrating][docs], I stumble upon SQL in the micrate up and down sections, I do not know how Granite interfaces with its tables. Therefore I do not know how to define these migrations, and I take a look at the alternative `Granite::Migrator#create` and `#drop`, which both work great, however they seem wholly impractical to use long term, at least compared to `micrate`. 

I hoped I could extract the SQL generated by `Granite::Migrator`, but seems that's not possible, hence I ask here whether it's a good idea or not to extract the generated code into separate methods. Practical or not it adheres better to the single responsibility principle... Sorry for the spaghetti text.

I'll happily implement this if it's accepted as a good idea.

[docs]: https://github.com/amberframework/granite/blob/master/docs/migrations.md#creating-a-migration

Jens0512:
Fixes #319. 

I haven't managed to run the specs (difficulties with docker), but the changes shouldn't change any behaviour, only expose functionality.

### DIFF
--- a/spec/granite/migrator/migrator_spec.cr
+++ b/spec/granite/migrator/migrator_spec.cr
@@ -17,15 +17,15 @@ describe Granite::Migrator::Base do
         Review.migrator.create_sql.should eq <<-SQL
           CREATE TABLE "reviews"(
           "id" BIGSERIAL PRIMARY KEY,
-          "name" VARCHAR(255)
+          "name" TEXT
           ,
           "downvotes" INT
           ,
           "upvotes" BIGINT
           ,
-          "sentiment" FLOAT
+          "sentiment" REAL
           ,
-          "interest" REAL
+          "interest" DOUBLE PRECISION
           ,
           "published" BOOL
           ,
@@ -35,8 +35,8 @@ describe Granite::Migrator::Base do
 
         Kvs.migrator.create_sql.should eq <<-SQL
           CREATE TABLE "kvs"(
-          "k" VARCHAR(255) PRIMARY KEY,
-          "v" VARCHAR(255)
+          "k" TEXT PRIMARY KEY,
+          "v" TEXT
           ) ;\n
           SQL
 
@@ -77,7 +77,7 @@ describe Granite::Migrator::Base do
           ,
           `sentiment` FLOAT
           ,
-          `interest` REAL
+          `interest` DOUBLE
           ,
           `published` BOOL
           ,
--- a/spec/granite/migrator/migrator_spec.cr
+++ b/spec/granite/migrator/migrator_spec.cr
@@ -1,0 +1,98 @@
+require "../../spec_helper"
+
+describe Granite::Migrator::Base do
+  describe "#drop_sql" do
+    it "generates correct SQL with #{{{ env("CURRENT_ADAPTER") }}} adapter" do
+      {% if env("CURRENT_ADAPTER") == "mysql" %}
+        Review.migrator.drop_sql.should eq "DROP TABLE IF EXISTS `reviews`;"
+      {% else %}
+        Review.migrator.drop_sql.should eq "DROP TABLE IF EXISTS \"reviews\";"
+      {% end %}
+    end
+  end
+
+  describe "#create_sql" do
+    it "generates correct SQL with #{{{ env("CURRENT_ADAPTER") }}} adapter" do
+      {% if env("CURRENT_ADAPTER") == "pg" %}
+        Review.migrator.create_sql.should eq <<-SQL
+          CREATE TABLE "reviews"(
+          "id" BIGSERIAL PRIMARY KEY,
+          "name" VARCHAR(255)
+          ,
+          "downvotes" INT
+          ,
+          "upvotes" BIGINT
+          ,
+          "sentiment" FLOAT
+          ,
+          "interest" REAL
+          ,
+          "published" BOOL
+          ,
+          "created_at" TIMESTAMP
+          ) ;\n
+          SQL
+
+        # Also check Array types for pg
+        ArrayModel.migrator.create_sql.should eq <<-SQL
+          CREATE TABLE "array_model"(
+          "id" SERIAL PRIMARY KEY,
+          "str_array" TEXT[]
+          ,
+          "i16_array" SMALLINT[]
+          ,
+          "i32_array" INT[]
+          ,
+          "i64_array" BIGINT[]
+          ,
+          "f32_array" REAL[]
+          ,
+          "f64_array" DOUBLE PRECISION[]
+          ,
+          "bool_array" BOOLEAN[]
+          ) ;\n
+          SQL
+
+      {% elsif env("CURRENT_ADAPTER") == "mysql" %}
+        Review.migrator.create_sql.should eq <<-SQL
+          CREATE TABLE `reviews`(
+          `id` BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+          `name` VARCHAR(255)
+          ,
+          `downvotes` INT
+          ,
+          `upvotes` BIGINT
+          ,
+          `sentiment` FLOAT
+          ,
+          `interest` REAL
+          ,
+          `published` BOOL
+          ,
+          `created_at` TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+          ) ;\n
+          SQL
+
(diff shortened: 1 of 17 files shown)


Jens0512:
But alas, as per Murphy's Law, "If something can go wrong, it will." I did. Fix coming right up

Jens0512:
There, obvious mistake on my end :sweat_smile:. Fixed now

robacarp:
@Jens0512 This is a nice added feature. What do you think about adding a test to show that these methods are working as expected?

Jens0512:
Well, while that's a great idea, I'm not quite great enough at SQL to know what the should return...

Blacksmoke16:
@Jens0512 Its pretty easy.  It would however depend on what the current adapter is.  Assuming for a model like:

```crystal
class Post < Granite::Base
  adapter pg
  table_name posts

  field name : String
  field body : String
  timestamps
end
```
* MySQL
  * Drop - ``DROP TABLE IF EXISTS `posts`;``
  * Create - ``CREATE TABLE `posts`( `id` BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY, `name` VARCHAR(255), `body` VARCHAR(255), `created_at` TIMESTAMP DEFAULT CURRENT_TIMESTAMP, `updated_at` TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP)`` 
* PG
  * Drop - `DROP TABLE IF EXISTS "posts";`
  * Create - ``CREATE TABLE "posts"( "id" BIGSERIAL, "name" VARCHAR(255), "body" VARCHAR(255), "created_at" TIMESTAMP, "updated_at" TIMESTAMP)`` 
* SqLite
  * Drop - `DROP TABLE IF EXISTS "posts";`
  * Create - ``CREATE TABLE "posts"( "id" INTEGER NOT NULL, "name" VARCHAR(255), "body" VARCHAR(255), "created_at" VARCHAR, "updated_at" VARCHAR)``

robacarp:
@Blacksmoke16 is right about the technique "run the method, copy the output, and assert that the result is always the output".

Typically, I wouldn't say that's a good practice, but in this case we already know it's working (if only by experience). So asserting that the output is consistent will do the job of ensuring that no accidental changes to the output of the methods are introduced without raising a warning.

Jens0512:
> "run the method, copy the output, and assert that the result is always the output"

Sounds good to me, as long as we know the initial output is correct.

Thanks a ton for the help @Blacksmoke16. 

> Ideally we would use a model that has all data types in it to fully test the migrator.

Shall I create one and put it in `spec/spec_models.cr` then?

Blacksmoke16:
@Jens0512 Yea that would be fine.  Be sure to also test the PG array types.  There is already a model that has them defined.  Will have to scope it to run on PG only since the other DBs dont support it.  Can reference https://github.com/amberframework/granite/blob/master/spec/granite_spec.cr#L261-L321

Jens0512:
(Seems there was a model with all types already: [Review](https://github.com/amberframework/granite/blob/master/spec/spec_models.cr#L133))

Jens0512:
The formatting of the generated SQL is terrible to behold haha

Jens0512:
Oh no, ameba has detected a terrible line of whitespace...

Jens0512:
Maybe mention somewhere in the README that you should run ameba before you push. Just in case you put a 2-3 extra white spaces anywhere... :D

Blacksmoke16:
@Jens0512 This is looking great thanks!  I remembered there would be two other cases we should test.  

1. Natural PK, can use https://github.com/amberframework/granite/blob/master/spec/spec_models.cr#L190
2. UUID PK, can use https://github.com/amberframework/granite/blob/master/spec/spec_models.cr#L351

Would you mind adding specs for those as well?

Blacksmoke16:
Sorry Satisfactory sidetracked me this weekend.  I'll get this reviewed tonight.

@robacarp @eliasjpr If you guys have any input as well, it would be nice.

Jens0512:
Nothing to be sorry about :)

spec/granite/migrator/migrator_spec.cr
```diff
@@ -0,0 +1,135 @@
...
+          "id" BIGSERIAL PRIMARY KEY,
+          "name" VARCHAR(255)
+          ,
+          "downvotes" INT
+          ,
+          "upvotes" BIGINT
+          ,
+          "sentiment" FLOAT
+          ,
+          "interest" REAL
```
Blacksmoke16: Im thinking the type of `interest` should be `DOUBLE PRECISION`.  Based on the [docs](https://www.postgresql.org/docs/9.5/datatype.html#DATATYPE-TABLE) that would be their 8 byte floating point type.

spec/granite/migrator/migrator_spec.cr
```diff
@@ -0,0 +1,135 @@
...
+        Review.migrator.create_sql.should eq <<-SQL
+          CREATE TABLE "reviews"(
+          "id" BIGSERIAL PRIMARY KEY,
+          "name" VARCHAR(255)
+          ,
+          "downvotes" INT
+          ,
+          "upvotes" BIGINT
+          ,
+          "sentiment" FLOAT
```
Blacksmoke16: Im thinking the type of `sentiment` should be `REAL`.  Based on the [docs](https://www.postgresql.org/docs/9.5/datatype.html#DATATYPE-TABLE) that would be their 4 byte floating point type.

spec/granite/migrator/migrator_spec.cr
```diff
@@ -0,0 +1,135 @@
...
+          ,
+          "interest" REAL
+          ,
+          "published" BOOL
+          ,
+          "created_at" TIMESTAMP
```
Blacksmoke16: We _could_ use `TIMESTAMPZ` which would do conversion of the datetime object and store it in UTC.  However that would probably be a breaking change as people's datetime's would start returning differently then.  Can save that for another discussion, related to #225 .

spec/granite/migrator/migrator_spec.cr
```diff
@@ -0,0 +1,135 @@
...
+    end
+  end
+
+  describe "#create_sql" do
+    it "generates correct SQL with #{{{ env("CURRENT_ADAPTER") }}} adapter" do
+      {% if env("CURRENT_ADAPTER") == "pg" %}
+        Review.migrator.create_sql.should eq <<-SQL
+          CREATE TABLE "reviews"(
+          "id" BIGSERIAL PRIMARY KEY,
+          "name" VARCHAR(255)
```
Blacksmoke16: Should probably make `name` be `TEXT`, seems to be the preferred variable length string type for PG.

spec/granite/migrator/migrator_spec.cr
```diff
@@ -0,0 +1,135 @@
...
+          `id` BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+          `name` VARCHAR(255)
+          ,
+          `downvotes` INT
+          ,
+          `upvotes` BIGINT
+          ,
+          `sentiment` FLOAT
+          ,
+          `interest` REAL
```
Blacksmoke16: Can make this `DOUBLE` based on [this chart](http://www.peachpit.com/articles/article.aspx?p=1752305&seqNum=2)

spec/granite/migrator/migrator_spec.cr
Jens0512: So I'll leave this untouched?

Blacksmoke16: Yea.

Jens0512:
Thanks for the feedback, anything else?

Issue #329, opened 2019-04-21
Crystal 0.28.0 support


lebogan:

lebogan:
Using the --warnings flag to build one of my cli-based apps, I get a warning about using `Time.local` or `Time.utc` instead of `Time.now` in `lib/granite/src/granite/transactions.cr`. 

As this has been noted as a breaking change (depreciation) by the Crystal team, I updated my local copy and recompiled. 

Is this something to be overly concerned about?

Blacksmoke16:
@lebogan These are just warnings letting you know they are deprecated and will be removed in a _future_ version.  It does not affect Granite's ability to work on `0.28.0`; just that we will want to replace the usages for the next crystal version.

Blacksmoke16:
@lebogan However I'll submit a PR to take care of this and update CI to use `0.28.0`.

PR #330, opened 2019-04-22
Some updates for Crystal 0.28.0

Blacksmoke16:
* Updates Dockerfile to `0.28.0` for CI
* Updates deprecated references of `Time.utc_now` and `Time.now` (#329)
* Updates docker-compose file to show all deprecation warnings
  * Also will error on something being deprecated.  I could go either way if we want CI to fail or not if something becomes deprecated.  Thoughts @drujensen @robacarp @amberframework/granite?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM crystallang/crystal:0.27.2
+FROM crystallang/crystal:0.28.0
 
 ARG sqlite_version=3110000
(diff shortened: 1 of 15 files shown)


robacarp:
@Blacksmoke16 failing CI on deprecated makes sense... it can be rolled back later if needed, but it's good to know as early as possible about that sort of thing

PR #341, opened 2019-06-10
Disallow union types

### LINKED ISSUE #340: Error undefined macro method 'Union#resolve'
Hi the last PR introduced an error when type is a union

```crystal
expanding macro
in lib/granite/src/granite/fields.cr:168: undefined macro method 'Union#resolve'

            {% elsif type.resolve <= Array %}
                          ^~~~~~~

in src/models/game.cr:10: undefined macro method 'Union#resolve'

  field sent_code_at : Time?
```

Blacksmoke16:
Resolves #340 

```cr
...
in spec/spec_helper.cr:16: while requiring "./spec_models"

require "./spec_models"
^

in spec/spec_models.cr:8: The type of Teacher#id cannot be a Union.  The 'primary' macro declares the type as nilable by default.

{% begin %}
^
```
```cr
...
in spec/spec_helper.cr:16: while requiring "./spec_models"

require "./spec_models"
^

in spec/spec_models.cr:8: The type of Teacher#age cannot be a Union.  The 'field' macro declares they type as nilable by default.  Use the 'field!' macro to declare a not nilable field.

{% begin %}
```

### DIFF
--- a/src/granite/fields.cr
+++ b/src/granite/fields.cr
@@ -14,6 +14,7 @@ module Granite::Fields
 
   # specify the fields you want to define and types
   macro field(decl, **options)
+    {% raise "The type of #{@type.name}##{decl.var} cannot be a Union.  The 'field' macro declares the type as nilable by default.  Use the 'field!' macro to declare a not nilable field." if decl.type.is_a? Union %}
     {% CONTENT_FIELDS[decl.var] = options || {} of Nil => Nil %}
     {% CONTENT_FIELDS[decl.var][:type] = decl.type %}
   end
--- a/src/granite/table.cr
+++ b/src/granite/table.cr
@@ -28,6 +28,7 @@ module Granite::Table
   end
 
   macro primary(decl, **options)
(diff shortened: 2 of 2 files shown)


src/granite/fields.cr
```diff
@@ -14,6 +14,7 @@ module Granite::Fields
 
   # specify the fields you want to define and types
   macro field(decl, **options)
+    {% raise "The type of #{@type.name}##{decl.var} cannot be a Union.  The 'field' macro declares the type as nilable by default.  Use the 'field!' macro to declare a not nilable field." if decl.type.is_a? Union %}
```
robacarp: I love these error messages!

They're long enough that the suffix `if` kind of gets lost, would it be better to break this out into multiple lines for maintainability?

Blacksmoke16: :man_shrugging: I don't it really matters.

drujensen: @robacarp are you ok with this?  we can refactor this later if you feel it's hard to read.

drujensen: @amberframework/contributors Any other feedback?  If not, I will merge.

PR #342, opened 2019-06-16
Support custom field types

### LINKED ISSUE #152: Support Adapter-specific types
Using DB-specific type

```crystal
class Batch < Granite::ORM::Base
  adapter mysql
  field start_time : Time::Span
```

causes

```
no overload matches 'Hash(String, Bool | Float32 | Float64 | Int32 | Int64 | Slice(UInt8) | String | Time | Nil)#[]=' with types String, (Time::Span | Nil)
Overloads are:
 - Hash(K, V)#[]=(key : K, value : V)
Couldn't find overloads for these types:
 - Hash(String, Bool | Float32 | Float64 | Int32 | Int64 | Slice(UInt8) | String | Time | Nil)#[]=(key : String, value : Time::Span)
```

#### status
- ✅ MySQL supports `time` type
- ✅ `mysql` shard supports `time` type as `Time::Span`
- ✅ `crystal-db` shard can handle any custom types.  (see [example](https://github.com/crystal-lang/crystal-db/blob/master/spec/custom_drivers_types_spec.cr#L97-L98))
- ❌ Granite supports only `DB::Any`

#### codes

Currently, we use `DB::Any` statically in `module Foo` like this.

```crystal
    def to_h
      fields = {} of String => DB::Any
```

I think that the correct solution is to rewrite this part by macro.
```diff
-      fields = {} of String => DB::Any
+      fields = {} of String => {{adapter}}::Any
```

I tried this a few hours, but I don't know how to pass `adapter` macro value into all other files.
Or, is it better to define modules as `Granite::ORM::Fields(T)` where `T` is one of adapters?

Thus, it was not easy to be solved. So I created an issue first. 😄

### LINKED ISSUE #193: Support MySQL JSON data type
Granite, as far as I can tell, does not support MySQL JSON data type.  

 ```Crystal
    adapter mysql
    table_name items

    primary type_id : Int64, auto: false
    field locations : JSON::Type | JSON::Any
  end
```

```SQL
CREATE TABLE items
(
  type_id   BIGINT NOT NULL PRIMARY KEY,
  locations JSON                    NOT NULL
)
INSERT INTO items (type_id, locations) VALUES (4, '[60003760, 1022887354289, 1023736622218, 1022568876022, 1025064405048, 1026712174905, 1026960906556]');

```
```Crystal
  p Item.first
```

```
Missing hash key: 245_u8 (KeyError)
  from /usr/share/crystal/src/hash.cr:0:9 in 'fetch'
  from /usr/share/crystal/src/hash.cr:62:5 in '[]'
  from lib/mysql/src/mysql.cr:9:7 in 'column_type'
  from lib/mysql/src/mysql/result_set.cr:81:13 in 'read'
  from lib/db/src/db/result_set.cr:79:15 in 'read'
  from src/EveToolsApi/models/item.cr:4:3 in 'set_attributes'
  from src/EveToolsApi/models/item.cr:4:3 in 'from_sql'
  from lib/granite_orm/src/granite_orm/querying.cr:49:17 in 'raw_all'
  from lib/kemal/src/kemal/helpers/helpers.cr:0:3 in '~procProc(Array(EveToolsApi::Models::Item))'
  from lib/granite_orm/src/granite_orm/collection.cr:255:3 in 'collection'
  from lib/granite_orm/src/granite_orm/collection.cr:0:5 in 'Granite::ORM::Collection(EveToolsApi::Models::Item)'
  from lib/granite_orm/src/granite_orm/querying.cr:69:5 in 'first'
  from lib/granite_orm/src/granite_orm/querying.cr:68:3 in 'first'
  from src/EveToolsApi.cr:42:3 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:11:3 in '_crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:112:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:101:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:135:3 in 'main'
  from __libc_start_main
  from _start
  from ???
```
It will work if i set the column type as text and field type as a string, parse the string as json, but ideally granite should return JSON::Any already without that extra step.

Blacksmoke16:
Initial pass at supporting custom DB types. Resolves #193 and resolves #152.

* Allows fields to have converters to convert from their actual type, to a type that could be used by the DB.
  * Serializing a class to json on save, but deserializing it from json on read.
* Adds converters for `UUID`, `Enum`, `JSON`, and `PG::Numeric`.
  * Refactors the current auto UUID implementation to use the new converter
* This _DOESNT_ address the issue that #152 mentions in that the union of supported types is limited and fixed. But it would help alleviate the more pressing limitations with the current implementation.  I.e. would allow for easily handling the conversion of `PG::Numeric` into `Float64`, or a `String` into an actual `UUID` object when saving/reading from the DB.
* Allows the user to specify the actual data type of eac field.  Only useful ATM for the migrator for specs.

TODO:
* ~Update docs~


spec/spec_models.cr
```diff
@@ -352,7 +352,14 @@ require "uuid"
...
 
-    primary uuid : String, auto: :uuid
+    primary uuid : UUID, converter: Granite::Converters::UuidConverter
+  end
+
+  class UUIDNaturalModel < Granite::Base
+    adapter {{ adapter_literal }}
+    table_name uuids
+
+    primary uuid : UUID, converter: Granite::Converters::UuidConverter, auto: false
```
drujensen: Can you define a `field uuid : UUID` as well?

Blacksmoke16: Yea, should be able to.  I can add a spec for `field` and `field!` as well.  Would of course have to set the converter on them tho.

src/granite/converters.cr
```diff
@@ -0,0 +1,11 @@
+module Granite::Converters
```
drujensen: What do you think about adding the `to_db` and `from_rs` to the custom class by re-opening it? That way you don't need to specify a converter?

Blacksmoke16: In this case you mean reopen `UUID` to add them?

drujensen: Hhhm.  that would work but not sure I like re-opening it.  Another option would be an abstract base class that we would inherit from.

For example:
```
class User < Granite::Base
...
  primary uuid : MyUUID
  field encrypted_password : EncryptedPassword
end

class MyUUID < Granite::CustomType
  property uuid : UUID

  def to_db
    uuid.to_s
  end

  def from_rs(value)
    uuid = UUID.new(value)
  end
end

class EncryptedPassword < Granite::CustomType
  property password : String

  def to_db
    //encrypt before storing
  end

  def from_rs
    //decrypt after retrieval???
  end
end
```

Just throwing out some ideas here.  Let me know what you think.

Blacksmoke16: Hmm, I dunno.  That seems like a lot more work than just adding a `, converter: Granite::Converters::UuidConverter` to the field.

drujensen:
Ok. I’m fine with it the way it is.

Blacksmoke16:
Currently planning on implementing the following converters:
* UUID - Both binary and string formats.
* Enum - Both integer and string formats.
* JSON - Both binary and string formats
* PG::Numeric - Allows using types like `DECIMAL` in PG.

Any others you think that would be helpful?

drujensen:
@Blacksmoke16 Sounds like a good plan.  We can always add more converters later.

Blacksmoke16:
@drujensen What are your thoughts on how we should handle the different return values from each DB driver? I.e. PG returns `JSON::Any` for its JSON(B) types, but SQLite just returns a string, which messes up the read.

I'm thinking we can either:
1. Have global converters to handle each case, i.e. `BinaryJson`, `StringJson`, `Json`, for `Bytes`, `String`, and `JSON::Any`.
2. Namespace the converters by driver, which gets included when including that adapter. i.e. `Granite::Converters::PG::Json`.
3. Similar to 2, but only support driver specific types, i.e. for PG, only support JSON(B) types, and not `String` or `BYTEA` types; for SQLite support `String`/`Blob`.

Option 1 would have the most flexibility, but also the most converters.
Option 2 would mostly be 3 near identical sets based on option 1.
Option 3 would have the least amount of duplication, but also the least flexible.

drujensen:
I’m leaning toward option 1.  It seems the most flexible.

drujensen:
Is there a way to get rid of the large case statement to perform type conversions using a set of converters?

Blacksmoke16:
Probably yea.  But prob best for another PR.

Blacksmoke16:
@drujensen One thing I thought of that would help cut down on the amount of duplication, would be to have the user provide the type that should be read from the column.

`Granite::Converters::Json(String)` or `Granite::Converters::Json(Bytes)`

Then you would have a singular `Json` converter module that just defines the to/from methods based on the type of that generic.

```cr
module Json(T)
  {% if T == String %}
    # Add from_rs/to_db methods that go to/from a String
  {% elsif T == JSON::Any %}
    # Add from_rs/to_db methods that go to/from JSON::Any
  {% elsif T == Bytes %}
    # Add from_rs/to_db methods that go to/from a Bytes
  {% else %}
    {% raise "#{{{@type.name}}} does not currently support #{T}" %}
  {% end %}
end
```

Is still going to be some duplication, but this would at least limit the total number of converters and keep the code centralized.

drujensen:
@Blacksmoke16 Yes, good idea.  Another option is having just one converter that supports a Union of `(String | Bytes | JSON::Any)`.  then add logic to check the type before converting.

Blacksmoke16:
Hmm, I'll have to try that how to see how it works with `rs.read`.

Blacksmoke16:
@drujensen I tested it out, it worked pretty good for the reading from the db, however you wouldn't really have a way to know what to do with the value on save.  I.e. for an enum, should I call `.to_i` or `.to_s`?

spec/spec_models.cr
```diff
@@ -400,6 +401,92 @@ require "uuid"
...
+          END IF;
+      END$$;
+      TYPE
+    )
+  {% elsif env("CURRENT_ADAPTER") == "sqlite" %}
+    class ConverterModel < Granite::Base
+      adapter {{ adapter_literal }}
+      table_name converters
+
+      field binary_json : MyType, column_type: "BLOB", converter: Granite::Converters::Json(MyType, Bytes)
```
drujensen: This seems very verbose with a lot of types being defined:  MyType, BLOB, Json, MyType, Bytes.  Anyway to reduce complexity here?

spec/granite/converters/converters_spec.cr
```diff
@@ -0,0 +1,351 @@
+require "../../spec_helper"
+
+describe Granite::Converters do
```
drujensen: Lots of tests!  💯

spec/spec_models.cr
Blacksmoke16: Not without manually defining a converter for every case. Like `PgJsonStringConverter`, `JsonStringConverter`, etc.

In reality the end user would most likely only have to do like:
`field binary_json : MyType, converter: Granite::Converters::Json(MyType, Bytes)` as the `column_type` is currently only used within Granite's migrator for specs.

spec/granite/converters/converters_spec.cr
Blacksmoke16: More to come :wink:

Blacksmoke16:
@drujensen @robacarp This should be good for final review.

drujensen:
@amberframework/contributors Any other feedback?  If not, I will merge.

PR #343, opened 2019-06-22
Adds an `exists?` class method

Blacksmoke16:
Adds a method that will return true if a record exists that matches the provided criteria, or false otherwise.  Based off of [ActiveRecord.exists?](https://apidock.com/rails/ActiveRecord/Base/exists%3F/class).

* If given a `String` or `Number`, attempts to lookup a records with a PK with that value.
* If given a `Hash` or `NamedTuple`, attempts to lookup a record that matches the given criteria.  
* Refactors some of the querying logic to support `find_by(Hash)`, and to share some code between `exists?` and `find_by`
* Fixes a bug with `find_by` that made it not correctly match nil values.
* Fixes the type restriction on various querying methods to allow `Array` and `UUID` values.


drujensen:
@amberframework/contributors Any other feedback?  If not, I will merge.

robacarp:
Hm, I think I'd expect the syntax to be `Model.where(col: val).exists?`. 

The type based overloading makes me a little nervous, but I think it's well thought out.

Blacksmoke16:
@robacarp That's also an option.  What would we do if they did like `Model.where(col: val).order(:created_at).limit(50)`  Just ignore the `order` and `limit`?

robacarp:
@Blacksmoke16 man it'd be great if it _weren't possible_ by compile-time-enforcement to call #exists? on a query which had #limit called on it. However, I think that's a bigger refactor than is needed here, and that just raising a runtime warning of some sort is fine.

To be clear, I don't think having a Class#exists? method is mutually exclusive with a Class#where()#exists?, I think we can provide both.

Blacksmoke16:
> To be clear, I don't think having a Class#exists? method is mutually exclusive with a Class#where()#exists?, I think we can provide both.

Right, as long as we're OK with just ignoring other parts of the query I can implement this.  Im imaging it to just be a new executor like `select` but does the exists query instead.

Blacksmoke16:
@robacarp I added that if you wanted to review again.

src/granite/query/assemblers/base.cr
```diff
@@ -170,6 +170,17 @@ module Granite::Query::Assembler
...
+    def exists? : Executor::Value(Model, Bool)
+      sql = build_sql do |s|
+        s << "SELECT EXISTS(SELECT 1 "
+        s << "FROM #{table_name} "
+        s << where
+        s << ")"
+      end
+
+      Executor::Value(Model, Bool).new sql, numbered_parameters, default: false
+    end
```
robacarp: nice

src/granite/query/builder.cr
```diff
@@ -178,6 +178,10 @@ class Granite::Query::Builder(Model)
     assembler.select.run
   end
 
+  def exists? : Bool
+    assembler.exists?.run
+  end
+
```
robacarp: 👍

src/adapter/pg.cr
```diff
@@ -141,7 +141,7 @@ class Granite::Adapter::Pg < Granite::Adapter::Base
     log statement, elapsed_time, value
   end
 
-  def ensure_clause_template(clause : String) : String
+  protected def ensure_clause_template(clause : String) : String
```
robacarp: 👍 seems like a good visibility change. I feel like the name of this method is totally opaque, but I can't think of a better name right now.

src/granite/query/executors/value.cr
```diff
@@ -10,15 +10,11 @@ module Granite::Query::Executor
       # db.scalar raises when a query returns 0 results, so I'm using query_one?
       # https://github.com/crystal-lang/crystal-db/blob/7d30e9f50e478cb6404d16d2ce91e639b6f9c476/src/db/statement.cr#L18
 
-      raise "No default provided" unless @default
+      raise "No default provided" if @default.nil?
```
robacarp: Is this just a personal preference change?

src/granite/querying.cr
```diff
@@ -109,11 +136,23 @@ module Granite::Querying
...
+    @@adapter.exists? quoted_table_name, clause, params
+  end
+
+  private def build_find_by_clause(criteria : Hash(Symbol | String, Granite::Fields::Type))
+    criteria = criteria.dup
+    {
+      criteria.keys.map { |name| "#{quoted_table_name}.#{quote(name.to_s)} #{(criteria[name].nil? ? (criteria.delete(name); "IS NULL") : "= ?")}" }.join(" AND "),
+      criteria.values,
+    }
+  end
```
robacarp: I'm a short line, quick to read kind of guy. What do you think of writing this method as:

```crystal
private def build_find_by_clause(criteria : Hash(Symbol | String, Granite::Fields::Type))
  names = criteria.keys
  values = criteria.dup

  clauses = names.map do |name|
    if values[name]
      matcher = "= ?"
    else
      matcher = "IS NULL"
      values.delete name
    end

    "#{quoted_table_name}.#{quote name.to_s} #{matcher}"
  end

  { clauses.join(" AND "), values }
end
```

src/granite/query/executors/value.cr
Blacksmoke16: No, it's required since for the `.exists?` assembler, I'm passing `false` as the default, which was causing it to always raise.

Blacksmoke16:
@robacarp that's updated.  I also set the MySQL shard to the latest commit so that CI passes, as https://github.com/crystal-lang/crystal-mysql/pull/78 got merged.

robacarp:
@drujensen thoughts on targeting a non-release version of mysql? A commit hash is just as unique as a version number but since it's pinned to the exact commit could it cause version conflicts with projects which require mysql themselves?

@Blacksmoke16 any idea when that'll be released to a numbered version?

Blacksmoke16:
For us it's only a development dependency.  But yea, someone else using the mysql shard would have to use master or this commit.

I do not, that would be a question for @bcardiff.

drujensen:
@robacarp looks like 0.7.0 will be released soon https://github.com/crystal-lang/crystal-mysql/pull/79

Blacksmoke16:
Should be good to go now @drujensen @robacarp

robacarp:
thanks as always @Blacksmoke16

PR #345, opened 2019-07-06
replace @@var with their getter

Blacksmoke16:
I discovered this when working on another project.  The jist of it is when restricting an ivar to `Granite::Base`, the compiler will complain that

```cr
can't infer the type of class variable '@@primary_name' of Granite::Base.class
    exec_exists "#{@@primary_name} = ?", [id]
```

I'm guessing that since we're operating on the class level and these class vars get set via the macro included/finished hooks, they are not set.  However using the class getters gets around that since their types are known and not reliant on the actual setting of the value.

### DIFF
--- a/spec/granite_spec.cr
+++ b/spec/granite_spec.cr
@@ -1,7 +1,25 @@
 require "./spec_helper"
 require "logger"
 
+class SomeClass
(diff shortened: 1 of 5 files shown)


drujensen:
:100:

Issue #344, opened 2019-07-02
Feature Request: Support slave/master database connections


Xosmond:
Hello i have a problem, I have two different database url's, one for reading/writing and one for only reading. I would like to have the ability to change the adapter on an instance object so I can write over it. Like:

```crystal
user = User.find(id)
if user.code == 'M'
  user.switch_adapter(:write)
  user.update(money: 2000)
end
```

Blacksmoke16:
I think this would be a good feature to have but would like to think though the design a bit first.

Thoughts @drujensen @robacarp?

Xosmond:
Closing due to no recent activity.

PR #346, opened 2019-07-07
Port Granite to use Annotations

Blacksmoke16:
I took a stab at migrating Granite to use Annotations.  This PR is the initial working version of the port.  I tried to keep breaking changes to a minimum, basically just porting existing functionality, vs refactoring a bit as well. 

There is still some refactoring to do, which will clean things up more I'd like to do.  But wanted to start a discussion before getting too far into it as this would be a good time to change things.

This PR:
* Refactors the `field` macro to define a property and add an annotation on it
* Refactors the `table_name` macro to add an annotation to the model's class
* Removes all mutable constants and `__process_*` macros

TODO:
* ~Look into merging the `field` and `field!` macro into a `column` macro that uses nilibility of the type to define whether it should raise on nil or not.~
* ~Refactor the migrator to support more options.~
  * Probably save this for another PR.

Benefits:
* Annotations can now be added directly above the field declaration
* Removed the big case statement in `cast_to_field`
* Removes a bunch of complexity (IMO)

Drawbacks:
* I don't have a good way to define a default primary key if none is specified, now having a model w/o a primary key raises a compile time error
* Some of the extra conversions that Granite did in `cast_to_field`, such as `String` to `Number` isn't currently implemented

Questions:
* ~Is it worth trying to find a way to auto define the PK?~
  * No
* Is it worth the extra logic to convert strings etc
  * Is a bit cleaner now with the new sys, but I don't think it should be Granite's responsibility
* Any other thoughts/ideas/questions others come up with.


drujensen:
I'm torn here.  I am not a fan of annotations.  One of the greatest advantages of Ruby (and now Crystal) is readability.   Granted, macro's are worse than annotations for readability but the macro's can be hidden in the framework where annotations are seen by the end user.  

Should we simplify the framework code and push some of the ugliness to the user?  Or should we keep ugly macro's and provide a clean DSL to the user?  @eliasjpr @elorest @robacarp any thoughts on this?

Is there a way to keep both syntaxes?

drujensen:
@Blacksmoke16 Is the annotations only for JSON Serialization and optional?

drujensen:
@Blacksmoke16 So it looks like the macro's for `primary`, `field`, and `table` are adding the annotation to the field and then the annotation takes care of mapping.  This is the best of both worlds.  We can have our cake and eat it too.  💯 This is awesome!

Blacksmoke16:
@drujensen Right yea.  This is mostly a port of the old backend to be powered off annotations vs the `__proces_*` macros.  The user facing DSL hasn't changed.

I'm mainly trying to figure out how much of a breaking change do we want this to be.  Have an opportunity now to change some functionality. I.e. maybe merge `field`, `field!`, and maybe `primary` into a singular `column` macro.  Where you do like `column id : Int64, primary: true`.  Or do we want to just merge `field` and `field!` and keep `primary`.

Or if there are any other pain points, now would be the time to discuss/implement them.

drujensen:
I am all for simplification.  Do you prefer having `column` instead of `field`?  I can live with everything using a single macro with a flag instead of a separate `primary` macro.  One of the things we need to consider is supporting composite primary keys.

Blacksmoke16:
Would have to think about it but should be doable.  For now the plan is to raise a compile time error if you define more than one, and have that as a future iteration.

I think it makes more sense, considering it's basically defining a database column.  But if others like `field` more I don't really care either way.

spec/adapter/adapters_spec.cr
```diff
@@ -2,6 +2,8 @@ require "../spec_helper"
 
 class Foo < Granite::Base
   adapter mysql
+
+  primary id : Int64
```
drujensen: I am good with requiring all fields to be defined including primary.

spec/granite/querying/from_sql_spec.cr
```diff
@@ -18,12 +18,12 @@ macro build_review_emitter
 end
 
 def method_which_takes_any_model(model : Granite::Base.class)
-  model.as(Granite::Base).from_sql build_review_emitter
+  model.as(Granite::Base).from_rs build_review_emitter
```
drujensen: yup, better name.

spec/spec_models.cr
```diff
@@ -133,6 +133,9 @@ require "uuid"
   class Review < Granite::Base
     adapter {{ adapter_literal }}
     table_name reviews
+
+    primary id : Int64
+
```
drujensen: ✂️

spec/spec_models.cr
```diff
@@ -214,7 +219,10 @@ require "uuid"
     table_name books
     has_many :book_reviews, class_name: BookReview
     belongs_to author : Person
-    belongs_to publisher : Company, foreign_key: publisher_id : Int32, json_options: {ignore: true}, yaml_options: {ignore: true}
+
+    @[JSON::Field(ignore: true)]
+    @[YAML::Field(ignore: true)]
+    belongs_to publisher : Company, foreign_key: publisher_id : Int32
```
drujensen: 👍 I can live with this instead of `json_options`

spec/spec_models.cr
```diff
@@ -303,6 +313,8 @@ require "uuid"
     adapter {{ adapter_literal }}
     table_name todos
 
+    primary id : Int64
+
```
drujensen: ✂️

spec/spec_models.cr
```diff
@@ -312,6 +324,8 @@ require "uuid"
     adapter {{ adapter_literal }}
     table_name after_json_init
 
+    primary id : Int64
+
```
drujensen: ✂️

spec/spec_models.cr
```diff
@@ -360,42 +375,64 @@ require "uuid"
...
     adapter {{ adapter_literal }}
     table_name todos_json
 
-    field name : String, json_options: {key: "task_name"}
-    field priority : Int32, json_options: {ignore: true}
-    field updated_at : Time, json_options: {ignore: true}
-    field created_at : Time, json_options: {key: "posted"}
+    primary id : Int64
+
+    @[JSON::Field(key: "task_name")]
```
drujensen: yeah, I wouldn't want to see this in my codebase. 👎   I would prefer seeing a `src/serializers/TodoJsonOptionsSerializer.cr` that handles this.  The good news is this that this is optional.

spec/spec_models.cr
```diff
@@ -360,42 +375,64 @@ require "uuid"
...
 
   class DefaultValues < Granite::Base
     adapter {{ adapter_literal }}
     table_name defaults
 
-    field name : String, default: "Jim"
-    field is_alive : Bool, default: true
-    field age : Float64, default: 0.0
+    primary id : Int64
+
```
drujensen: ✂️ and all the rest for consistency

src/adapter/mysql.cr
```diff
@@ -54,7 +54,7 @@ class Granite::Adapter::Mysql < Granite::Adapter::Base
     last_id
   end
 
-  def import(table_name : String, primary_name : String, auto : String, fields, model_array, **options)
+  def import(table_name : String, primary_name : String, auto : Bool, fields, model_array, **options)
```
drujensen: nice catch!

src/granite/base.cr
```diff
@@ -36,21 +41,34 @@ abstract class Granite::Base
   macro inherited
     include JSON::Serializable
     include YAML::Serializable
-    macro finished
```
drujensen: nice!

src/granite/fields.cr
```diff
@@ -1,27 +1,64 @@
...
   end
 
   # specify the fields you want to define and types
   macro field(decl, **options)
     {% raise "The type of #{@type.name}##{decl.var} cannot be a Union.  The 'field' macro declares the type as nilable by default.  Use the 'field!' macro to declare a not nilable field." if decl.type.is_a? Union %}
-    {% CONTENT_FIELDS[decl.var] = options || {} of Nil => Nil %}
-    {% CONTENT_FIELDS[decl.var][:type] = decl.type %}
+    {% column_type = (options[:column_type] && !options[:column_type].nil?) ? options[:column_type] : nil %}
+    {% converter = (options[:converter] && !options[:converter].nil?) ? options[:converter] : nil %}
+    @[Granite::Column(column_type: {{column_type}}, converter: {{converter}})]
```
drujensen: The macro and annotation name should match.  `field` -> `@[Granite::Field()]` or `column` -> `@[Granite::Column()]`

src/granite/fields.cr
```diff
@@ -1,27 +1,64 @@
...
+    def {{decl.var.id}}! : {{decl.type}}
+      raise NilAssertionError.new {{@type.name.stringify}} + "#" + {{decl.var.stringify}} + " cannot be nil" if @{{decl.var}}.nil?
+      @{{decl.var}}.not_nil!
+    end
   end
 
   # specify the raise-on-nil fields you want to define and types
   macro field!(decl, **options)
-    field {{decl}}, {{options.double_splat(", ")}}raise_on_nil: true
+    {% raise "The type of #{@type.name}##{decl.var} cannot be a Union.  The 'field' macro declares the type as nilable by default.  Use the 'field!' macro to declare a not nilable field." if decl.type.is_a? Union %}
```
drujensen: Something is wrong with this error message.  They are already using the `field!` macro and we are recommending them to use it again?

src/granite/fields.cr
```diff
@@ -30,156 +67,53 @@ module Granite::Fields
     field updated_at : Time
   end
 
-  macro __process_fields
```
drujensen: 🥇 💯 🎉

src/granite/table.cr
```diff
@@ -9,11 +9,37 @@ macro disable_granite_docs?(stmt)
...
+    def table_name : String
+      {% begin %}
+        {% table_ann = @type.annotation(Granite::Model) %}
+        {{table_ann && !table_ann[:table].nil? ? table_ann[:table] : @type.name.underscore.stringify}}
+      {% end %}
+    end
+  end
+
+  macro table_name(name)
+    @[Granite::Model(table: {{(name.is_a?(StringLiteral) ? name : name.stringify) || nil}})]
```
drujensen: hmm.  Seems like the macro and annotation should match here as well but not sure what name works for both.

Blacksmoke16: Model makes it flexible enough if we wanted to add other fields to it in the future.

src/granite/fields.cr
Blacksmoke16: Probably a copy paste error yea.

src/granite/fields.cr
Blacksmoke16: Kinda made this with the assumption that I would change it to be `column` in the future.  PR is still a WIP so thats on the TODO list.

robacarp:
> The user facing DSL hasn't changed.

👌 I like the direction this is headed in as well @Blacksmoke16, thanks for your work here

src/granite/table.cr
Blacksmoke16: But thats a good question of what would happen if we had another macro that did something else.  

What about making the macro like `model table: "people"`, then it could support extra stuff if anything else makes sense.

drujensen: I like `table` and `column` as the macros and annotation names.  Maybe consider `database` instead of adapter.

Blacksmoke16: I could live with that, we'll just have to add more annotations if we ever think of something else we would use.

Blacksmoke16:
Just a little update:

* Merged `field`, `field!`, and `primary` into a singular `column` macro
  * Whether the column should raise or not is based on if the type is nilable or not.
    * Primary columns are nilable and auto by default
* Renamed the `table_name` macro to `table`
  * This required me to also rename the module to `Tables` as to not conflict with the `Table` annotation.

```cr
class SongThread < Granite::Base
  adapter {{ env("CURRENT_ADAPTER").id }}
  table "threads" # `table threads` also works

  column id : Int64, primary: true
  column name : String?
  column not_nil_int : Int32
  column default : Bool = false
end
```

I also managed to add some compile time safety into the setters.

Current behavior:
```cr
class SongThread < Granite::Base
  table_name "threads" # `table threads` also works

  primary id : Int64
  field name : String
  field! not_nil_int : Int32
  field! default : Bool = false
end

thread = SongThread.new

pp thread
#<SongThread:0x7f6253609cc0
 @_current_callback=nil,
 @default=nil,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @name=nil,
 @new_record=true,
 @not_nil_int=nil>

thread.default = nil

pp thread
#<SongThread:0x7f6253609cc0
 @_current_callback=nil,
 @default=nil,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @name=nil,
 @new_record=true,
 @not_nil_int=nil>
```

This PR:
```cr
class SongThread < Granite::Base
  table "threads" # `table threads` also works

  column id : Int64, primary: true
  column name : String?
  column not_nil_int : Int32
  column default : Bool = false
end

thread = SongThread.new
#<SongThread:0x7f5293590cc0
 @_current_callback=nil,
 @default=false,
 @destroyed=false,
 @errors=[],
 @id=nil,
 @name=nil,
 @new_record=true,
 @not_nil_int=nil>

thread.default = nil
#  35 | thread.default = nil
#              ^------
# Error: no overload matches 'SongThread#default=' with type Nil

# Overloads are:
#  - SongThread#default=(default : Bool)
```

The thing to note here is that this is only applied on the setters.  By default Granite makes the fields nilable.  This allows for a future PR to specify certain fields to select and only populate those values.  But it now prevents you from settings a non nilable field to nil.

This doesn't remove the runtime `Model#column cannot be nil errors`.  But I'm thinking there isn't much you can do about that?

elorest:
Nice work. I like that the user facing DSL will remain the same.

drujensen:
@Blacksmoke16 I love it.  Couple questions. 
- Should we rename `adapter` to `database`?  
- Are we going to change anything with the `timestamps` macro that generates the `created_at`, `updated_at` columns?

Blacksmoke16:
Hmm, I was actually thinking of `connection`.  Since you are basically giving the macro a named connection string.  I.e. could have one pointing at a PG db and another at MySQL db.

Also was going to make it so if you only have the one adapter registered, it uses that by default, or the first if you have more than one.  This would allow you to register your primary connection first, then have it be the default for all your models.

Would also be open to just renaming the macro here, then do another pass of the adapter logic in #344.

In regards to the `timestamp` macro, only thing i changed was making them nilable.  Did you have something else in mind?

drujensen:
yeah, `connection` probably is a better name.  I was going with the `database` -> `table` -> `column` naming because it seemed clean.

I had nothing in mind with `timestamps`.  I was just thinking if we are making a breaking change, we should do it once and it deals with mapping fields.  Do you plan to have annotations for timestamps?

Blacksmoke16:
>  Do you plan to have annotations for timestamps?

Not sure I follow.  I already updated the `timestamps` macro to use `column` with type of `Time?`.  Isn't really anything else to do.

drujensen:
I guess you can just define columns as `updated_at` and `created_at` and it will automatically update them, right?

Blacksmoke16:
Yea, thats still handled in https://github.com/amberframework/granite/blob/8a36750b2923529384fd0be057ca614045c4a128/src/granite/transactions.cr#L80-L90

spec/spec_models.cr
Blacksmoke16: It's the easiest way.  One line annotation vs a whole new class seems a bit much. 

Is just the nature of `JSON::Serializable`.

Blacksmoke16:
Latest update:

* Renamed `adapter` macro to `connection`
* Renamed `Granite::Adapters` to `Granite::Connections`
* Removes the need for the hash when newing up an adapter
  * `Granite::Connections << Granite::Adapter::Mysql.new(name: "mysql", url: ENV["MYSQL_DATABASE_URL"])`

Think i'll start updating docs now.  Less there is anything else we would like to change before this is merged?

Blacksmoke16:
Ok.  Going to request another review and give an overview of what changed to make upgrading easier before marking it ready.

@drujensen @robacarp @eliasjpr @amberframework/contributors @amberframework/core-team @amberframework/granite 

# Upgrading

This PR brings a good amount of breaking changes. 

## Connections

* Renamed `Granite::Adapters` to `Granite::Connections`
* Made the adapter initializer arguments no longer require a hash
* The `adapter` macro has been renamed to `connection`
  * Accepts a string or macroId

### Registering a Connection

```crystal
Granite::Connections << Granite::Adapter::Pg.new(name: "pg", url: ENV["PG_DATABASE_URL"])
# or
Granite::Connections << Granite::Adapter::Pg.new("pg", ENV["PG_DATABASE_URL"])
```

### Using a Connection

```crystal
class User < Granite::Base
  connection pg
  # or
  connection "pg"

  # Columns
end
```

## Defining columns
* `field`, `field!`, and `primary` macros have been consolidated into a singular `column` macro
* The primary key must now be manually defined

### Nilable Columns

Denoted by a type that is nilable.

```crystal
column deleted_at : Time?
```

### Not Nilable Columns

Denoted by a type that is not nilable.

```crystal
column name : String
```

### Primary Key

Denoted by defining a column with the `primary: true` option.  By default primary keys are AUTO INCREMENT.

> **NOTE:** The type of the primary key can be set to either nilable or not nilable, but will always be made nilable behind the scenes.  I.e. since a model wouldn't have an ID until it is saved.

```crystal
column id : Int64, primary: true
```

### Natural Primary Keys

Denoted by defining a column with the `primary: true` and `auto: false` options.

```crystal
column custom_id : Int32, primary: true, auto: false
```

## Setting Values

Some additional compile time safety has been added, that would prevent setting a non-nilable column to `nil`.

```crystal
class Model < Granite::Base
  column id : Int64, primary: true
  column default : Bool = false
end

model = Model.new
thread.default = nil
#  35 | model.default = nil
#              ^------
# Error: no overload matches 'Model#default=' with type Nil

# Overloads are:
#  - Model#default=(default : Bool)
```

Before this would have just set the value to `nil` even if it was defined with the `field!` macro.

## Defining Table Name

* The `table_name` has been renamed to `table`.
  * Accepts a string or macroId

```crystal
class User < Granite::Base
  # some connection
  table users
  # or
  table "users"

  # Columns
end
```

## Generating Documentation

Generating documentation that includes Granite methods/constants/properties has changed from using the `DISABLE_GRANITE_DOCS` ENV variable to a `granite_docs` compile time flag.

```bash
crystal docs -D granite_docs
```

# TODO

* Refactor the `fields`, `content_fields` methods to return an array of like `ColumnInfo` structs that would contain all information related to a specific column.  Something like:

```crystal
record ColumnInfo(T) < ColumnBase,
  property_name : String,
  column_name : String,
  nilable : Bool,
  auto : Bool = false,
  primary : Bool = false,
  default : T? = nil,
  type : T.class = T
```

Then methods could be added to it, such as to generate the SQL for that specific column.

* TBD

docs/models.md
```diff
@@ -2,36 +2,37 @@
...
 class Foo < Granite::Base
-  adapter legacy_db
+  connection legacy_db
   
   # model fields
 end
 
 class Bar < Granite::Base
-  adapter new_db
+  connection "new_db"
```
drujensen: Using quotes here might be confusing or considered a typo since this section isn't related to the type provided to the macro.

docs/models.md
```diff
@@ -41,50 +42,51 @@ Would be equivalent to:
...
-## Custom Primary Key
+## Primary Keys
 
-For legacy database mappings, you may already have a table and the primary key is not named `id` or `Int64`.
+Each model is required to have a primary key defined.  Use the `column` macro with the `primary: true` option to denote the primary key. 
 
-Use the `primary` macro to define your own primary key
+> **NOTE:** Composite primary keys are not yet supported.
 
 ```crystal
```
drujensen: Add an example here of the standard primary key of `column id : Int64, primary: true` and then show how to define a custom id below that.

docs/models.md
```diff
@@ -95,45 +97,51 @@ book.isbn # => RFC4122 V4 UUID string
...
+  @[Bar::Settings(other_option: 7)]
+  column password : String
 
-  field name : String
-  field password : String, annotations: ["@[Foo::Options(option1: true)]", "@[Bar::Settings(other_option: 7)]"]
-  field age : Int32
+  column name : String
+  column age : Int32
 end
 ```
```
drujensen: Do you want to document the new annotations that are available in Granite?

docs/models.md
```diff
@@ -2,36 +2,37 @@
 
 ## Multiple Adapters
```
drujensen: Multiple Connections?

docs/models.md
Blacksmoke16: I think I would rather them be an implementation detail for the moment.

docs/models.md
Blacksmoke16: What benefit would that bring?  Would be the same just with a different name.

drujensen: You can change this to `id` if you don't think we need two examples but it's a well known naming convention (at least in the rails world) that `id` is used for the primary key and `{model_name}_id` is used for a foreign key.  Right now in the models.md, there is no example of the recommended naming convention.

spec/granite/validations/validator_spec.cr
```diff
@@ -3,27 +3,33 @@ require "../../spec_helper"
 {% begin %}
   {% adapter_literal = env("CURRENT_ADAPTER").id %}
   class NameTest < Granite::Base
-    adapter {{ adapter_literal }}
-    field name : String
+    connection {{ adapter_literal }}
+
+    column id : Int64, primary: true
```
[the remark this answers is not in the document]
Blacksmoke16: It wouldn't compile.  Would error saying like `Composite primary keys are not supported yet`.

src/granite/connections.cr
```diff
@@ -0,0 +1,14 @@
+module Granite
+  class Connections
+    class_getter registered_connections = [] of Granite::Adapter::Base
+
+    def self.<<(adapter : Granite::Adapter::Base) : Nil
+      raise "Adapter with name '#{adapter.name}' has already been registered." if @@registered_connections.any? { |conn| conn.name == adapter.name }
+      @@registered_connections << adapter
+    end
+
+    def self.get_connection(name : String) : Granite::Adapter::Base?
```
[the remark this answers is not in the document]
Blacksmoke16: 😮 👍

spec/granite/validations/validator_spec.cr
```diff
@@ -3,27 +3,33 @@ require "../../spec_helper"
...
+    column name : String?
 
     validate :name, "cannot be blank", ->(s : NameTest) do
```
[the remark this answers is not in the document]
Blacksmoke16: https://github.com/amberframework/granite/pull/346#discussion_r302119593 Are my personal thoughts on the matter.

src/granite/base.cr
```diff
@@ -36,21 +41,34 @@ abstract class Granite::Base
...
+    getter? destroyed : Bool = false
+
+    # Returns true if the record is persisted.
+    disable_granite_docs? def persisted?
+      !(new_record? || destroyed?)
     end
 
-    def initialize(**args : Granite::Fields::Type)
-      set_attributes(args.to_h)
+    def initialize(**args : Granite::Columns::Type)
```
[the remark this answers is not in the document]
Blacksmoke16: Yes, its just an alias type that allows `DB::Any` plus `Array(T)` and `UUID`.

https://github.com/amberframework/granite/blob/0af97216c925ac18513d7f813c409f54331b971a/src/granite/columns.cr#L5-L6

docs/models.md
```diff
@@ -158,10 +155,14 @@ enum OrderStatus
...
 
   # Other fields
-  field status : OrderStatus, converter: Granite::Converters::Enum(OrderStatus, String) 
+  column status : OrderStatus, converter: Granite::Converters::Enum(OrderStatus, String) 
 end
 ```
+
+## Serialization
+
+Granite implements [JSON::Serializable](https://crystal-lang.org/api/JSON/Serializable.html) and [YAML::Serializable](https://crystal-lang.org/api/YAML/Serializable.html) by default.  As such, models can be serialized to/from JSON/YAML via the `#to_json`/`#to_yaml` and `.from_json`/`.from_yaml` methods.
```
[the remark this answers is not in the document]
Blacksmoke16: I could go either way.  On one hand it's nice to have there ready to go, but on the other if you're not going to use it it's basically just dead code.

@drujensen @robacarp WDYT?

src/granite/columns.cr
```diff
@@ -0,0 +1,135 @@
...
+
+    {% column_type = (options[:column_type] && !options[:column_type].nil?) ? options[:column_type] : nil %}
+    {% converter = (options[:converter] && !options[:converter].nil?) ? options[:converter] : nil %}
+    {% primary = (options[:primary] && !options[:primary].nil?) ? options[:primary] : false %}
+    {% auto = (options[:auto] && !options[:auto].nil?) ? options[:auto] : false %}
+    {% auto = (!options || (options && options[:auto] == nil)) && primary %}
+
+    {% nilable = (type.is_a?(Path) ? type.resolve.nilable? : (type.is_a?(Union) ? type.types.any?(&.resolve.nilable?) : (type.is_a?(Generic) ? type.resolve.nilable? : type.nilable?))) %}
+
+    @[Granite::Column(column_type: {{column_type}}, converter: {{converter}}, auto: {{auto}}, primary: {{primary}}, nilable: {{nilable}})]
```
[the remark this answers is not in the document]
Blacksmoke16: @eliasjpr For this initial version I tried to keep the user facing DSL as close to what it is currently.  I'm treating the annotations as an implementation detail for the moment.  If that's something we want to expose more to the end use it is doable since thats pretty much what the `column` macro does.

Just know the macro also does some other stuff for you.  Such as defining the ivars as nilable but defines the getters to be no nilable, and making the setter not be nilable if the column is not nilable.

src/granite/columns.cr
```diff
@@ -0,0 +1,135 @@
...
+      def {{decl.var.id}} : {{type.id}}
+        raise NilAssertionError.new {{@type.name.stringify}} + "#" + {{decl.var.stringify}} + " cannot be nil" if @{{decl.var}}.nil?
+        @{{decl.var}}.not_nil!
+      end
+    {% end %}
+  end
+
+  # include created_at and updated_at that will automatically be updated
+  macro timestamps
+    column created_at : Time?
```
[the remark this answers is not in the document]
Blacksmoke16: When you just new up a new model and it hasn't been saved yet.

src/granite/columns.cr
```diff
@@ -0,0 +1,135 @@
+require "json"
+require "uuid"
+
+module Granite::Columns
+  alias SupportedArrayTypes = Array(String) | Array(Int16) | Array(Int32) | Array(Int64) | Array(Float32) | Array(Float64) | Array(Bool)
+  alias Type = DB::Any | SupportedArrayTypes | UUID
+  TIME_FORMAT_REGEX = /\d{4,}-\d{2,}-\d{2,}\s\d{2,}:\d{2,}:\d{2,}/
```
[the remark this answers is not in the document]
Blacksmoke16: It might not be anymore?  I'll look.

src/granite/columns.cr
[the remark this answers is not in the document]
eliasjpr: Thanks. I think the team will agree on using `annotations as an implementation detail`.

spec/spec_models.cr
eliasjpr: I vote for us to use annotations in the background while providing the user an unchangecahable dsl, I think is a good idea to move the validation and serialization into annotations as long as we do that  as a `implementation detail`, obscuring the user from it.

src/granite/columns.cr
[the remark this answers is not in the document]
eliasjpr: We need composite indexes definition 😅

spec/spec_models.cr
Blacksmoke16: This is easier said than done IMO.  I don't think you could get a similar syntax.

E.x. https://play.crystal-lang.org/#/r/78k2

Annotations have to be applied to the property when it's defined.

docs/models.md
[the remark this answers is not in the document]
drujensen: I personally don't use it.  I prefer using serializers.  However, I know this was requested several times so I'm ok leaving it since it seems like others prefer this approach.

spec/spec_models.cr
drujensen: I'm curious, how would you create a custom annotation if a validation depends on two fields like `password` and `password_confirmation`?

Blacksmoke16: Off the top of my head, pass the getter/ivar name as an annotation field.

```
class PasswordAssertion(ModelPropertyType)
  include CrSerializer::Assertions::Assertion

  @message : String = "'{{actual}}' is not a valid '{{field}}'"

  def initialize(
    field : String,
    message : String?,
    @actual : ModelPropertyType,
    @confirmation : String
  )
    super field, message
  end

  def valid? : Bool
	# Do something 
  end
end


@[PasswordAssertion(confirmation: password_confirmation)]
column password : String
```

Something like that maybe?

drujensen: Ok.  I can see that working.  I was trying to come up with a validation that is not `field/column` related or you would want it associated with multiple fields.

Blacksmoke16:
Planning on finishing this up this week.  Will address pending feedback and mark it ready for review.

Blacksmoke16:
Should be good if you guys want to do one last review.

@drujensen @robacarp @eliasjpr @elorest.

I also updated https://github.com/amberframework/granite/pull/346#issuecomment-511130830 to represent some of the newest updates.

PR #348, opened 2019-07-14
Update CRUD delete example code (#347)

### LINKED ISSUE #347: Error in docs about deletion
Not too important but it caused me some headaches because I thought it was not working.

[I'm talking about this exampe](https://github.com/amberframework/granite/blob/master/docs/crud.md#L109)
with this code
```crystal
post = Post.find 1
post.destroy
puts "deleted" unless post
```

It seems the `unless` parts no longer works as intended, since I never got any output.

Also related #145, could be accounted in the docs 👍 

It could be better maybe a bit like this:

```crystal
post = Post.find 1
post.destroy if post
puts "deleted" if post.destroyed?
```

EnricoMonese:
Resolves #347

### DIFF
--- a/docs/crud.md
+++ b/docs/crud.md
@@ -106,7 +106,7 @@ Delete a specific record.
 ```crystal
 post = Post.find 1
 post.destroy if post
-puts "deleted" post.destroyed?
+puts "deleted" if post.destroyed?
 
 post = Post.find 1
 post.destroy! # raises when delete failed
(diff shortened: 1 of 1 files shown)


docs/crud.md
```diff
@@ -105,8 +105,8 @@ Delete a specific record.
 
 ```crystal
 post = Post.find 1
-post.destroy
-puts "deleted" unless post
+post.destroy if post
+puts "deleted" post.destroyed?
```
Blacksmoke16:
```suggestion
puts "deleted" if post.destroyed?
```

PR #352, opened 2019-07-31
Count to Counts with MultiValue and Count fixed

unequaled86:
so i can use count without group by and counts with group by


drujensen:
@unequaled86 Thanks for the contribution.  

Can you provide some details as to why you want this change?  I have never seen `counts` used before and doesn't really roll off the tongue for me.  

Also, its kinda frowned on to have two methods do the same thing.  Please change my mind.  ;-)

unequaled86:
```
page = 2
limit = 10

 # its add_aggregate_field for group by
posts.where(:title, :like, "%#{search_text}")

 # (already not working with where bc db.query_one)
count = posts.count.run

# pagination
posts.limit(limit)
posts.offset((page - 1)*limit)
results = posts.select

{"results" => results, "pages_size" => count.fdiv(limit).ceil.to_i, "posts_size" => count.to_i}.to_json
```
in that rest api scenario
it was giving error because of `group by` clause on `count` with `db.query_one`, unable to use with
`where`. also i need single value `count` with `where` method without `group by`
i created `counts` its fixed version of old `count`
deleted `group_by` from `count`
so we can use `count` for that scenario without  and `group by` `counts` with `group by`
fast reasonable solution

maybe we can change `count` method with additional parameters like a add `group_by` to sql and always `multivalue` in the future

drujensen:
@unequaled86 ok.  Thanks for the explanation.  

I'm not a fan of this fix.  I recommend finding a way to get `count` to work with and without a `group_by`.

Is this something you would be interested in looking into?

unequaled86:
@drujensen rewrited and its look like acceptable

spec/granite/query/assemblers/mysql_spec.cr
```diff
@@ -9,18 +9,18 @@ require "../spec_helper"
       end
 
       it "simple counts" do
-        builder.count.raw_sql.should match ignore_whitespace "select count(*) from table"
+				sql = "select count(*) from table"
```
drujensen: fix indentation.

spec/granite/query/assemblers/mysql_spec.cr
```diff
@@ -9,18 +9,18 @@ require "../spec_helper"
...
+        builder.count.raw_sql.should match ignore_whitespace sql
       end
-    end
 
-	context "counts" do
-      it "adds group_by fields for where/count queries" do
+			it "adds group_by fields for where/count queries" do
         sql = "select count(*) from table where name = ? group by name"
-        builder.where(name: "bob").counts.raw_sql.should match ignore_whitespace sql
+        builder.where(name: "bob").count(true).raw_sql.should match ignore_whitespace sql
```
drujensen: Remove passing in a boolean.  Determine this by looking at the `group_by` itself.

spec/granite/query/assemblers/sqlite_spec.cr
```diff
@@ -9,18 +9,18 @@ require "../spec_helper"
...
+        builder.count.raw_sql.should match ignore_whitespace sql
       end
-    end
 
-	context "counts" do
-      it "adds group_by fields for where/count queries" do
+			it "adds group_by fields for where/count queries" do
         sql = "select count(*) from table where name = ? group by name"
-        builder.where(name: "bob").counts.raw_sql.should match ignore_whitespace sql
+        builder.where(name: "bob").count(true).raw_sql.should match ignore_whitespace sql
```
drujensen: passing a boolean to `count` is not intuitive and is a breaking change for existing users.  To be honest, I don't see why we need a `group by name`  added at all here.  Only if someone specifies a `group_by` should it be included.

src/granite/query/assemblers/base.cr
```diff
@@ -120,16 +120,20 @@ module Granite::Query::Assembler
...
     end
 
-    def count : Executor::Value(Model, Int64)
+    def count(use_group_by : Bool = false) : (Executor::MultiValue(Model, Int64) | Executor::Value(Model, Int64))
       sql = build_sql do |s|
         s << "SELECT COUNT(*)"
         s << "FROM #{table_name}"
         s << where
-        s << group_by
+        s << group_by if use_group_by
```
drujensen: I think you should remove the `group_by` for now.  This shouldn't be automatically adding a group by unless someone specifies it.  Not sure why this was added to begin with.

docs/querying.md
```diff
@@ -38,6 +38,18 @@ With direction
 Post.order(created_at: :desc, title: :asc)
 ```
 
+## Group
+
+Order is using the QueryBuilder and supports providing an GROUP BY clause:
```
drujensen: Order -> Group By

docs/querying.md
```diff
@@ -38,6 +38,18 @@ With direction
...
+```crystal
+posts = Post.group(:published)
+```
+
+Multiple fields
+```crystal
+Post.order([:published, :author_id])
```
drujensen: `Post.group_by`

spec/granite/query/assemblers/mysql_spec.cr
```diff
@@ -3,13 +3,19 @@ require "../spec_helper"
 {% if env("CURRENT_ADAPTER").id == "mysql" %}
   describe Granite::Query::Assembler::Mysql(Model) do
     context "count" do
-      it "adds group_by fields for where/count queries" do
-        sql = "select count(*) from table where name = ? group by name"
-        builder.where(name: "bob").count.raw_sql.should match ignore_whitespace sql
+			it "counts for where/count queries" do
+				sql = "select count(*) from table where name = ?"
+				builder.where(name: "bob").assembler.count.raw_sql.should match ignore_whitespace sql
```
drujensen: You shouldn't add `assembler` here.  It's not intuitive and it should work without it.

docs/querying.md
```diff
@@ -38,6 +38,18 @@ With direction
 Post.order(created_at: :desc, title: :asc)
 ```
 
+## Group
+
+Order is using the QueryBuilder and supports providing an GROUP BY clause:
+```crystal
+posts = Post.group(:published)
```
drujensen: I think I prefer `group_by` instead.  wdyt?

unequaled86:
me too prefer `group_by` and `order_by` over `order`
it was `order` so i did `group`

Blacksmoke16:
Minor, but you should run `crystal tool format` as well.  Standard is to use spaces vs tabs for indentation.

drujensen:
hmm, I see what you mean.  Rails uses `order` and `group_by`.  I think it's ok to follow their lead.

docs/querying.md
unequaled86: me too prefer `group_by` and `order_by` over `order`
it was `order` so i did `group`

unequaled86:
i run `crystal tool format` before commit
i can change it to `group_by`

spec/granite/query/assemblers/mysql_spec.cr
unequaled86: i need to use assembler for receive count's raw sql
if i use only `raw_sql` its giving select's raw sql

Blacksmoke16:
> i run `crystal tool format` before commit

Perfect, could you update the tabs to use spaces?  Guess it doesn't do that for you.

unequaled86:
> > i run `crystal tool format` before commit
> 
> Perfect, could you update the tabs to use spaces? Guess it doesn't do that for you.
i did now

spec/granite/query/assemblers/mysql_spec.cr
drujensen: hmm, something doesn't sound right if that is the case.  I would expect `builder.where(name: "bob).count.raw_sql` to return the correct sql.

unequaled86: `builder` have this code so `where(name: "bob")` `return builder`
```
  def raw_sql
    assembler.select.raw_sql
  end
```
when we call `select` or `count` or `delete` its `return executer` and its have `raw_sql` their sql not same

docs/querying.md
```diff
@@ -38,6 +38,18 @@ With direction
 Post.order(created_at: :desc, title: :asc)
 ```
 
+## Group
```
drujensen: Update the documentation to `Group By` and `Post.group_by`

spec/granite/query/assemblers/mysql_spec.cr
```diff
@@ -3,13 +3,19 @@ require "../spec_helper"
 {% if env("CURRENT_ADAPTER").id == "mysql" %}
```
drujensen: Add specs for `group_by`.  Include a test that uses multiple columns i.e. `Post.group_by([:name, :age])`

src/granite/query/builder.cr
```diff
@@ -178,6 +201,10 @@ class Granite::Query::Builder(Model)
     assembler.select.run
   end
 
+  def count
+    assembler.count
```
drujensen: :+1:

docs/crud.md
```diff
@@ -65,7 +65,7 @@ end
 post = Post.first! # raises when no records exist
 ```
 
-### where, order, limit, offset
+### where, order, limit, offset, group
```
drujensen: group -> group_by

drujensen:
@unequaled86 hate to do this to you but we merged a (large breaking change)[https://github.com/amberframework/granite/pull/346] that is causing issues in this PR.  Can you take a look?

unequaled86:
probably fixed

drujensen:
@unequaled86 Thank you

PR #353, opened 2019-07-31
Upgrade to Crystal v0.30.0

bcardiff:
This PR adds support for Crystal 0.30.0 without dropping the support for earlier versions.

The fences 0.30.0-0 can be replaces by 0.30.0 after the compiler is released if desired.

Read about `UUID#to_slice` deprecation in https://github.com/crystal-lang/crystal/pull/7901 . I guess the underlying design might need an iteration to have a cleaner implementation, probably without using `Slice` to store a `UUID`.

crystal-db shards will need to be updated also as stated in https://github.com/bcardiff/granite/commit/1d8a77c08dc3bbde4b97b28f0b778b7f7ada0696 after each of them are released. Do not use my forks as dependencies.

* https://github.com/crystal-lang/crystal-db/pull/108
* https://github.com/crystal-lang/crystal-mysql/pull/80
* https://github.com/crystal-lang/crystal-sqlite3/pull/43
* https://github.com/will/crystal-pg/pull/178

### DIFF
--- a/spec/granite/converters/uuid_spec.cr
+++ b/spec/granite/converters/uuid_spec.cr
@@ -25,7 +25,11 @@ describe Granite::Converters::Uuid do
     describe ".to_db" do
       it "should convert a UUID enum into Bytes" do
         uuid = UUID.random
-        Granite::Converters::Uuid(Bytes).to_db(uuid).should eq uuid.to_slice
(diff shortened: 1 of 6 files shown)


drujensen:
@bcardiff I'm wondering if the crystal version fences should be included?  I'm assuming an older amber application will still work with crystal 0.30.0 but it will show a deprecation warning, correct?  Wouldn't a warning be preferred in that scenario?

PR #370, opened 2019-10-08
Refactor initialization

Blacksmoke16:
Preface:  This is still a WIP and are a few things left to do.  However I wanted to make the draft to start to discuss how we expect this to ultimately work.

### Changes
* Removes `*::Serializable`
  * People can just include it/whatever serialization lib they want to use (as long as it works with properties)
* Removes `Granite::Type.convert_type`
  * Granite no longer handles converting strings/etc to their proper type
* Removes `set_attributes`
  * Changes implementation to use overloaded `#initialize` methods
  * Updates `#create` and `#update` to use them
  * No longer accepts `Hash`

### Initialization/ivars
The big questions I have are: 

1. Should we support `Model.new hash`?
2. Should we define properties to be nilable under the hood but hidden via `property!`?
3. Should we just stop trying to define `#initialize` for the user in first place?

From what I did so far, these are my findings (not diving in to deep).

Supporting 1:
* Moves compile time errors to be runtime errors
  * Such as passing `String` when column is typed as a `Bool`
  * Hashes come with a Union of types that tuples and namedTuples do not which has to be dealt with

Supporting 2:
* Allows compile errors when trying to set a not nilable column to `nil`
  * Current implementation uses `nil` or the column's default if defined
* Would make it so partial hydration of a model is not possible
  * I.e. would have to return a namedTuple/hash if you didn't want _ALL_ the columns
* Would make it harder to work with deserializing from JSON
  * Since a value would have to be provided for every column, even if some are set internally.  I.e. think of some being `readonly`, not able to be set by the client.

Supporting 3:
* Would remove the need for any of this and let the user define initializers that work exactly how they want them to.
  * Would have more control/compile time errors since properties would be typed more accurately
  * Would remove some _magic_ since it would require an initializer just like any other Crystal class

Once we figure this out I'll update tests to make sure everything works as expected for the chosen implementation.


src/granite/base.cr
```diff
@@ -40,35 +40,46 @@ abstract class Granite::Base
...
 
     # Returns true if the record is persisted.
     disable_granite_docs? def persisted?
       !(new_record? || destroyed?)
     end
 
-    disable_granite_docs? def initialize(**args : Granite::Columns::Type)
-      set_attributes(args.to_h.transform_keys(&.to_s))
+    # Consumes the result set to set self's property values.
+    disable_granite_docs? def initialize(result : DB::ResultSet) : Nil
```
drujensen: love the overload here

src/granite/base.cr
```diff
@@ -40,35 +40,46 @@ abstract class Granite::Base
...
+            end
+          end
+        {% end %}
+      {% end %}
     end
 
-    disable_granite_docs? def initialize(args : Granite::ModelArgs)
-      set_attributes(args.transform_keys(&.to_s))
-    end
+    disable_granite_docs? def initialize(*args, **named_args)
```
drujensen: will this would work if someone defines their own transient properties?

src/granite/columns.cr
```diff
@@ -32,66 +32,26 @@ module Granite::Columns
...
-      end
-    {% else %}
-      def {{decl.var.id}}=(@{{decl.var.id}} : {{type.id}}); end
-
-      def {{decl.var.id}} : {{type.id}}
-        raise NilAssertionError.new {{@type.name.stringify}} + "#" + {{decl.var.stringify}} + " cannot be nil" if @{{decl.var}}.nil?
-        @{{decl.var}}.not_nil!
-      end
-    {% end %}
+    {% if !primary || (primary && !auto) %} property{{(nilable || !decl.value.is_a?(Nop) ? "" : '!').id}} {% else %} getter {% end %} {{decl.var}} : {{decl.type}} {% unless decl.value.is_a? Nop %} = {{decl.value}} {% end %}
```
drujensen: since all the columns are defined as properties (accept id), can we just use the default initializer?

src/granite/transactions.cr
```diff
@@ -196,30 +181,19 @@ module Granite::Transactions
...
 
-  def update(**args)
-    update(args.to_h)
-  end
-
-  def update(args : Granite::ModelArgs)
-    set_attributes(args.transform_keys(&.to_s))
-
+  def update(*args, **named_args)
+    initialize(*args, **named_args)
```
drujensen: Is it ok to call `initialize`?

src/granite/type.cr
```diff
@@ -62,24 +48,4 @@ module Granite::Type
   def from_rs(result : DB::ResultSet, t : Time?.class) : Time?
     result.read(Time?).try &.in(Granite.settings.default_timezone)
   end
-
-  {% for type, method in NUMERIC_TYPES %}
```
drujensen: Will this logic be moved to a `from_params` method?  Or are you thinking this will be removed from Granite?

Blacksmoke16: It depends on the approach we take with option 1. If we support hash it would just work with `#new`. If we don't want to support hashes I was thinking of another overload that accepts a `HTTP::Params` object.  However, I haven't messed with that at all yet.

src/granite/columns.cr
Blacksmoke16: Property macro doesn't create an initializer, if we go with option 3, then the user could define whatever they want it to be.

src/granite/base.cr
Blacksmoke16: Oo no, might have to introduce an annotation or something to mark a property as "virtual".  The problem with just doing `@type.instance_vars` (without the annotation checks) would be it would try to set _any_ ivar.  Whether thats internal stuff like `@new_record`, ivars from other shards/modules etc.

src/granite/transactions.cr
Blacksmoke16: 🤷‍♂ probably?

src/granite/columns.cr
drujensen: I am thinking that we don't create an initializer and leave it up to the user how to set the properties.  This will make the class less magic and will handle the `virtual` fields.  They can use `getter` or `property` macros or override the getter and setter methods manually.  Leave it up to the user how to set the properties.

Blacksmoke16: That is one of the benefits.  However I'll leave this open for a while as I'd rather have other maintainers/users give their thoughts as well.

src/granite/base.cr
robacarp: Why not just look over the list of fields we home we want to initialize? My memory is out of date but I recall a FIELDS constant that’s built by the individual field calls.

src/granite/transactions.cr
robacarp: It might be better to put the functionality in a shared method elsewhere, even if it means repeating the code so that crystal still sees the property initialization correctly. If the idea here is to encourage developers to create their own init methods, we should avoid calling it in hidden places. Developers are likely to attach weird logic there.

robacarp:
I like the idea of empowering the developer with the ability to override the functionality of the initalizer, but I also to boilerplate to a minimum. My priorities go something like this:

1) provide compile time errors where possible, 
2) avoid surprising the developer, (eg. by calling the init method when they’re not expecting it)
3) “just work” out of the box without relying on generating a bunch of boilerplate, and
4) Empower the developer to make changes where they want to extend the functionality without requiring them to reinvent the wheel

This seems like it might be closely related with the idea that we could remove the column macro and replace it with a columns macro.

src/granite/transactions.cr
Blacksmoke16: Maybe some private initializer that's only used internally?

src/granite/base.cr
Blacksmoke16: The mutable constant stuff got removed when we switched to using annotations.

Blacksmoke16:
> This seems like it might be closely related with the idea that we could remove the column macro and replace it with a columns macro.

How would this help?  Personally I'd rather see the `column` macro replaced with the user defining properties/annotations themselves.  

```cr
@[Granite::Column(auto: true, primary: true, nilable: true)]
getter id : Int64?

@[Granite::Column(nilable: true)]
property year : Int32?

@[Granite::Column]
property! name : String
```

One thought I had was if we don't define an initializer, and make the ivars nilable behind the scenes, then you could still do `model = User.new` then set your values.  But if you wanted to do `User.new name: "Jim", age: 19` you would need to define that initializer in your class (just like any other Crystal class).

I suppose even if we still define our built in initializers, they could still define their own if they wanted more custom behavior?

I'm conflicting on how to balance everything :/

drujensen:
@robacarp @Blacksmoke16 My take is that the user will not expect (Roberts' #2) an initializer to be created for me.  I would not expect a not_nilable property to be nilable, so I would expect that I would need to create an `initializer` to set that value when I perform a `new`.

To me, the `column` macro inherits from the `property` macro but adds ORM mapping for the `save` method.  Setting the properties should be the same as you do in Crystal.  

I vote we remove some of the magic.  maybe we can remove the `create` and `update` methods as well and have the user only perform a `save`.  If they want to bulk update properties, they can define their own `update` method.  WDYT?  

Also, adding `to_params` and `from_params` with an `include ParamsSerializer` similar to `to_json` and `from_json` with an `include JsonSerializer` makes the API clean and easy to grok.

Blacksmoke16:
> My take is that the user will not expect (Roberts' #2) an initializer to be created for me.

Right, if we go that way the only initializer that will be defined by default is the argless one that comes with `Class`.  E.x.

```cr
abstract class Base
end

class User < Base
  property name : String?
end

User.new
```
But if you want to initialize `@name` you would have to define your own.
```cr
class User < Base
  property name : String?
  
  def initialize(@name : String); end
end

User.new "Fred"
```

> I would not expect a not_nilable property to be nilable

I think there might be some misunderstanding on our viewpoints of this.  My thought is that _internally_ all properties should be nilable, not the getter/setter for those properties.  The current implementation of using `property!` for the not nilable columns handles this for us.  E.x.

`column name : String` gets expanded to `property! name : String` with the annotation applied of course.

Using this class:
```cr
class User
  property! name : String
end

# Are still able to new up a model
user = User.new # => @name=nil

# Trying to access the value at this point would raise an exception
# After https://github.com/crystal-lang/crystal/pull/8296 is merged it would be like
user.name # => Unhandled exception: User#name cannot be nil (NilAssertionError)

# The default getter removes `Nil` from the union so you won't run into like
# `undefined method '+' for Nil (compile-time type is (String | Nil))`
user.name = "Fred"
typeof(user.name) # => String

# A nilable query method is also added
typeof(user.name?) # => String?

# `Nil` is removed from the setter types so this won't compile
#
# Error: no overload matches 'User#name=' with type Nil
#
# Overloads are:
#  - User#name=(name : String)
user.name = nil
```
IMO this gives us more options internally while still giving good protection to the end user; such as partial hydration, allowing some properties to be set later (that would be not null in the DB) instead of requiring everything up front.  An example of when this would be useful would be: Imagine a JSON api where a user can create a blog post with a body like:

```json
{
  "title": "Title of blog",
  "published": true,
  ...
}
```

Every post was created by a user, so there would also be a `column author_id : Int32` (or could be a like `has_one User`) that represents the user that created that post.  If every field was not nilable, you would be required to supply it within the POST body so it wouldn't fail on `Post.from_json`.  Having it nilable allows you to consume the body then within your controller do like `post.author_id = current_user.id`.

> Also, adding `to_params` and `from_params` with an `include ParamsSerializer`

I'm assuming this would handle `HTTP::Params`?  IMO this is one of the benefits the switch to annotations brings, modules that just do stuff to properties just work 💯.

drujensen:
@Blacksmoke16 thanks for the details.

> column name : String gets expanded to property! name : String with the annotation applied of course.

would it make more sense to use `column! name : String` so that users know what to expect when the macro expands?

Blacksmoke16:
I don't think so.  Documentation would probably do the trick.  I'm not sure what benefit it would bring, the type of the column should be enough indication of what it will do.

src/granite/transactions.cr
robacarp: Yeah that's what I mean by duplicating the code.

Blacksmoke16: If we go the "remove default initializer route", which seems to be the popular opinion, then this wouldn't matter anymore.  

We should be able to support `#create` as it's essentially the same flow as calling `#new` but also calls `#save` before returning the object.  Supporting `#update` will be harder.

spec/granite/columns/primary_key_spec.cr
```diff
@@ -11,10 +11,6 @@ describe "#new" do
 end
 
 describe "#new(primary_key: value)" do
-  it "ignores the value in default" do
-    Parent.new(id: 1_i64).id.should eq(nil)
-  end
```
Blacksmoke16: This wouldn't compile anymore assuming the user doesn't include `id` within the initializer (which wouldn't make sense since its auto increment anyway).

spec/granite/transactions/save_spec.cr
```diff
@@ -15,12 +15,6 @@ describe "#save" do
     parent.persisted?.should be_false
   end
 
-  pending "does not save a model with type conversion errors" do
-    model = Comment.new(articleid: "foo")
-    model.errors.size.should eq 1
-    model.save.should be_false
-  end
```
Blacksmoke16: This wouldn't compile anymore since the initializer would be expecting an `Int64?` not `String`.  If the user wanted they could accept both and set `articleid` within the initialize method; converting the input value if needed.

Blacksmoke16: This wouldn't compile anymore since the initializer would be expecting an `Int64?` not `String`.  If the user wanted they could accept both and set `articleid` within the initialize method; converting the input value if needed.

spec/granite/transactions/save_spec.cr
```diff
@@ -46,17 +40,6 @@ describe "#save" do
...
-  pending "does not update when the conflicted primary key is given to the new record" do
-    parent1 = Parent.new
-    parent1.name = "Test Parent"
-    parent1.save.should be_true
-
-    parent2 = Parent.new
-    parent2.id = parent1.id
-    parent2.name = "Test Parent2"
-    parent2.save.should be_false
-  end
```
Blacksmoke16: Won't be able to do this anymore by default

```
In spec/granite/transactions/save_spec.cr:49:13

 49 | parent2.id = parent1.id
              ^-
Error: undefined method 'id=' for Parent
```

src/granite/transactions.cr
```diff
@@ -187,16 +185,6 @@ module Granite::Transactions
     save || raise Granite::RecordNotSaved.new(self.class.name, self)
   end
 
-  def update(*args, **named_args)
```
Blacksmoke16: We could probably support `#update` if we wanted.
1. Call `#initialize` with the values
2. Support only named arguments and do a loop with a case statement to set the ivar when there is a match

Blacksmoke16:
@robacarp @drujensen I updated the code to support no default initializers.  Also left some comments on test cases that I removed with the reasoning behind why they were removed.

So far from there I would suggest checking out this branch on some small local projects and see how things work.  From there we'll still need to:
1. Include an overload that supports `HTTP::Params`
2. Decide if we want to support `#update`
3. Update the docs
4. Anything else I'm missing?

Blacksmoke16:
@robacarp @drujensen One other thing I thought of.  Currently we require the PK to be nilable since it wouldn't have one until it was saved.  However that makes things like 
```cr
user = User.first!
user.id + 1
```
fail since `#id` is nilable.  What are your thoughts on how to handle this?

I'm thinking the majority of cases are going to be working with objects that originate from a DB query, thus having a PK.  So:

1. Keep it the way it is, requiring the user to handle `nil` as a possible PK value everywhere
2. Switch to requiring the PK be not-nilable and use `getter!`, in order to keep it nilable but hide that from the user.  Would still allowing checking for a PK value using like `#id?`.
3. Remove the type requirement on the PK and just define either a `getter` or `getter!` depending on if the type they give it is nilable or not-nilable.

robacarp:
@Blacksmoke16 hiding a nilable field from the implementer feels risky to me.

I think at this point in my Crystal experience, I'd like to see a different object which represents persisted database entries and has a non-nilable primary key and other fields. I digress.

That said, I think the switch you made here is pretty straightforward. Having a nilable PK field is a pain over and over again -- any nilable fields are.

Blacksmoke16:
> hiding a nilable field from the implementer feels risky to me.

I should probably stop saying `hide`, it's more like just changing the default getter to remove nil from the type.

> I'd like to see a different object which represents persisted database entries and has a non-nilable primary key and other fields

A whole separate object isn't ideal IMO.  What would that get us that `getter!` or `property!` don't? 
 If the user does `model.id`, most of the time i would say is they got it from a db query and it's going to exist.  `getter!` makes the use case the default, while still allowing for a nilable getter, via `#id?`.

robacarp:
Well, the point of using a non-nilable field is that you get compile time checks against it. And for situations where the field is actually nil able, an unsaved model, you get compile time enforcement that the field’s nil-ability must be handled. `getter!`’s nil check is tunable code, which throws an exception, which should be handled _somewhere_.

With the `getter!` implementation, when someone makes a mistake and the server goes kaput, how do we lead them down the right path?

Blacksmoke16:
> With the `getter!` implementation, when someone makes a mistake and the server goes kaput, how do we lead them down the right path?

My thinking is that this is not much different than before where we had `model.column` and `model.column!`.  We would have documentation that says if you declare a type as non-nilable, then it will be implemented with `property!`  If there is a possibility that a non-nilable column is nil, such as that column wasn't selected, or hasn't been saved yet etc, then they should use `model.column?`.

drujensen:
@Blacksmoke16 @robacarp Where did we end up with this PR?  This is almost a year old and not sure what we decided.

Blacksmoke16:
@drujensen IIRC, we never came to a consensus if this is something we wanted to move forward with; given it's a breaking change and changes the semantics quite a bit.

PR #372, opened 2019-11-29
Make db error verbose

msa7:


Blacksmoke16:
Could we maybe get some specs on this?

Also I think you could use the block version of `#join`, then just grab `message` off each error?

drujensen:
> Also I think you could use the block version of `#join`, then just grab `message` off each error?

Looks like `to_s` already returns the message:
https://github.com/crystal-lang/crystal/blob/41bd18fbea4aec50aad33aa3beb7a0bf30544186/src/exception.cr#L42

src/granite/exceptions.cr
```diff
@@ -3,7 +3,7 @@ module Granite
     getter model : Granite::Base
 
     def initialize(class_name : String, model : Granite::Base)
-      super("Could not process #{class_name}")
+      super("Could not save #{model.errors.map(&.to_s).join(", ")}")
```
drujensen: Tests are failing and the message is more cryptic than before IMO.
```
      Expected: "Could not process Parent"
                got: "Could not save Name Name cannot be blank"
```
Maybe `"Could not save #{class_name}. #{model.errors.map(&.to_s).join(", ")}" would be more clear.

PR #388, opened 2020-04-08
Add #group_by to list of methods delegated to the Query Builder in BuilderMethods

### LINKED ISSUE #387: Issue with `group_by` ?
I'm using granite `0.19.0` and there seems to be an issue with `group_by` or I'm making a mistake somewhere.

If I define a model, I can query it with `#all` and `#order` just fine, but `#group_by` gives me:

```crystal
Showing last frame. Use --error-trace for full trace.

In src/commands/build/content_loader.cr:28:24

 28 | puts ContentItem.group_by(:published_at)
                       ^-------
Error: undefined method 'group_by' for ContentItem.class
```

What am I doing wrong?

sardaukar:
Not sure how to spec this out, or even test my fork of the shard locally - any pointers?

Fixes #387

### DIFF
--- a/src/granite/query/builder_methods.cr
+++ b/src/granite/query/builder_methods.cr
@@ -12,5 +12,5 @@ module Granite::Query::BuilderMethods
     Builder(self).new(db_type)
   end
 
-  delegate where, order, offset, limit, to: __builder
+  delegate where, order, offset, limit, group_by, to: __builder
 end


Blacksmoke16:
I would update the version of granite you're using in your project to your forked branch.  I.e.

```yaml
granite:
  github: sardaukar/granite
  branch: add-group_by-to-builder-methods
```

Then see if everything works.

If everything works now, add a spec to https://github.com/amberframework/granite/blob/master/spec/granite/querying/query_builder_spec.cr.

Imagine the output of using `.group_by` is like `Hash(String, T)`?  If so creating three records, two with the same name, and another that is diff, then using `group_by` to ensure the value of the one has 2 and the value of the other has 1.

sardaukar:
Closing this PR due to shards error:

```
Failed git tag --list --contains add-group_by-to-builder-methods --column=never (malformed object name add-group_by-to-builder-methods). Maybe a commit, branch or file doesn't exist?
```
Gonna try with a shorter name...?

sardaukar:
@Blacksmoke16 I don't get it :shrug: 

Changed the branch name to `add-group-by` and finally `dev` to get rid of hyphens but still getting:

```
Failed git tag --list --contains dev --column=never (malformed object name dev). Maybe a commit, branch or file doesn't exist?
```

when I run `shards`. Any thoughts?

Blacksmoke16:
I'm not sure I follow.  What exactly are you doing?

Really you just need to do like
```
dependencies:
  granite:
    github: sardaukar/granite
    branch: add-group-by
```

You don't need to tag anything, just update the github reference and branch in your `shard.yml` to point to your repo/branch.

sardaukar:
I'm doing just that, but I get that error message from `shards`

Blacksmoke16:
Are you sure?  I don't suppose you can share your `shard.yml` of your project where you're trying to test with this branch?

PR #389, opened 2020-04-12
Support Crystal 0.34.0

drujensen:


Blacksmoke16:
There's a bug in the current ameba version. Probably should save that one for later.

drujensen:
@Blacksmoke16 So I should roll that back?

Blacksmoke16:
If you can fix them then its fine.  But I think the macro ones will fail even tho they are valid.

drujensen:
@Blacksmoke16 any ideas why Postgres 9.6 and 11 are failing?

drujensen:
nm, forgot to update the crystal docker file to 0.34.0

drujensen:
strange, pg 9.6 tests passed on my laptop when I updated the crystal version to 0.34.0

drujensen:
looks like postgres docker image requires a password now.

drujensen:
@Blacksmoke16 can you provide a quick review?

PR #392, opened 2020-04-13
Adding #group_by to list of methods delegated to query builder

sardaukar:


sardaukar:
I'm kinda confused by `#group_by`'s behavior. I must be doing something wrong...?

drujensen:
@sardaukar based on the test written, I think you might have been expecting `Enumerable#group_by` functionality instead of a `SQL GROUP BY` clause.

https://crystal-lang.org/api/0.34.0/Enumerable.html#group_by(&block:T-%3EU)forallU-instance-method

https://www.w3schools.com/sql/sql_groupby.asp

PR #395, opened 2020-04-24
Fix #394 by stripping the module name from the table name

### LINKED ISSUE #394: Sqlite3 Exception when database name contains a colon
The following code produces a table with the name `cwiki::user`

```crystal
module Cwiki
  class User < Granite::Base
  end
end
```

This is an issue as statements that reference the table look something like this: "SELECT * FROM cwiki::user". Sqlite throws errors on table names that contain a colon unless the name is surrounded by quotation marks.

I think this should be an easy fix, but I'm not familiar with this codebase. I'll fork and check it out and report back if I figure things out.

3n-k1:
Fixes #394

### DIFF
--- a/spec/granite/table/table_spec.cr
+++ b/spec/granite/table/table_spec.cr
@@ -9,6 +9,10 @@ describe Granite::Table do
     it "sets the table name based on class name if not specified" do
       SongThread.table_name.should eq "song_thread"
     end
+
+    it "strips the namespace when defining the default table now" do
+      MyApp::Namespace::Model.table_name.should eq "model"
+    end
   end
 
   describe ".primary_name" do
--- a/spec/spec_models.cr
+++ b/spec/spec_models.cr
@@ -509,6 +509,10 @@ end
     column my_enum : MyEnum?, column_type: "TEXT", converter: Granite::Converters::Enum(MyEnum, String)
   end
 
+  class MyApp::Namespace::Model < Granite::Base
+    column id : Int64, primary: true
+  end
+
   {% if env("CURRENT_ADAPTER") == "pg" %}
     class ConverterModel < Granite::Base
       connection {{ adapter_literal }}
--- a/spec/spec_models.cr
+++ b/spec/spec_models.cr
@@ -510,6 +510,8 @@ end
   end
 
   class MyApp::Namespace::Model < Granite::Base
+    connection {{ adapter_literal }}
+
     column id : Int64, primary: true
   end
 
--- a/src/granite/table.cr
+++ b/src/granite/table.cr
@@ -45,7 +45,7 @@ module Granite::Tables
     def table_name : String
       {% begin %}
         {% table_ann = @type.annotation(Granite::Table) %}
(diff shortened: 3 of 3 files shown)


Blacksmoke16:
Nice!  A spec to go with it would be :100:.

3n-k1:
I'll be honest, I'm pretty new to crystal and have no idea how to write a spec. Could you point me in the direction of a tutorial for it?

Blacksmoke16:
@3n-k1 https://crystal-lang.org/reference/guides/testing.html would be a good start.

After reading that, define a new model type within: https://github.com/amberframework/granite/blob/master/spec/spec_models.cr

Something like
```cr
class Foo::Bar::Model < Granite::Model
  column id : Int64, primary: true
end
```

The spec would be like:
```cr
it "strips the namespace when defining the default table now" do
  Foo::Bar::Model.table_name.should eq "model"
end
```

3n-k1:
Great, thank you very much!

Blacksmoke16:
@3n-k1 Sorry, `Granite::Base` not `Granite::Model` my bad 😆

3n-k1:
No problem at all :)

Blacksmoke16:
Ogh, apparently there is already a `Foo` model somewhere.

Lets go with: `MyApp::Namespace::Model`.

3n-k1:
It's a good thing you're here, I'd never be able to debug this haha

spec/granite/table/table_spec.cr
```diff
@@ -9,6 +9,10 @@ describe Granite::Table do
     it "sets the table name based on class name if not specified" do
       SongThread.table_name.should eq "song_thread"
     end
+
+    it "strips the namespace when defining the default table now" do
+      Foo::Bar::Model.table_name.should eq "model"
```
Blacksmoke16, line 14:
```suggestion
      MyApp::Namespace::Model.table_name.should eq "model"
```

3n-k1: Whoops, sorry, missed that

3n-k1:
Running the tests locally this time to be safe

Blacksmoke16:
Looks like specs are still failing.  I'll take a look.

3n-k1:
That's strange, since they run fine for me in docker. I appreciate you looking into it.

Blacksmoke16:
Sorry for the delay, imma check this out today.

spec/spec_models.cr
```diff
@@ -509,6 +509,10 @@ end
     column my_enum : MyEnum?, column_type: "TEXT", converter: Granite::Converters::Enum(MyEnum, String)
   end
 
+  class MyApp::Namespace::Model < Granite::Base
+    column id : Int64, primary: true
+  end
```
Blacksmoke16, line 514:
```suggestion
  class MyApp::Namespace::Model < Granite::Base
    connection {{ adapter_literal }}
  
    column id : Int64, primary: true
  end
```

3n-k1:
Ah, can't believe I missed that. Thanks for the merge!

PR #397, opened 2020-05-04
Fix read_attribute, see #396

### LINKED ISSUE #396: Read_attribute does not work when column is array
```
In lib/granite/src/granite/columns.cr:146:58

 146 | def read_attribute(attribute_name : Symbol | String) : DB::Any
                                                              ^------
Error: method must return (Bool | Float32 | Float64 | Int32 | Int64 | Slice(UInt8) | String | Time | Nil) but it is returning (Array(Int32) | Array(Int64) | Bool | Int32 | Int64 | String | Time | Nil)
```

msa7:
Fixes #396

### DIFF
--- a/spec/granite/columns/read_attribute_spec.cr
+++ b/spec/granite/columns/read_attribute_spec.cr
@@ -3,7 +3,7 @@ require "../../spec_helper"
 describe "read_attribute" do
   # Only PG supports array types
   {% if env("CURRENT_ADAPTER") == "pg" %}
-    it "ably to read arrays" do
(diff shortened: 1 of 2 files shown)


Blacksmoke16:
Nice!  Maybe add a spec to ensure this doesn't regress?

drujensen:
@msa7 will you be able to add a spec?  if not, maybe someone can pick up the mantel and get this over the finish line.

msa7:
@drujensen I will do specs

msa7:
I hope the spec is enough

PR #401, opened 2020-05-11
Add a to_json method for Granite::Error

3n-k1:
I wasn't sure if this was something that was at all wanted, so I figured I'd send a PR to be safe.

This PR manually defines a `to_json` method for `Granite::Error`. We can't `include JSON::Serializable` due to the type of `Granite::Error.field`, so a manual solution is required.

### DIFF
--- a/spec/granite/error/error_spec.cr
+++ b/spec/granite/error/error_spec.cr
@@ -1,0 +1,7 @@
+require "../../spec_helper"
+
+describe Granite::Error do
+  it "should convert to json" do
+    Granite::Error.new("field", "error message").to_json.should eq %({"field":"field","message":"error message"})
+  end
+end
--- a/src/granite/error.cr
+++ b/src/granite/error.cr
@@ -4,6 +4,13 @@ class Granite::Error
   def initialize(@field : (String | Symbol | JSON::Any), @message : String? = "")
   end
 
+  def to_json(builder : JSON::Builder)
+    builder.object do
+      builder.field "field", @field
+      builder.field "message", @message
+    end
+  end
+
   def to_s(io)
     if field == :base
       io << message


Blacksmoke16:
What is this actually needed for?

3n-k1:
I found it useful for quickly and easily returning validation errors.

```crystal
post = Post.new
post.save
if post.errors.size > 0
  post.errors.to_json
end
```

Like I said, wasn't sure if this was wanted or not. I use it in my codebase so I thought I'd send a PR.

src/granite/error.cr
```diff
@@ -1,9 +1,15 @@
+require "json"
+
 class Granite::Error
   property field, message
 
   def initialize(@field : (String | Symbol | JSON::Any), @message : String? = "")
   end
 
+  def to_json
```
Blacksmoke16, line 9: You actually want to define a `def to_json(builder : JSON::Builder)` then use the `builder` object to create the object.

See https://crystal-lang.org/api/master/JSON/Builder.html

src/granite/error.cr
```diff
@@ -1,9 +1,15 @@
+require "json"
```
Blacksmoke16, line 1: Is this actually needed?  I think we require it somewhere else but don't quote me.

src/granite/error.cr
drujensen: @Blacksmoke16 For my own learning, what are the advantages of using the JSON Builder over having just a `to_json` method with no params?

Blacksmoke16: Essentially the same idea as why you define `to_s(io : IO)` instead of `to_s`.

See https://github.com/crystal-lang/crystal/blob/master/src/json/to_json.cr#L2-L12

src/granite/error.cr
3n-k1: It was necessary due to the way I was testing. I've removed it.

3n-k1:
@drujensen for some reason I can't run tests on my local machine. Still, I'm not sure how I would test it. Just write a test that makes sure the json is valid?

drujensen:
> @drujensen for some reason I can't run tests on my local machine. Still, I'm not sure how I would test it. Just write a test that makes sure the json is valid?

Correct.  You can test just do something like:
```pseudo code
error = Granite::Error.new "field", "error message"
expect error.to_json to equal {field: "error message"}
```

spec/granite/error/error_spec.cr
```diff
@@ -0,0 +1,8 @@
+require "../../spec_helper"
+
+describe Granite::Error do
+  it "should convert to json" do
+    error = Granite::Error.new "field", "error message"
+    expect error.to_json to equal "{\"field\": \"error message\"}"
```
Blacksmoke16, line 6:
```suggestion
    Granite::Error.new("field", "error message").to_json.should eq %({"field":"error message"})
```

3n-k1: Sorry, the tests don't like my local machine, had to push to test :P

3n-k1:
thanks!

Issue #402, opened 2020-05-20
Query builder for DB specific feature?


Serdnad:
In my project I currently build a query using a chain of several `where` statements, many of them included conditionally. Now, I need to modify this query to only select rows where an array column (I'm using Postgres) contains a certain element. Is it currently possible to do this using a query builder? Effectively, I'm looking for the converse of an :in operator, a sort of :has or includes, if you will.

I'm aware I could do this without a query builder, using `#all('where ?=ANY(trusted)', [user_id])` for example, where trusted is an array type such as integer[], but then I'd lose out on the convenience of the query builder and would have to convert the entire query to a single SQL statement, as far as I understand it.

If this is something that doesn't currently exist but would be welcome, I'd be willing to give it a shot and submit a PR.

robacarp:
@Serdnad there really isn't right now. The query builder approaches SQL from the perspective of what can be queried across databases, and there isn't a pattern for including dsl segments which only apply to a specific database engine.

This general idea comes up a lot, and I can think of two ways to solve it:

1) Provide an adapter specific dsl "opener". Something like `where.pg.any_equal(user_id)` might work. This would have to be included on a per-project basis, so in addition to `require "granite"` you'd need `require "granite/pg-query-extensions"`.

1) Provide a better way to make the query builder extensible so that third party shards could extend the granite query builder to add `any_equal` or whatever other selects. This is a significant lift, but could provide some long term runway.

Either of these would require some low level tweaks to the way the query builder assembles the sql. The builder makes a lot of assumptions right now, and finding a way to render well without turning the builder into spaghetti might be a big task. Perhaps there are some other options?

Serdnad:
I see, thanks @robocarp for taking the time to explain the philosophy behind the current state of things.

It makes sense that the project would prioritize consistency across all supported databases. Unfortunately, my own project does not benefit from this, and since I'm in need of support for this sort of query in the short term, going the proper route and implementing one of the solutions you've outlined doesn't make sense for me, personally, to undertake at the moment.

However, I took your advice and thought about other options for my particular issue quite a bit, and realized that the most general solution that would solve this specific issue, as well as potentially many other query-related, possibly database-specific issues, would be to add one more method to the query builder that allowed custom SQL, similar to #all, but would still be usable as part of a surrounding query builder. In use, this would look something like:

```
users = User.where(color: "blue")
  .where("age > 18") # == where(:age, :gt, 18)
[...]

PR #403, opened 2020-05-21
Support custom queries in builder

Serdnad:
This is a first attempt at addressing #402 , and implements support for building simple queries through the use of handwritten SQL like so:
```
users = Users.where("age > ?", 18) # == where(:age, :gt, 18)
```

with the intention of providing a way to support more sophisticated queries, such as
```
users = User.where("LOWER(name) = ?", name)
```
and even database specific features, like checking for membership in an array using Postgres:
```
users = User.where("? = ANY(trusted)", id)
```

The current implementation only allows one argument per query, but it should be rather straightforward to extend this to allow multiple args per query, if that's desired.

A simple spec is included for each DB adapter.

Note: I've opened this PR as a draft because I also intend to include an addition to the docs, but wanted to make this available for review as early as possible.

### DIFF
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -16,10 +16,11 @@ Post.where(:created_at, :gt, Time.local - 7.days)
 
 Supported operators are :eq, :gteq, :lteq, :neq, :gt, :lt, :nlt, :ngt, :ltgt, :in, :nin, :like, :nlike
 
-Alternatively, where accepts a raw SQL clause, with "?" as a placeholder to avoid SQL Injection.
+Alternatively, `#where`, `#and`, and `#or` accept a raw SQL clause, with an optional placeholder (`?` for MySQL/SQLite, `$` for Postgres) to avoid SQL Injection.
 ```crystal
+# Example using Postgres adapter
 Post.where(:created_at, :gt, Time.local - 7.days)
-  .where("LOWER(author_name) = ?", name)
+  .where("LOWER(author_name) = $", name)
   .where("tags @> '{"Journal", "Book"}') # PG's array contains operator
 ```
 This is useful for building more sophisticated queries, including queries dependent on database specific features not supported by the operators above. However, **clauses built with this method are not validated.**
--- a/docs/querying.md
+++ b/docs/querying.md
@@ -16,6 +16,13 @@ Post.where(:created_at, :gt, Time.local - 7.days)
 
 Supported operators are :eq, :gteq, :lteq, :neq, :gt, :lt, :nlt, :ngt, :ltgt, :in, :nin, :like, :nlike
 
+Alternatively, where accepts a raw SQL clause, with "?" as a placeholder to avoid SQL Injection. This is useful for building more sophisticated queries, including queries dependent on database specific features not supported by the operators above. However, **clauses built with this method are not validated.**
+```crystal
+Post.where(:created_at, :gt, Time.local - 7.days)
+  .where("LOWER(author_name) = ?", name)
+  .where("tags @> '{"Journal", "Book"}') # PG's array contains operator
+```
+
 ## Order
 
 Order is using the QueryBuilder and supports providing an ORDER BY clause:
--- a/src/granite/query/builder.cr
+++ b/src/granite/query/builder.cr
@@ -28,7 +28,8 @@ class Granite::Query::Builder(Model)
   end
 
   getter db_type : DbType
-  getter where_fields = [] of NamedTuple(join: Symbol, field: String, operator: Symbol, value: Granite::Columns::Type)
+  getter where_fields = [] of (NamedTuple(join: Symbol, field: String, operator: Symbol, value: Granite::Columns::Type) |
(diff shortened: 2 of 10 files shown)


src/granite/query/builder.cr
```diff
@@ -88,6 +93,12 @@ class Granite::Query::Builder(Model)
     self
   end
 
+  def and(stmt : String, value : Granite::Columns::Type)
+    @where_fields << {join: :and, stmt: stmt, value: value}
+
+    self
+  end
+
   def or(**matches)
```
Blacksmoke16, line 102: Should probably also support `OR` clauses.

Serdnad: Ah, right. Done.

Serdnad:
Tweaked the implementation to allow for a nil value, to support running a raw SQL string with no parameter substitutions.

docs/querying.md
```diff
@@ -16,6 +16,15 @@ Post.where(:created_at, :gt, Time.local - 7.days)
 
 Supported operators are :eq, :gteq, :lteq, :neq, :gt, :lt, :nlt, :ngt, :ltgt, :in, :nin, :like, :nlike
 
+Alternatively, where accepts a raw SQL clause, with "?" as a placeholder to avoid SQL Injection.
```
Blacksmoke16, line 19:
```suggestion
Alternatively, `#where` and `#or` accept a raw SQL clause, with "?" as a placeholder to avoid SQL Injection.
```

Since `?` is not a valid placeholder in all databases, I would suggest either rewording the sentence to specify that the example is for MySQL/Sqlite, and ofc wouldn't work in Postgres.

docs/querying.md
```diff
@@ -16,6 +16,15 @@ Post.where(:created_at, :gt, Time.local - 7.days)
...
+```crystal
+Post.where(:created_at, :gt, Time.local - 7.days)
+  .where("LOWER(author_name) = ?", name)
+  .where("tags @> '{"Journal", "Book"}') # PG's array contains operator
```
Blacksmoke16, line 23: Related to the above comment.  Since this is using PG specific behavior, the other placeholder should probably be updated to match.

src/granite/query/builder.cr
Blacksmoke16: Getting some specs on it as well would be :100:.

Serdnad: I added a spec for `#where`, but I didn't see any existing specs for `#and` and `#or`, so wasn't sure what was appropriate. I can try throwing some up though.

docs/querying.md
Serdnad: The implementation actually swaps out any `?` placeholders using the `add_parameter` method, so in the case of MySQL and SQLite it'll leave in `?`, but when using the Postgres adapter it'll replace it properly with `$n`. Not sure how to link it here, but you can see the difference if you compare the spec I added for MySQL/SQLite with the Postgres one.

Maybe I should make this clearer though, in the documentation?

docs/querying.md
Serdnad: Replied above.

docs/querying.md
Blacksmoke16: Just updating the example to have placeholders specific to PG would be good enough I think.  Could possibly reword the sentence to not use a specific symbol, but just say something like `using placeholders related to the DB you're using`.

Serdnad: Hmm, but the "correct" placeholder to use is indeed `?` for all databases, because (at least for Postgres) the builder needs to assign the placeholders sequentially. This also keeps that aspect abstracted from the user. 

To be honest, I pretty much only now realized Postgres doesn't seem to use `?`, but I'm not sure what the alternative would be - maybe using just `$`, and having it replace that instead for just Postgres?

Serdnad:
Added specs for `#where`, `#or`, and `#and` as requested, and updated Postgres to use `$` as a placeholder instead of `?`.

Blacksmoke16:
@Serdnad Does your `@placeholder` stuff work?  AFAIK `$` on its own is not a valid PG placeholder.  It would need the index value to work correctly.

Serdnad:
@Blacksmoke16 This placeholder just tells the assembler that there's a placeholder in the query, so that the assembler can then replace this with the proper, final placeholder, such as `$1` in the case of Postgres. This way, you can chain several where queries without having to manually keep track of the number. This is also why my previous implementation worked even when `?` was used with PG, because the assembler would replace this with the correct placeholder, even though PG doesn't actually use `?`.

@drujensen That's a good point. I just checked, and it seems like the array would have to be formatted as `(1, 2, 3)` for that to work. Would it make sense to treat this as a separate issue, or do you think that should be included before merging?

drujensen:
@Serdnad It can be treated as a separate issue.  If @Blacksmoke16 approves, we can merge.

robacarp:
@Serdnad this looks great!

Serdnad:
Thanks @robacarp ! This turned out very differently than what I was initially looking for, but definitely for the better. Huge thanks for all the input and feedback from all of you :)

Issue #182, opened 2018-04-19
Should Granite have a built in Migrator?


eliasjpr:
Given this PR https://github.com/amberframework/granite/pull/172 and https://github.com/amberframework/granite/pull/168 which as basic support for table migration and bulk import, I have been wondering if Granite should fully support migrations?

```Crystal
class SomeMigration < Granite::Migration
  def up
      User.migrator.create
      User.execute("Custom SQL here")

      User.each do |record|
          record.status = active
          record.save!
      end
  end

 def down
      User.migrator.drop
      User.execute <<-SQL
         Custom SQL here
      SQL
  end
end
```

I think is time :) 

Open implementation detail discussion

drujensen:
@eliasjpr Granite actually did have this but it was impossible to know the intent of the developer so we removed in favor of micrate.

An example is renaming a column in the db.  You can create a new column but the old one should not be removed until the data is migrated.

The technique I used when I built this what that the migrations were always additive unless you included a flag to perform a cleanup to remove old columns.  This sorta worked but required much more care and handling than just supporting migration scripts.

faustinoaq:
Well, I think a migration system using crystal is a bit overloaded since crystal is compiled and not an scripting language, so every time you need to run a migration action you would need to compile it first :sweat_smile: 

Currently, I just edit the SQL and run `bin/amber db migrate` without requiring crystal compiler

So, by example, If I want to execute migrations in Heroku, and I want to change something for the migration itself, (some type, bugfix, etc), I can't because I would need to compile the migrations first.

Another example, if you try to use Lucky record you will notice that `lucky db.migrate` isn't instantaneous because their migrations are made in crystal and need to be compiled first.

Pepan:
Ruby starts some time too, and running migration is like start Ruby and then run script. So I thing speed will be very similar when you compile migration and then run it. Crystal already can do something like crystal run src/my/main.cr ...

igbanam:
The question which kept coming to mind while considering this is "Is it faster in Crystal or in SQL?" 

In deliberating on ☝️ with myself, I realized that the Migrator for Crystal would only be sugar on SQL. This would mean that over time, the Migrator may begin to struggle to do what SQL does best. Also, leaving the migrations in SQL forces the developer to think about their data from a data perspective — this is one of the few things which attracted me to Amber, as opposed to Rails where everything is Rails-y

I believe having migrations in Crystal may constrain what developers can do with their databases.

drujensen:
[...]

PR #413, opened 2020-07-14
DSL for migrations

drujensen:
This adds a new DSL for migrations.  It uses the same syntax as Rails migrations. 

```
require "granite/migration"

class CreatePost < Granite::Migration
  connection :pg

  def up
    create_table :post do |t|
      t.primary :id
      t.reference :user_id
      t.string :title
      t.text :body
      t.boolean :published
      t.integer :age
      t.timestamp :published_at
      t.custom "json_field JSON"
      t.custom "uuid_field UUID"

      t.timestamps
    end

    rename_table :post, :posts

    create_index :posts, :user_id

    add_column :posts, :test, :string
    rename_column :posts, :test, :new_test
    remove_column :posts, :new_test

    execute <<-SQL
      ALTER TABLE posts ADD CONSTRAINT posts-user_id-fk FOREIGN KEY (user_id) REFERENCES users(id)
    SQL
  end

  def down
    drop_table :posts
    drop_index :posts, :user_id
  end
end
```

To execute:
```
require "granite/adapter/pg"

Granite::Connections << Granite::Adapter::Pg.new(name: "pg", url: "postgres://url")

create_post = CreatePost.new
puts create_post.generate_sql(:up)
puts create_post.generate_sql(:down)
```


src/granite/table.cr
```diff
@@ -51,12 +51,12 @@ module Granite::Tables
   end
 
   macro table(name)
-    @[Granite::Table(name: {{(name.is_a?(StringLiteral) ? name : name.stringify) || nil}})]
+    @[Granite::Table(name: {{(name.is_a?(StringLiteral) ? name : name.id.stringify) || nil}})]
```
drujensen, line 54: @Blacksmoke16 by adding the `.id`, this adds support for symbols so you can do `table :posts`.

Blacksmoke16:
Given the result of https://github.com/amberframework/granite/issues/182, is this something we actually want to include?

If so, I think I would rather remove, or at least deprecate, the current internal migrator so we don't end up with just another way to do the same thing, as this would be the third way of doing migrations.

drujensen:
@Blacksmoke16  Over the years, I have collected feedback on why developers prefer other ORM's over granite and the lack of a DSL for DDL rises to the top.  

This is not a full featured DSL yet.  There are several options that still need to be built into it.  Also, there are some missing types yet to be added.

Currently this is a self contained class with little changes to existing code.  I could see merging or having `Migrator` use `Migration` to avoid duplicate code.

Let me know your thoughts.  I can close this if you think its not worth the effort.

Blacksmoke16:
@drujensen Personally I'm fine with raw SQL migrations but if this is a commonly requested feature I don't have a problem with the addition.  That is, assuming you/someone will be maintaining it.

drujensen:
@Blacksmoke16 @robacarp thanks for spending the time looking at this.  

@robacarp I agree about the `default` and there are 4 other options that rails supports.  I will look at adding them.

Regarding the specific data types per db, I agree with this as well.  I will look into refactoring out the `Fields` class to be a base class and inherit for each database.  That makes more sense.

spec/granite/migration/migration_spec.cr
```diff
@@ -0,0 +1,147 @@
+require "../../spec_helper"
+{% begin %}
+  {% adapter_literal = env("CURRENT_ADAPTER").id %}
+  class AllFields < Granite::Migration
+    connection :{{ adapter_literal }}
+
+    def up
+      create_table :posts do |t|
+        t.primary :id
+        t.serial :serial1
```
eliasjpr, line 10: I really like the dsl, how can we specify other column parameters?

drujensen:
Thinking about this more, this should be included in the micrate project instead of here.  Closing this PR.

PR #414, opened 2020-08-12
Added a way to disable validation on save method

D1ceWard:
This PR add an named argument to `#save` method which allow to disable model validation, example:

```crystal
class Example < Granite::Base
  ...
  column name : String?
  
  validate :name, "Name cannot be blank" do |parent|
    !parent.name.to_s.blank?
  end
  ...
end

new_instance = Example.new
new_instance.name = ""
new_instance.save(validate: false)
new_instance.persisted? #=> true

another_instance = Example.new
another_instance.name = ""
another_instance.save(validate: true)
another_instance.persisted? #=> false
```

No breaking changes and works as well with `#save!`

### DIFF
--- a/src/granite/transactions.cr
+++ b/src/granite/transactions.cr
@@ -169,12 +169,12 @@ module Granite::Transactions
   # The save method will check to see if the primary exists yet. If it does it
   # will call the update method, otherwise it will call the create method.
   # This will update the timestamps appropriately.
-  def save
+  def save(**args)
     {% begin %}
     {% primary_key = @type.instance_vars.find { |ivar| (ann = ivar.annotation(Granite::Column)) && ann[:primary] } %}
     {% raise raise "A primary key must be defined for #{@type.name}." unless primary_key %}
     {% ann = primary_key.annotation(Granite::Column) %}
-    return false unless valid?
+    return false if args.fetch("validate", true) && !valid?
 
     begin
       __before_save
@@ -199,8 +199,8 @@ module Granite::Transactions
   {% end %}
   end
 
-  def save!
-    save || raise Granite::RecordNotSaved.new(self.class.name, self)
+  def save!(**args)
+    save(**args) || raise Granite::RecordNotSaved.new(self.class.name, self)
   end
 
   def update(**args)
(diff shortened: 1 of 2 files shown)


Blacksmoke16:
What's the use case for this?

robacarp:
This pattern is common with ActiveRecord and it's _fine_, but it reveals a misalignment of the way validations are currently implemented with the real state of the way applications are built. It is often true that complex models end up with several states of validity: a value can be blank except when some other value is true; or validations need only be applied when a final state is reached; etc. 

But I think this change is a relevant way to extend the current model so that it can be used in the various circumstances encountered with building applications.

src/granite/transactions.cr
```diff
@@ -169,12 +169,12 @@ module Granite::Transactions
   # The save method will check to see if the primary exists yet. If it does it
   # will call the update method, otherwise it will call the create method.
   # This will update the timestamps appropriately.
-  def save
+  def save(**args)
```
Blacksmoke16, line 172:
```suggestion
  def save(*, validate : Bool = true)
```

src/granite/transactions.cr
```diff
@@ -169,12 +169,12 @@ module Granite::Transactions
...
     {% begin %}
     {% primary_key = @type.instance_vars.find { |ivar| (ann = ivar.annotation(Granite::Column)) && ann[:primary] } %}
     {% raise raise "A primary key must be defined for #{@type.name}." unless primary_key %}
     {% ann = primary_key.annotation(Granite::Column) %}
-    return false unless valid?
+    return false if args.fetch("validate", true) && !valid?
```
Blacksmoke16, line 177:
```suggestion
    return false if validate && !valid?
```

src/granite/transactions.cr
```diff
@@ -199,8 +199,8 @@ module Granite::Transactions
   {% end %}
   end
 
-  def save!
-    save || raise Granite::RecordNotSaved.new(self.class.name, self)
+  def save!(**args)
+    save(**args) || raise Granite::RecordNotSaved.new(self.class.name, self)
```
Blacksmoke16, line 203:
```suggestion
  def save!(*, validate : Bool = true)
    save(validate: validate) || raise Granite::RecordNotSaved.new(self.class.name, self)
```

eliasjpr (approved):

PR #420, opened 2020-09-01
Fix #416: Examples given in documentation do not compile

### LINKED ISSUE #416: Examples given in documentation do not compile
In [the relationships section](https://github.com/amberframework/granite/blob/master/docs/relationships.md), the `find` method is used instead of `find!`, but nil is not checked. This causes a compilation error.

yujiri8:


eliasjpr (approved):

yujiri8:
Can this be merged? Two people have approved it.

PR #429, opened 2020-09-26
Migrator: create columns with NOT NULL as appropriate

yujiri8:
The built-in Model.migrator.create_sql will create columns
with NOT NULL unless the types are in fact nullable.


drujensen (approved):

drujensen:
@Blacksmoke16 any concerns with this change?

Blacksmoke16:
Some specs would be nice.

yujiri8:
Updated with a spec.

Issue #432, opened 2020-11-11
Migrator doesn't support Enums


wonderix:
*Problem*

The following code 

```crystal
Granite::Connections << Granite::Adapter::Sqlite.new(name: "default",url: "sqlite3::memory:")

enum LogLevel
  Error
  Warning
  Info
  Debug
end 

class LogMessage < Granite::Base
  connection default
  column id : Int64, primary: true
  column level : LogLevel = LogLevel::Info, converter: Granite::Converters::Enum(LogLevel, String)
end

LogMessage.migrator.create

```

Produces

```
Unhandled exception: Migrator(Granite::Adapter::Sqlite) doesn't support 'LogLevel' yet. (Exception)
  from lib/granite/src/granite/migrator.cr:44:50 in '->'
  from ../../../../usr/local/Cellar/crystal/0.35.1_1/src/primitives.cr:255:3 in 'create_sql'
  from lib/granite/src/granite/migrator.cr:86:7 in 'create'
```

Blacksmoke16:
The migrator is more of an internal thing used to testing Granite itself.  You'd be better off using https://github.com/amberframework/micrate or something similar.

Refs:
* https://github.com/amberframework/granite/issues/421#issuecomment-686085633
* https://github.com/amberframework/granite/issues/182
* https://github.com/amberframework/granite/pull/413

wonderix:
@Blacksmoke16, thanks for the fast response. For clarification I created a PR #433 to change the documentation accordingly.

PR #433, opened 2020-11-12
Change documentation to prefer micrate

wonderix:
According to #432, migrator should only be used for internal testing. Therefore, I changed the documentation accordingly.

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -1,24 +1,5 @@
 # Migrations
 
-## migrator
-
-Provides `drop`, `create` and `drop_and_create` methods
-
-   ```crystal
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
I'd honestly just not even mention there actually is a migrator and just point to micrate.

Thoughts @drujensen / @eliasjpr ?

watzon:
If the built in migrator isn't meant to be used then I'd tend to agree. Mentioning it at all is just confusing.

wonderix:
I removed the `migrator` section completely.

Blacksmoke16 (approved):

robacarp (approved):

PR #438, opened 2021-01-19
add support for foreign key converters

repomaa:
mainly useful for defining `belongs_to` relations on uuid foreign keys


hugopl:
Maybe would be nice to set the default `foreing_key` type... I'm having problems because the database I'm accessing uses Int32 as primary_keys and foreing_keys. As Granite has `Int64` hardcoded in their macro I did the ugliest workaround ever in the model:

```Crystal
class WhateverMode < Granite::Base
  alias Int64 = Int32
  belongs_to foo
end
```

Luck I'm not using an Int64 type anywhere in the model. :grin:

crimson-knight (approved):

PR #442, opened 2021-02-12
Add types to `abstract Granite::Adapter::Base` abstract methods

ashishbista:
In Crystal 0.36.0, abstract `def` implementations must honor abstract declaration regarding type restrictions.
Crystal 0.36.0 release documentation: https://crystal-lang.org/2021/01/26/crystal-0.36.0-released.html

And, update crystal docker image to 0.36.1


ashishbista:
Just noticed amberframework/granite/pull/437 is already doing the same thing. Closing it.

PR #445, opened 2021-05-11
Enable non snake-case naming in Postgres

Gembot-ai-admin:
We are integrating Crystal with other services written in NodeJS.

Since the Node project is responsible for managing the schema all the Table names are CamelCaps and the field names are camelCase.

This is at odds with the ruby/crystal world of snake_case, so this PR aims to enable users of Granite the option at least to use alternative naming conventions within their db schema, to allow interoperability with other code-bases.


crimson-knight:
Can anyone add some tests for this?

PR #450, opened 2021-05-27
crystal support for 1.0.0

dare892:


shard.yml
```diff
@@ -1,8 +1,8 @@
 name: granite
 
-version: 0.22.0
+version: 1.0.0
 
-crystal: 0.35.0
+crystal: 1.0.0
```
robacarp, line 5: Does the version specifier need to be more specific here? The semantics of the crystal version comparison [aren't quite what most people expect](https://github.com/crystal-lang/shards/blob/master/docs/shard.yml.adoc)

shard.yml
```diff
@@ -1,8 +1,8 @@
 name: granite
 
-version: 0.22.0
+version: 1.0.0
```
Blacksmoke16, line 3: I would ensure the Amber team actually wants this to be a `1.0.0` release.

dare892: @elorest ?

shard.yml
dare892: updated, thanks

shard.yml
robacarp: Good question @Blacksmoke16. I would say no, I don't think that's warranted. A 0.23 is what I'd expect.

Blacksmoke16:
@elorest I hope you're making another PR before cutting a release. Otherwise you're going to end up with Granite 1.0.0.

Blacksmoke16:
NVM, I see you fixed it.

Issue #463, opened 2022-04-20
Doesn't work in Crystal 1.4.0


confact:
Tried to upgrade to Crystal 1.4.0

I get this error when running crystal spec:
```
$ crystal spec
Showing last frame. Use --error-trace for full trace.
There was a problem expanding macro 'column'
Code in macro 'timestamps'
 2 | column updated_at : Time?
     ^
Called macro defined in lib/granite/src/granite/columns.cr:63:3
 63 | macro column(decl, **options)
Which expanded to:
 > 14 |     
 > 15 | 
 > 16 |     @[Granite::Column(column_type: nil, converter: nil, auto: false, primary: false, nilable: true)]
            ^
Error: can't annotate @updated_at in Collaboration because it was first defined in Granite::Base
```
Moving down to 1.3.2 makes it work again.

VanyaNeytrino:
I found a way out: comment out the line `column update_at : Time?` in https://github.com/amberframework/granite/blob/master/src/granite/columns.cr#L108 to avoid duplicating `updated_at`

VanyaNeytrino:
My PR https://github.com/amberframework/granite/pull/464

PR #464, opened 2022-05-14
fix Doesn't work in Crystal 1.4.0

VanyaNeytrino:
trying to solve the issue https://github.com/amberframework/granite/issues/463 if you know a better way, I will be glad to hear and make corrections.

### DIFF
--- a/src/granite/columns.cr
+++ b/src/granite/columns.cr
@@ -105,7 +105,6 @@ module Granite::Columns
   # include created_at and updated_at that will automatically be updated
   macro timestamps
     column created_at : Time?
-    column updated_at : Time?
   end
 
   def to_h


Blacksmoke16:
I'm pretty sure all you need to do is upgrade to Crystal `1.4.1`. This was a regression that was fixed in https://github.com/crystal-lang/crystal/pull/11995.

VanyaNeytrino:
> I'm pretty sure all you need to do is upgrade to Crystal `1.4.1`. This was a regression that was fixed in [crystal-lang/crystal#11995](https://github.com/crystal-lang/crystal/pull/11995).
unfortunately this is not the case
Screenshot 2022-05-15 at 12 05 13

Blacksmoke16:
Oh wow okay, I guess it's been broken since `1.0.0` then :grimacing:. In that case, I think a better fix would be replacing https://github.com/amberframework/granite/blob/master/src/granite/transactions.cr#L264 with a call to `set_timestamps`, much like it does in https://github.com/amberframework/granite/blob/master/src/granite/transactions.cr#L141 and that should fix it, and be backwards compatible.

confact:
Tested master with 1.4.1 and #465 seemed to have fixed this issue.

robacarp:
@VanyaNeytrino I think this is fixed elsewhere so I'm going to close. Feel free to re-open if there's other stuff in here that's relevant.

PR #470, opened 2022-11-27
add #next_column_index to db mocker

pan:
This is required by changes made in crystal-db v0.11.0


pan:
Without this change it will run into this compiling error.
```
$ crystal spec
Showing last frame. Use --error-trace for full trace.

In spec/mocks/db_mock.cr:47:1

 47 | class FieldEmitter < DB::ResultSet
      ^
Error: abstract `def DB::ResultSet#next_column_index()` must be implemented by FieldEmitter
```

robacarp (approved):

PR #474, opened 2023-04-28
Update spec tests and add ghaction

kalinon:
* update docker to use crystal 1.8 and added github action
* fix workflow if statements
* split workflow into separate jobs for each database
* workflow: fix missing envs
* workflow: add psql db var
* attempt to use workflow services
* split tests by database to speed things up

Sqlite and MySQL specs are passing. PSQL still is having issues.

PSQL: `284 examples, 1 failures, 3 errors, 0 pending`


.github/workflows/spec.yml
```diff
@@ -0,0 +1,121 @@
+name: spec
+on:
+  push:
+  pull_request:
+    branches: [main, master]
+  schedule:
+    - cron: "0 6 * * 6" # Every Saturday 6 AM
```
crimson-knight, line 7: I think we can remove this, no real need for regularly scheduled CI-runs since it'll run frequently while under active development

.env
```diff
@@ -2,7 +2,7 @@ PG_DATABASE_URL=postgres://granite:password@localhost:5432/granite_db
 MYSQL_DATABASE_URL=mysql://granite:password@localhost:3306/granite_db
 SQLITE_DATABASE_URL=sqlite3:./granite.db
 CURRENT_ADAPTER=pg
-PG_VERSION=10.4
+PG_VERSION=15.2
 MYSQL_VERSION=5.7
```
crimson-knight, line 6: Probably should change this to use v8 instead of v5.7 since 5.7 is EOL

spec/spec_helper.cr
```diff
@@ -6,12 +6,32 @@ Granite::Connections << Granite::Adapter::Mysql.new(name: "mysql", url: ENV["MYS
...
-
 require "spec"
 require "../src/granite"
 require "../src/adapter/**"
 require "./spec_models"
 require "./mocks/**"
+
+Spec.before_suite do
+  Granite.settings.default_timezone = Granite::TIME_ZONE
+  {% if flag?(:spec_logs) %}
```
crimson-knight, line 17: That's an interesting way to get more detailed logging without having to change much in the specs themselves

spec/spec_helper.cr
```diff
@@ -6,12 +6,32 @@ Granite::Connections << Granite::Adapter::Mysql.new(name: "mysql", url: ENV["MYS
...
+      source: "*",
+      level: ::Log::Severity::Trace,
+      backend: ::Log::IOBackend.new(STDOUT, dispatcher: :sync),
+    )
+  {% end %}
+end
+
+Spec.before_each do
+  # I have no idea why this is needed, but it is.
+  Granite.settings.default_timezone = Granite::TIME_ZONE
```
crimson-knight, line 29: I wonder if this is something we should detect at run-time and use the timezone of the server by default instead of having to specify it.

I'm not sure, just an idea and noting it here in case anyone else has an opinion/thought

crimson-knight (approved):

.env
kalinon: agreed, i did run it against 8, but there's a bug/issue with connecting to 8 with the adapter. 
https://github.com/crystal-lang/crystal-mysql/issues/62

I suggest we handle moving to 8 for testing in a separate issue, as it will take me a bit to work around it.

spec/spec_helper.cr
kalinon: agreed

spec/granite/query/assemblers/mysql_spec.cr
```diff
@@ -48,13 +48,13 @@ require "../spec_helper"
       end
 
       it "property defines IN query" do
-        sql = "SELECT #{query_fields} FROM table WHERE date_completed IS NULL AND status IN ('outstanding','in_progress') ORDER BY id DESC"
+        sql = "SELECT #{query_fields} FROM table WHERE date_completed IS NULL AND status IN (?,?) ORDER BY id DESC"
```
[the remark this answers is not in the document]
kalinon: Yes, this was an change i made to the underlying lib. Previously granite was attempting to escape the string itself by wrapping it in single quotes, but it broke with strings that contained them such as "joe's burger". Now strings are passed down to the underlying adapter for variable substitution.

Only `Bool` and `Number` are substituted by granite.

shard.yml
```diff
@@ -19,15 +19,12 @@ development_dependencies:
...
   pg:
     github: will/crystal-pg
     version: ~> 0.26.0
-
-  ameba:
-    github: crystal-ameba/ameba
-    version: ~> 1.3.1
+  # ameba:
+  #   github: crystal-ameba/ameba
+  #   version: ~> 1.3.1
```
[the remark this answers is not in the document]
kalinon: ah, these were disabled to speed up the workflow when i was testing. we can re-enable it if its wanted. the `spec/run_all_specs.sh` uses it

spec/spec_helper.cr
crimson-knight: I'm going to say this is a sane default and let's move forward from here. However, it would be wise to surface some docs on this and maybe make configuring it easier (ie make it more obvious how to set it).

Issue #475, opened 2023-05-03
Feature: Automatic Connection Switching & Horizontal Sharding


crimson-knight:
This is pretty important, especially since most large Rails apps have moved to use the automatic connection switching between readers/writers, and probably also for sharding.

It is currently possible to have multiple connections like this:
```crystal
Granite::Connections << Granite::Adapter::Mysql.new(name: "mysql_primary", url: "YOUR_PRIMARY_DATABASE_URL")
Granite::Connections << Granite::Adapter::Mysql.new(name: "mysql_reader", url: "YOUR_READER_DATABASE_URL")
```

However, this does _not_ set the actual roles for the db's shown, and you can only select a single database connection from the model.

I think we'll need to modify the `connection` macro to allow accepting a `NamedTuple` or `Tuple` so we can pass in some configuration settings like:

```crystal
{
  writer: "mysql_primary",
  reader: "mysql_reader"
}
```

The current connection macro is here: https://github.com/amberframework/granite/blob/3f7a41510f91ca58e8995bb1c064a708cab2cd4c/src/granite/table.cr#L59
```crystal
macro connection(connection_details)
  {% if connection_details.is_a?(StringLiteral) %}
  class_getter adapter : Granite::Adapter::Base = Granite::Connections[{{connection_details}}]
  {% elsif connection_details.is_a?(TupleLiteral) || connection_details.is_a?(NamedTupleLiteral) %}
  # Put the code here to set the reader/writer
  {% else %}
  # Probably don't actually want to raise here like this
  raise "No registered connection with the name '{{name.id}}'"
  {% end %}
end
```

Just throwing this out there for now, but I think it'll be easier than it may first appear to add automatic connection switching.

We would also need:
- A variable to track when the last `write` transaction was done to get the time since the last write
- A configuration to set the delay from switching between the writer and the reader, probably 2 seconds to allow for propagation
- A way to tell if the action being performed is a read or write action. This is pretty straightforward since writing actions are `insert`, `update` & `upsert` and most other queries will be to read. We want to default to using the reader as often as possible.

PR #476, opened 2023-05-05
Feature: Read replica support

a-alhusaini:
This is the first step in handling #475. The next step is to create a mechanism that lets the adapter know about the reader/writer if they exist and decide which one to run based on the information provided to it.

I however, deem this a good milestone for a PR. This change is large and handling merge conflicts would be a pain. Better to PR it in step by step.

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,6 @@
 PG_DATABASE_URL=postgres://granite:password@localhost:5432/granite_db
 MYSQL_DATABASE_URL=mysql://granite:password@localhost:3306/granite_db
+MYSQL_REPLICA_URL=mysql://granite:password@localhost:3306/granite_replica_db
 SQLITE_DATABASE_URL=sqlite3:./granite.db
 SQLITE_REPLICA_URL=sqlite3:./granite_replica.db
 CURRENT_ADAPTER=pg
--- a/.env
+++ b/.env
@@ -1,6 +1,7 @@
 PG_DATABASE_URL=postgres://granite:password@localhost:5432/granite_db
 MYSQL_DATABASE_URL=mysql://granite:password@localhost:3306/granite_db
 SQLITE_DATABASE_URL=sqlite3:./granite.db
+SQLITE_REPLICA_URL=sqlite3:./granite_replica.db
 CURRENT_ADAPTER=pg
 PG_VERSION=15.2
 MYSQL_VERSION=5.7
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,4 @@ shard.lock
 
 # Ignore bin because they will be build with shards install
 bin
-.env_kalinon
+.env
--- a/src/granite/base.cr
+++ b/src/granite/base.cr
@@ -39,7 +39,7 @@ abstract class Granite::Base
   extend Integrators
   extend Select
 
-  @@last_write_time = Time.utc
+  @@last_write_time = Time.monotonic
 
   def self.last_write_time
     @@last_write_time
@@ -47,7 +47,7 @@ abstract class Granite::Base
 
   # This is done this way because callbacks don't work on class mthods
   def self.update_last_write_time
-    @@last_write_time = Time.utc
+    @@last_write_time = Time.monotonic
   end
 
   def update_last_write_time
@@ -55,7 +55,7 @@ abstract class Granite::Base
   end
 
   def self.time_since_last_write
-    Time.utc - @@last_write_time
+    Time.monotonic - @@last_write_time
   end
 
   def time_since_last_write
--- a/src/granite/base.cr
+++ b/src/granite/base.cr
@@ -68,6 +68,10 @@ abstract class Granite::Base
     end
   end
 
+  def switch_to_reader_adapter
+    self.class.switch_to_reader_adapter
+  end
+
   def self.switch_to_writer_adapter
     @@adapter = @@writer_adapter
   end
@@ -79,7 +83,7 @@ abstract class Granite::Base
   def self.schedule_adapter_switch
     spawn do
       sleep 2.seconds
-      self.class.switch_to_reader_adapter
+      switch_to_reader_adapter
     end
   end
 
--- a/src/granite/base.cr
+++ b/src/granite/base.cr
@@ -62,6 +62,16 @@ abstract class Granite::Base
     self.class.time_since_last_write
   end
 
+  def self.switch_to_reader_adapter
+    if time_since_last_write > 2.seconds
+      @@adapter = @@reader_adapter
+    end
+  end
+
+  def self.switch_to_writer_adapter
+    @@adapter = @@writer_adapter
+  end
+
   macro inherited
     protected class_getter select_container : Container = Container.new(table_name: table_name, fields: fields)
 
--- a/src/granite/base.cr
+++ b/src/granite/base.cr
@@ -39,6 +39,29 @@ abstract class Granite::Base
   extend Integrators
   extend Select
 
+  @@last_write_time = Time.utc
+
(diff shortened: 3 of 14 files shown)


src/granite/base.cr
```diff
@@ -70,5 +93,10 @@ abstract class Granite::Base
 
     disable_granite_docs? def initialize
     end
+
```
crimson-knight, line 96: Just a side note, I wonder if there is a way to ensure that these are ALWAYS performed first, that way if someone writes a callback that fails, we don't miss updating the last write time

crimson-knight: We can worry about this once it's fully fleshed out, I bet there will be other aspects that crop up that might resolve this on it's own

crimson-knight:
Great job so far :)

Let's keep the progress going!

src/granite/base.cr
```diff
@@ -39,6 +39,29 @@ abstract class Granite::Base
   extend Integrators
   extend Select
 
+  @@last_write_time = Time.utc
```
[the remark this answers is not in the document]
a-alhusaini: I thought about that. It can wait because its only costing us 1*number_of_models of queries when the app is launched. It seems like a such a tiny inefficiency. 

I'd say its low priority.

.gitignore
```diff
@@ -12,3 +12,4 @@ shard.lock
 
 # Ignore bin because they will be build with shards install
 bin
+.env_kalinon
```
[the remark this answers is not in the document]
a-alhusaini: @kalinon ?

.env
```diff
@@ -1,6 +1,7 @@
 PG_DATABASE_URL=postgres://granite:password@localhost:5432/granite_db
+PG_REPLICA_URL=postgres://granite:password@localhost:5432/granite_db
```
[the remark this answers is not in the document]
a-alhusaini: In my opinion, this is out of scope for this PR. It is definitely a good idea though.

src/granite/querying.cr
```diff
@@ -28,13 +28,14 @@ module Granite::Querying
...
-  def all(clause = "", params = [] of Granite::Columns::Type)
+  def all(clause = "", params = [] of Granite::Columns::Type, use_primary_adapter = true)
+    switch_to_writer_adapter if use_primary_adapter == true
     Collection(self).new(->{ raw_all(clause, params) })
   end
 
   # First adds a `LIMIT 1` clause to the query and returns the first result
   def first(clause = "", params = [] of Granite::Columns::Type)
-    all([clause.strip, "LIMIT 1"].join(" "), params).first?
+    all([clause.strip, "LIMIT 1"].join(" "), params, false).first?
```
crimson-knight, line 38: This should be `true` when performing this read action on the same model that recently had a `write` action performed. I don't think I've seen anything for that logic yet

src/granite/querying.cr
```diff
@@ -86,7 +87,7 @@ module Granite::Querying
     end
 
     loop do
-      results = all "#{clause} LIMIT ? OFFSET ?", params + [limit, offset]
+      results = all "#{clause} LIMIT ? OFFSET ?", params + [limit, offset], false
```
crimson-knight, line 90: This is another spot where it should be `true` if this is a model that had a `write` action performed within the set period of time.

src/granite/querying.cr
```diff
@@ -28,13 +28,14 @@ module Granite::Querying
   # that you are using so you are not restricted or dummied down to support a
   # DSL.
   # Lazy load prevent running unnecessary queries from unused variables.
-  def all(clause = "", params = [] of Granite::Columns::Type)
+  def all(clause = "", params = [] of Granite::Columns::Type, use_primary_adapter = true)
```
crimson-knight, line 31:
```suggestion
  def all(clause = "", params = [] of Granite::Columns::Type, has_performed_recent_write_action = true)
```

src/granite/querying.cr
```diff
@@ -28,13 +28,14 @@ module Granite::Querying
...
+    switch_to_writer_adapter if use_primary_adapter == true
```
crimson-knight, line 32:
```suggestion
    switch_to_writer_adapter if has_performed_recent_write_action
```

src/granite/querying.cr
```diff
@@ -115,19 +116,22 @@ module Granite::Querying
   end
 
   def exec(clause = "")
+    switch_to_writer_adapter
```
crimson-knight, line 119: This looks like a good placeholder to put the trigger to check if we should be switching the connection. I see it repeated a bit, and I don't think we should assume switching to the writer so easily.

crimson-knight (changes requested):

src/granite/querying.cr
a-alhusaini: The switch only happens if > 2 seconds have passed since last write operation.

src/granite/querying.cr
a-alhusaini: See previous response.

src/granite/querying.cr
a-alhusaini: If the user calls `exec` directly, I don't think they want to perform a write operation.

src/granite/querying.cr
a-alhusaini: The logic for handling switching is in Granite::Base.

We use the primary database connection if the user calls this function directly. Any other functions that call this function actually use the reader adapter. The system schedules a switch to the reader adapter after any write operations. So in those cases, we don't need to explicitly set the adapter to the writer adapter because the system will only change the adapter 2 seconds after a write operation. We use whatever operator is available. Before 2 seconds we use writer adapter. After 2 seconds it automatically uses reader adapter. The use_primary = true is just to make sure we use the writer adapter when the user calls this method directly because that is what the user would expect.

src/granite/table.cr
```diff
@@ -11,7 +11,7 @@ end
 module Granite::Tables
   module ClassMethods
     def adapter : Granite::Adapter::Base
-      Granite::Connections.registered_connections.first? || raise "No connections have been registered."
+      Granite::Connections.registered_connections.first?.not_nil![:writer] || raise "No connections have been registered."
```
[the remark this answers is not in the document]
a-alhusaini: There are a few places where we resort to accessing a part of the tuple. I haven't managed to think of an abstraction that satisfies me for this piece of code and others like it. All suggestions are welcome!

src/granite/querying.cr
kalinon: agreed

src/granite/querying.cr
a-alhusaini: The idea behind this call (and others like it) is to default to using the primary adapter if the user is calling the function (presumably, if a user wants to write raw sql then defaulting to the writer adapter is a good default). If the user wants to use the reader adapter he has that option.

It is also passed in by other granite methods that call this one. Since most calls to all are from find operations those calls set the variable to false. Which makes granite use whatever the current adapter is. If it happens to be the writer adapter it will use that. But if a write operation has occurred and the time since that operation is sufficient, a callback in granite/base.cr would have switched to the reader adapter.

src/granite/querying.cr
a-alhusaini: This is addressed in the response to a similar change request.

src/granite/table.cr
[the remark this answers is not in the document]
a-alhusaini: This was solved by adding a first_reader/first_writer class to granite::connections.

That is really the only reason to access registered_connections directly. To get the first connection for when we need a fallback. That has been done successfully now.

crimson-knight (approved):

PR #482, opened 2023-08-08
Fixed error caused by db version upgrade

a-alhusaini:


a-alhusaini:
I tried running circleci myself but the errors I'm running into are caused due to environment variables not being created/loaded before the specs start.

crimson-knight (approved):
Thank you! I'll get CircleCi working from here

Issue #481, opened 2023-06-12
Import data maintaining timestamps


ggtrigg:
I'd like to be able to import data from another database and keep the created_at & updated_at times. From everythig I've tried it seems like granite always overrides these on create even if those values are explicitly provided.

crimson-knight:
You should be able to do bulk imports using `import` interface and not have it override the timestamps
https://github.com/amberframework/granite/blob/master/docs/imports.md

Did this also override your timestamps?

ggtrigg:
The `import` interface does override any supplied timestamps. I got around it by making a local modification so that the import only sets the timestamps unless both `created_at` & `updated_at` are supplied. vis.
```
model.set_timestamps if (model.updated_at.nil? || model.created_at.nil?)
```
at https://github.com/amberframework/granite/blob/59801deb1a503bfb3d047fdbcea11431690b6d2d/src/adapter/mysql.cr#L68

If it helps I could raise a pull request with this change (and also for the other db adapters)?

crimson-knight:
@ggtrigg this is resolved with the new `skip_timestamps` option that can be passed in

PR #486, opened 2023-09-27
Add `skip_timestamps` option

### LINKED ISSUE #318: .update() without changing timestamps
This is occasionally useful, it would be great to have a flag to disable callbacks and such.

stufro:
Resolves #318

This adds the ability to pass `skip_timestamps: true` to `update` and `update!` methods. To do this I had to change the `save` method to accept `skip_timestamps` so to avoid inconsistent behaviour I have also added the option to the `create` and `create!` methods.

To achieve this I had to remove the type definition of `args` on the method definition so that either a `Granite::ModelArgs` or a `NamedTuple` can be passed in with `skip_timestamps`, I originally added another method overload for this case as you can see in the first commit e6f917e4e383d69672a6acb6c5f39c299ef58d13. Open to thoughts on this.

I have also updated the crud.md docs to describe this feature and added a missing example for `save(validate: false)`

### DIFF
--- a/docker/docker-compose.sqlite.yml
+++ b/docker/docker-compose.sqlite.yml
@@ -1,7 +1,8 @@
 version: '2'
 services:
   spec:
-    command: 'bash -c "cd /app/user && bin/ameba && crystal tool format --check && sqlite3 --version && crystal spec"'
+    #command: 'bash -c "cd /app/user && bin/ameba && crystal tool format --check && sqlite3 --version && crystal spec"'
+    command: 'bash -c "cd /app/user && crystal spec"'
     extends:
       file: ../docker-compose.yml
       service: spec
--- a/docker/docker-compose.sqlite.yml
+++ b/docker/docker-compose.sqlite.yml
@@ -1,8 +1,7 @@
 version: '2'
 services:
   spec:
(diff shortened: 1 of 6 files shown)


robacarp:
Huh. I have wanted #update with skip_timestamps but never #create. Typically I want that because I want to "soft touch" some record without altering the breadcrumb trail for some reason...which I can't think of why I'd want to do that on a create. 

I think the timestamp columns are specified as NOT NULL (or at least they are in my databases). Wouldn't that cause a problem?

stufro:
Hmm yeah you're right. I hadn't considered DB level constraints.

I've removed it from create and also squashed some of the commits to tidy up the branch.

crimson-knight:
@stufro I think that you should keep the option to create records while skipping timestamps if the created_at and updated_at values are already present. That's the current gap, creating records in bulk that already have timestamps that want to be preserved but they aren't at the time of creation.

Database constraints will error on their own for the user, so I wouldn't worry about policing that from inside of Granite. I've had instances where we did imports and ignored the existence of the timestamps during creation but allowed them to be updated if/when the record was updated.

stufro:
@crimson-knight ah yes I suppose that might also help with resolving #481?

I have made the change.

crimson-knight (approved):

PR #487, opened 2023-10-16
refactor: fix linting issues on not_nil! usage

jphaward:
Currently ameba linting fails due to the usage of not_nil!. This can be fixed by giving the not_nil! an argument of an error message however I've chosen to prioritise readability.


crimson-knight:
@jphaward looks like CI failed, but just around PG. Can you rebase and check if this is just from CI?

jphaward:
> @jphaward looks like CI failed, but just around PG. Can you rebase and check if this is just from CI?

@crimson-knight mysql failed due to recently merged in timestamps changed. I've fixed it in a separate PR: #492

crimson-knight (approved):

PR #489, opened 2023-10-16
Close result set at the end of test in passthrough_spec.cr

stufro:
This fixes an issue when running against postgres `Expected PQ::Frame::ParseComplete but got PQ::Frame::ReadyForQuery`

The offending spec was: `spec/granite/querying/passthrough_spec.cr` Anything that ran after this spec would throw the above error.

Simply closing the result set before moving on fixed the errors.

Co-authored-by: @jphaward <jphaward@gmail.com>


crimson-knight (approved):
Thank you!

PR #490, opened 2023-10-16
ci: add ameba linting to formatting ci job

jphaward:
Currently ameba is not in the pipeline however is used when running the tests within Docker (command in the readme). It would be useful for these to be consistent.

I cannot test the pipeline changes locally so it may need tweaking once the workflow has been approved.


crimson-knight (approved):
LGTM, it failed on CI now that is does run Ameba. Which is fine, we can make another PR that resolves those

PR #491, opened 2023-10-16
Add reload method

### LINKED ISSUE #485: Add `reload` method
This is a feature request to add a `reload`, similar to what's currently implemented in [ActiveRecord](https://devdocs.io/rails~7.0/activerecord/persistence#method-i-reload)

This would find the record by its primary key from the DB and raise an error if it doesn't exist.

stufro:
Resolves #485

### DIFF
--- a/src/granite/querying.cr
+++ b/src/granite/querying.cr
@@ -154,8 +154,18 @@ module Granite::Querying
     end
   end
 
-  # reload returns the record with the attributes reloaded from the database
-  def reload
-    self.class.find!(primary_key_value)
-  end
+  {% if parse_type("Spec").resolve? %}
+    # Returns the record with the attributes reloaded from the database.
+    #
+    # **Note:** this method is only defined when the `Spec` module is present.
+    #
+    # ```
+    # post = Post.create(name: "Granite Rocks!", body: "Check this out.")
+    # # record gets updated by another process
+    # post.reload # performs another find to fetch the record again
+    # ```
+    def reload
+      self.class.find!(primary_key_value)
+    end
+  {% end %}
 end
--- a/src/granite/querying.cr
+++ b/src/granite/querying.cr
@@ -154,6 +154,7 @@ module Granite::Querying
     end
   end
 
+  # reload returns the record with the attributes reloaded from the database
   def reload
     self.class.find!(primary_key_value)
   end
--- a/src/granite/querying.cr
+++ b/src/granite/querying.cr
@@ -165,6 +165,9 @@ module Granite::Querying
     # post.reload # performs another find to fetch the record again
     # ```
     def reload
+      {% if !@top_level.has_constant? "Spec" %}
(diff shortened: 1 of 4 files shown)


crimson-knight:
This is basically an alias of `find!`, and the test example shows that it's meant to be used when you modify the record outside of the object. Other than the convenience of using `.reload` instead of `Model.find!()`, what's the benefit of adding this method?

Do you have a use case?

stufro:
We have 2 use cases for it. 

The first is reloading a record in specs:

```crystal
person = Person.create
PersonUpdater.run
person.reload.field.should eq "new value"
```

The second is in a distributed system when running more than 1 instance of the same service, you might want to reload the record in case another instance has updated the db record since it was loaded into memory. 

So yeah I agree it is just a convenience method, happy to go with your judgment 😊

a-alhusaini:
Might I suggest renaming it to `sync` instead of reload?

Or maybe `resync`.. since you're syncronizing with the database.. `reload` is an ok term but it is more imperative than declarative.. 

I think resync/sync are better names but I think we can come up with something better if we thought about it.

crimson-knight:
I think this would be helpful in the test suite, so let's wrap this in a macro that checks for the `Spec` module before it's included.

Can you also add the documentation to the code comments? Now that we have the crystaldoc.info I want to ensure we're writing docs that the community at large can consume in as many places as possible.

stufro:
Sure thing all done. 1 questions about the docs, when running `crystal docs` the reload method doesn't appear I'm assuming because the condition in the macro isn't met. Do you know a good way to deal with this?

src/granite/querying.cr
```diff
@@ -2,153 +2,170 @@ module Granite::Querying
...
+    # Returns the record with the attributes reloaded from the database.
+    #
+    # **Note:** this method is only defined when the `Spec` module is present.
+    #
+    # ```
+    # post = Post.create(name: "Granite Rocks!", body: "Check this out.")
+    # # record gets updated by another process
+    # post.reload # performs another find to fetch the record again
+    # ```
+    def reload
```
crimson-knight, line 167:
```suggestion
    def reload
      {% if !@top_level.has_constant? "Spec" %}
        raise "#reload is a convenience method for testing only, please use #find in your application code"    
      {% end %}
```

src/granite/querying.cr
```diff
@@ -2,153 +2,170 @@ module Granite::Querying
...
       end
 
-      "#{quoted_table_name}.#{quote(name.to_s)} #{matcher}"
+      {clauses.join(" AND "), criteria_hash.values}
     end
-
-    {clauses.join(" AND "), criteria_hash.values}
   end
+
+  {% if parse_type("Spec").resolve? %}
```
crimson-knight, line 157: Just remove this macro that wraps everything in favor of the macro inside of the method itself

crimson-knight (changes requested):

crimson-knight (approved):

PR #495, opened 2023-10-17
Update readme setup instructions

jphaward:
There was a mismatch with how env variables were exported to how the testing readme steps described. I've updated `export.sh` to dynamically export the vars declared in .env rather than having to maintain two files. 

I also updated the steps to show how to use this for both docker and local setups, whilst giving a few options for docker env file usage.


a-alhusaini:
Hmm, I added export.sh by accident.. It was a personal convenience for loading env vars.

Does it make sense to keep?

crimson-knight (approved):
LGTM, thank you!

jphaward:
> Hmm, I added export.sh by accident.. It was a personal convenience for loading env vars.
> 
> Does it make sense to keep?

@a-alhusaini The mentioned `source .env` in the readme to export the env vars did not work for me so I found it useful for setup. I'd vote it's a useful script to have to help people contribute who have different setups. Small changes that reduce the entry barrier to contribution are useful imo

PR #497, opened 2023-10-26
Fixed bug preventing Amber from booting

crimson-knight:
- Changed connection period to use less memory by switching to Int32 from Int64 
- Changed reader/writer to allow `nil` now that we have `nil` checks for when the connections are being used. This resolves the issue with Amber not being able to boot.

### DIFF
--- a/src/granite/connection_management.cr
+++ b/src/granite/connection_management.cr
@@ -4,12 +4,12 @@ module Granite::ConnectionManagement
     # database connection for read operations
     # all models use this value. Change it
     # to change it in all Granite::Base models.
-    class_property connection_switch_wait_period : Int64 = Granite::Connections.connection_switch_wait_period
+    class_property connection_switch_wait_period : Int32 = Granite::Connections.connection_switch_wait_period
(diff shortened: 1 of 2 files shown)


crimson-knight:
I tracked this down using the following steps:

If you care to test, you can go into your lib/granite/src/granite/connections.cr and use:
def self.first_connection
  begin
    first_connection = @@registered_connections.first?
    raise "First registered connection cannot be nil" if first_connection.nil?
  rescue ex
    puts "Any registered connections? #{@@registered_connections.any?}"
  ensure
    first_connection = { name: "pg", url: "some-url" }
  end
  first_connection
end


Then in your config/database.cr

require "granite/adapter/pg"

puts "Loading connections..."
Granite::Connections << Granite::Adapter::Pg.new(name: "pg", url: ENV["DATABASE_URL"]? || 
Amber.settings.database_url)

puts "Done loading connections!
Then when the app boots, you'll see that the puts from the rescue statement fires twice because the config/database.cr file ever fires

After testing on a demo app this appears to have solved the issue.

PR #501, opened 2023-11-16
Add docs for transaction methods

HCLarsen:
Documentation added for some common methods for `Granite::Base`.

### DIFF
--- a/src/granite/transactions.cr
+++ b/src/granite/transactions.cr
@@ -2,14 +2,23 @@ require "./exceptions"
 
 module Granite::Transactions
   module ClassMethods
+    # Removes all records from a table.
     def clear
       adapter.clear table_name
     end
 
+    # Creates a new record, and attempts to save it to the database. Returns the
+    # newly created record.
+    #
+    # **NOTE**: This method still outputs the new object even when it failed to save
+    # to the database. The only way to determine a failure is to check any errors on
+    # the object, or to use `#create!`.
     def create(**args)
       create(args.to_h)
     end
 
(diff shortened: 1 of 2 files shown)


src/granite/transactions.cr
```diff
@@ -196,9 +214,9 @@ module Granite::Transactions
   {% end %}
   end
 
-  # The save method will check to see if the primary key exists yet. If it does
-  # it will call the update method, otherwise it will call the create method.
-  # This will update the timestamps appropriately.
+  # Attempts to save the record to the database, returning `true` if successful,
+  # and `false` if not. If the save is unsuccessful, `#errors` will be populated
+  # with the errors which caused the save to fail.
```
crimson-knight, line 219: We should still include verbiage here that this will perform and `update` or `insert` depending on the presence of the primary key.

I haven't tried this yet, but what happens if someone tries to `save` a record with a PK that hasn't existed yet? Does it try updating and fail?

HCLarsen: If we do add something to state that it either updates the record or inserts a new one, I'd like to say it in a generic way, and not mention any specific method names, in case we change the implementation.

Currently, `#update` calls `#save`, which checks for the PK being in the DB or not, and then either calls private method `#__update` or `#__create`. It's a very roundabout way to do it, so I think it could stand to be cleaned up a bit.

In response to your second question, if you manually set the PK on a new object (`record.id = 1`), and `save` it, it'll create the new entry in the DB with that PK, instead of auto generating one.

PR #502, opened 2024-01-15
Use Crystal type inference to find the correct adapter.

hugopl:


hugopl:
Thinking again... Maybe this PR isn't good because it will make projects depend on all database backend shards. Not sure if it's already the case, but I believe people that just uses postgres don't need to have sqlite shard as dependency.

Blacksmoke16:
There's probably a way you could generate the case at compile time based on if those types exist. But debatable if that's worth the effort/complexity.

hugopl:
I took a look in the code and seems it's the case, Granite is requiring all kind of database client shards to compile, I'll open an issue for that.

PR #503, opened 2024-01-21
Propagate original callstack, save a fiber allocation if unnecessary

Vici37:
I'm working on a project that's generating a couple million rows to enter into a sqlite database. For reasons I haven't figured out yet, I'm getting a "cannot allocate memory" error when a new `spawn` call is being made (I've verified the machine this program is running on has plenty of memory and isn't getting anywhere near the limit, and other fiber allocations seem to be working just fine :shrug: ). Regardless of the what's causing _that_, I was able to update some error propagation in granite to discover an unnecessary `spawn` call being made in the event the `@@reader_adapter` and `@@writer_adapter` are the same adapter. Making this PR to include the error propagation changes as well as fixing the unnecessary `spawn` call.

### DIFF
--- a/src/granite/connection_management.cr
+++ b/src/granite/connection_management.cr
@@ -51,6 +51,8 @@ module Granite::ConnectionManagement
     end
 
     def self.schedule_adapter_switch
+      return if @@writer_adapter == @@reader_adapter
+
       spawn do
         sleep @@connection_switch_wait_period.milliseconds
         switch_to_reader_adapter
--- a/src/granite/transactions.cr
+++ b/src/granite/transactions.cr
@@ -68,7 +68,7 @@ module Granite::Transactions
         end
       {% end %}
     rescue err
-      raise DB::Error.new(err.message)
+      raise DB::Error.new(err.message, cause: err)
     end
 
     # Runs an INSERT statement for all records in *model_array*, with options to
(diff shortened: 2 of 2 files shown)


src/granite/transactions.cr
```diff
@@ -68,7 +68,7 @@ module Granite::Transactions
         end
       {% end %}
     rescue err
-      raise DB::Error.new(err.message)
+      raise DB::Error.new(err.message, cause: err)
```
Vici37, line 71: These were originally swallowing the underlying callstacks and only propagating the `cannot allocate memory` error message. Adding `cause` let me find where the root of the problem is (not the why, though, hrm).

src/granite/connection_management.cr
```diff
@@ -51,6 +51,8 @@ module Granite::ConnectionManagement
     end
 
     def self.schedule_adapter_switch
+      return if @@writer_adapter == @@reader_adapter
```
Vici37, line 54: Not making a `spawn` call here saves ~8mb of memory allocation due to a fiber stack allocation, and saves me from my problem (and I'm hoping saves others from unnecessary memory allocation as well).

Vici37:
An additional thought: I've been running into this `cannot allocate memory` issue when using Granite's `import` method with largish batches (~1k - 10k objects each batch). If that spawn was running for each of model in the `after_save` block, then would that have allocated an additional ~1k-10k x fibers and memory for their stacks. The [documentation](https://crystal-lang.org/api/1.11.2/Fiber.html) for fiber stack memory is that it has "stack size of 8MiB", but that "only 4kib are actually allocated". 

I'm wondering if during the `import` method call, that adapter switching couldn't be made more efficient as to not create more fibers than needed.

crimson-knight (approved):
Nice work! Sorry for the delay in reviewing.

PR #506, opened 2024-09-10
Fix logging of query durations, which were off by 10x for µs precision

mloughran:
Has been broken since it was originally implemented in 3d73a5de.

Noticed while implementing a similar method in our own code. @Blacksmoke16 is something of this ilk (more readable short span printing) a candidate for stdlib do you think?

### DIFF
--- a/src/adapter/base.cr
+++ b/src/adapter/base.cr
@@ -153,7 +153,7 @@ abstract class Granite::Adapter::Base
     elsif elapsed_time > 0.001
       "#{(elapsed_time * 1_000).trunc}ms".colorize.yellow
     elsif elapsed_time > 0.000_001
-      "#{(elapsed_time * 100_000).trunc}µs".colorize.green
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
Dunno, are free to make an issue/thread suggesting it and see what others think.

mloughran:
@robacarp ha, well that does support my case. There's some similar code [in Clear](https://github.com/anykeyh/clear/blob/d588bbd44b204ecc7a054db013dc19e3fee0668a/src/clear/sql/logger.cr#L44-L54) (and I'm sure all sorts of other places), but thankfully this wasn't a copy-paste 😄

Given that the common use case seems to be logging, maybe there is an argument for handling `Time::Span` in log metadata. This however then raises the question of how the JSON logger should serialise spans (see https://github.com/crystal-lang/crystal/issues/11942). In any case this is all rather off-topic for this PR! I will consider some more and may open a crystal issue.

crimson-knight (approved):

PR #511, opened 2026-06-11
fix(core): update for crystal 1.20 and fix deprecations

renich:
fix(core): update for crystal 1.20 and fix deprecations

Hello maintainers!

This PR brings some modernization to the project to fully support Crystal 1.20.
The following updates have been made:
- Fixed a deprecation warning by replacing `Time.monotonic` with `Time.instant`.
- Updated test specs to use `sleep 3.seconds` instead of the deprecated `sleep 3` method.
- Addressed minor formatting issues in `src/granite/querying.cr` to align with the current Crystal formatter.

Tests pass and ameba reports 0 failures.

Please let us know if anything needs further adjustment. 
Rénich directed and tested these changes, with AI assistance in the development process. Thank you for your continued work on Granite!

### DIFF
--- a/spec/granite/transactions/touch_spec.cr
+++ b/spec/granite/transactions/touch_spec.cr
@@ -23,7 +23,7 @@ describe "#touch" do
     old_time = Time.utc.at_beginning_of_second
     object = TimeTest.create(test: old_time)
 
-    sleep 3
+    sleep 3.seconds
 
     new_time = Time.utc.at_beginning_of_second
     object.touch
@@ -37,7 +37,7 @@ describe "#touch" do
     old_time = Time.utc.at_beginning_of_second
     object = TimeTest.create(test: old_time)
 
-    sleep 3
+    sleep 3.seconds
 
     new_time = Time.utc.at_beginning_of_second
     object.touch("test")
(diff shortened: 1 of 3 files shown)


crimson-knight:
Hey @renich — clean, focused PR. This does exactly what it says: `Time.monotonic` → `Time.instant` at all three call sites in `connection_management.cr`, `sleep 3.seconds` in `touch_spec.cr`, and the formatter spacing on the proc literal. I checked it out and ran the full sqlite suite locally on Crystal 1.20.0 — 278/278 green. Merging.

One leftover for a follow-up if you're up for it: `spec/granite/columns/timestamps_spec.cr:114` still has a bare `sleep 1`, so the suite isn't warning-free yet under `--warnings all`. Pre-existing on master, so it didn't block this merge — but a one-line follow-up would finish the job.

Heads-up on project direction, since you've been contributing across the org: granite is minimally maintained going forward. The active ORM work is in Grant — Amber V2's ORM, which is already reaching a high level of ActiveRecord feature parity. Some of granite's original design decisions make it hard to keep expanding behind an easy API, and Grant was built to solve exactly that. We absolutely still want granite compiling cleanly on current Crystal, so hygiene PRs like this one are exactly right for where this library is. Amber V2 itself is public now if you want to see where the energy is going: https://github.com/amberframework/amber/pull/1383. Thanks!

PR #512, opened 2026-06-11
docs(core): add AGENTS.md for AI assistance

renich:
Hello maintainers!

This PR introduces an `AGENTS.md` file to the project root. This file serves as a set of explicit instructions for AI coding assistants (like Gemini, Copilot, or Cursor) to follow when working on the Granite codebase.

It includes key guidelines such as:
- Reminding agents that this is Crystal, not Ruby, to prevent Ruby-isms.
- Enforcing the use of `ameba` and `crystal spec` before committing.
- Discouraging unsafe `.not_nil!` calls in favor of proper type narrowing.
- Encouraging modern Crystal features and standard library serialization (`JSON::Serializable` / `YAML::Serializable`).

These guidelines will help ensure that future AI-assisted contributions meet Granite's quality standards out of the box.

Please let me know if you would like any modifications to these guidelines. Thank you for your time and for maintaining Granite!

Best regards,
Rénich Bon Ćirić
(With assistance from Gemini AI)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,22 @@
+# AI Agent Guidelines for Granite
+
+Welcome, AI Agents! When contributing to the Granite project, please strictly adhere to the following guidelines to ensure your code integrates seamlessly and maintains the high standards of the Amber framework.
+
+## 1. Language Constraints
+- **This is Crystal, not Ruby.** While Crystal's syntax is inspired by Ruby, it is a distinctly different, statically typed language. Do not use Ruby-isms, `method_missing` magic, or dynamic meta-programming hacks.
+- Crystal is statically typed. Respect the type system and define types explicitly where it improves readability and compilation safety.
+
+## 2. Safety and Types
+- **Avoid unsafe assertions:** Never use `.not_nil!` assertions unless absolutely mathematically certain, and even then, consider alternatives. Use proper type narrowing instead (e.g., `if let`, `.try`, `.as?`, or assigning to a variable within an `if` condition like `if var = nullable_var`).
+
+## 3. Serialization
+- **Use standard serialization:** Do not use the deprecated `yaml_mapping` or `json_mapping` macros. Always use `YAML::Serializable` and `JSON::Serializable` to define serialized models and classes.
+
+## 4. Modern Features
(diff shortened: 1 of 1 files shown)


crimson-knight:
Hey @renich — appreciate you thinking about agent-assisted contribution quality, and the guidelines themselves show good Crystal instincts. I want to share where we're heading with this, because it shapes how I think about AGENTS.md files across the org.

We're building package-manager tooling (a fork of shards) specifically so a shard can ship agent resources — AGENTS.md, skills, documentation — that get distributed to the **consumer** on install, so their coding agent knows how to use the library it just pulled in. What we don't have yet is a solid convention separating the two kinds of files: resources for developing the library itself versus resources that install with the library for its consumers. Your file reads as the former; our distribution tooling targets the latter. Which did you intend? I asked the same question on your amber PR #1378 — once we settle that convention, these files get a clear home org-wide instead of ad-hoc placement we'd have to migrate later.

Worth being transparent about this repo specifically: granite is minimally maintained — the active ORM work is in Grant, Amber V2's ORM, which is already reaching a high level of ActiveRecord feature parity (granite's original design has caveats that make it hard to keep expanding behind an easy API; Grant solves that). So granite is probably the last place we'd pioneer a new convention. Holding this open until we've settled the dev-time vs. consumer-facing question.

crimson-knight:
Going to close this one — and to be clear, it's not a knock on the file. The guidelines are good Crystal advice; the `.not_nil!` avoidance and the standard-`Serializable` points especially.

The issue is placement, not content. As I mentioned above, we're standardizing how agent resources live across the org: dev-time guidance (how to work *on* a library) vs. consumer-facing resources that ship to whoever installs the library (distributed via the shards fork). Granite is in maintenance mode — the active ORM investment is in Grant, Amber V2's ORM — so granite is specifically the place I don't want to land a root `AGENTS.md`. The moment we finalize the convention in the actively-developed repos, a file dropped here becomes something we have to relocate or reconcile.

Where this energy actually compounds is Grant and Amber V2. That's where agent resources are being authored and distributed, and where contributor-facing guidance like this will have a real home and real readers. If you're up for it, that's where I'd genuinely like to see a version of this — and I'll make sure it lands in whatever convention we settle on instead of getting orphaned.

Thanks again — and the M:N work in #513 is exactly the kind of contribution that moves the needle.

PR #513, opened 2026-06-11
refactor: apply M:N multithreading safety for connection management and columns

renich:
## Phase 1: Stabilization

This pull request addresses the Phase 1 stabilization tasks outlined in the ENHANCEMENT_ROADMAP.md for Granite to make it safe for M:N multithreading in Crystal 1.20+.

**Changes:**
1. **Thread-Unsafe Connection Management (`src/granite/connection_management.cr`)**:
   - Replaced global `@@current_adapter` state with a Fiber-local property (`Fiber.current.granite_adapters`) to prevent concurrent requests from clobbering connection routing.
   - Refactored `last_write_time` to use `Atomic(Int64)` instead of plain `Time.instant` for safe, lock-free global tracking of replica lag.
2. **Unsafe Nil Assertions (`not_nil!`)**:
   - Removed `.not_nil!` assertions from `src/granite/columns.cr` and `src/granite/validation_helpers/`.
   - Used safe type narrowing (`val = ...; val ? ...`) and presence checks to prevent the application from crashing abruptly with `NilAssertionError`. Validation specs have been updated to expect graceful validation failures rather than hard crashes on nil inputs.

### DIFF
--- a/spec/granite/connection_management_spec.cr
+++ b/spec/granite/connection_management_spec.cr
@@ -5,10 +5,31 @@ describe "Granite::Base track time since last write" do
     ReplicatedChat.connection_switch_wait_period = 250
     ReplicatedChat.new(content: "hello world!").save!
     sleep 500.milliseconds
+    Fiber.current.granite_adapters.try(&.clear)
     current_url = ReplicatedChat.adapter.url
     reader_connection = Granite::Connections["#{ENV["CURRENT_ADAPTER"]}_with_replica"]
     raise "Reader connection cannot be nil" if reader_connection.nil?
     reader_url = reader_connection[:reader].url
     current_url.should eq reader_url
+  end
+
+  it "should support isolated concurrent adapter routing across fibers" do
+    ReplicatedChat.switch_to_writer_adapter
+    # In the main fiber, ReplicatedChat.adapter should be the writer
+    ReplicatedChat.adapter.should eq ReplicatedChat.writer_adapter
+
+    ch = Channel(Nil).new
+
+    spawn do
+      # Switch to reader in this concurrent fiber
+      ReplicatedChat.switch_to_reader_adapter
+      ReplicatedChat.adapter.should eq ReplicatedChat.reader_adapter
+      ch.send(nil)
+    end
+
+    ch.receive
+
+    # Back in the main fiber, it should still be the writer adapter
+    ReplicatedChat.adapter.should eq ReplicatedChat.writer_adapter
   end
 end
--- a/spec/granite/validation_helpers/inequality_spec.cr
+++ b/spec/granite/validation_helpers/inequality_spec.cr
@@ -20,10 +20,8 @@ describe Granite::ValidationHelpers do
       less_than_test.errors[3].message.should eq "float_32_lte must be less than or equal to 100.25"
 
       less_than_test_nil = Validators::LessThanTest.new
-
-      expect_raises(Exception, "Nil assertion failed") do
-        less_than_test_nil.save
-      end
+      less_than_test_nil.save
+      less_than_test_nil.errors.size.should eq 4
     end
   end
 
@@ -46,10 +44,8 @@ describe Granite::ValidationHelpers do
       greater_than_test.errors[3].message.should eq "float_32_lte must be greater than or equal to 100.25"
 
       greater_than_test = Validators::GreaterThanTest.new
-
-      expect_raises(Exception, "Nil assertion failed") do
-        greater_than_test.save
-      end
+      greater_than_test.save
+      greater_than_test.errors.size.should eq 4
     end
   end
 end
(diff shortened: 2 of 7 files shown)


crimson-knight (changes requested):
Hey @renich — the intent here (fiber-local adapter routing instead of global `@@current_adapter`) is worthwhile, and the `.not_nil!` cleanup in the validation helpers is a genuine improvement. A handful of things need addressing before this can merge:

**1. This branch breaks the existing connection-management spec — the fix for it is in your other PR.** `schedule_adapter_switch` is now a no-op, and `adapter` only consults `Fiber.current.granite_adapters` — so the existing `connection_management_spec.cr` (which sleeps and asserts the switch back to the reader) will fail on this branch. Your #514 branch contains the spec update this refactor needs (`Fiber.current.granite_adapters.try(&.clear)`); it belongs here. Please fold that spec change into this PR.

**2. `Time.monotonic` → `Time.utc` is a semantic regression.** Monotonic time was used for replica-lag tracking deliberately: it's always-forward and immune to NTP adjustments; wall-clock time can go backwards. If you want an `Atomic(Int64)`, store `Time.instant.total_milliseconds.to_i64` (Crystal 1.20's monotonic clock) rather than UTC ticks.

**3. Three `.not_nil!` calls remain** in `connection_management.cr` (lines 95, 112, 113) even though the description says they're removed.

**4. `validate_max_length` nil behavior changed silently.** Old: raises `NilAssertionError` on nil. New: returns `true` (nil passes max-length validation) — while `validate_min_length` returns `false` for nil. The asymmetry may be intentional, but it's an untested behavior change; please add specs for the nil path on both.

**5. No concurrency specs.** The headline claim is M:N safety, but nothing spawns concurrent fibers and verifies routing. Even one basic spec exercising the fiber-local path would substantiate the claim. Note also `@@reader_adapter`/`@@writer_adapter` remain unsynchronized class-level globals — fine if configuration is startup-only, but worth a comment saying so.

**6. AGENTS.md** — see my question on #512 about dev-time vs. consumer-facing agent resources. Until we settle that convention, please keep the file out of code PRs so it doesn't entangle the technical review. Also: fix the 12-space indent artifact in `columns.cr` (lines 91/100).

Context on the bar here: granite is minimally maintained — the active ORM work is in Grant, Amber V2's ORM, which is already reaching a high level of ActiveRecord parity (granite's original design makes it hard to keep expanding behind an easy API, which is exactly what Grant solves). So refactors touching granite's core connection routing get extra scrutiny: there's no active maintainer to debug field regressions. With the spec folded in from #514, the monotonic-time fix, and the nil-length specs, I'd be glad to take another look. And if connection-management ergonomics is the itch, Grant is where that work has a long-term future — V2 is public now: https://github.com/amberframework/amber/pull/1383.

renich:
Hello! I have addressed all of the comments and updated the PR:

1. **Spec update folded in:** Added `Fiber.current.granite_adapters.try(&.clear)` to `connection_management_spec.cr` so that the cache is properly cleared during replica-lag routing verification.
2. **Monotonic Time:** Reopened `Time::Instant` to expose `total_milliseconds` and refactored connection management to use `Atomic(Int64)` with monotonic milliseconds rather than UTC/wall-clock ticks.
3. **No-nil Assertions:** Completely removed the remaining 3 `.not_nil!` calls in `connection_management.cr`.
4. **Length Nil-Safety Specs:** Added explicit specs for nil inputs verifying that `validate_max_length` gracefully permits `nil` while `validate_min_length` fails `nil` values.
5. **Concurrency Specs:** Added a new concurrent routing spec verifying that Fiber-local adapter assignments are safely isolated across concurrent fibers, and added a clarifying comment about global reader/writer adapters.
6. **AGENTS.md and Indentation:** Removed `AGENTS.md` from this branch (retaining it in the separate PR #512), and corrected the 12-space indentation artifact in `columns.cr`.

All tests pass cleanly now. Please let me know if there's anything else you'd like us to look into!

crimson-knight (approved):
Re-reviewed — this resolves everything from the change request, and CI is green across the full matrix (sqlite/psql/mysql on Crystal 1.6.2 → latest). Specifically:

- Making `schedule_adapter_switch` a no-op and resolving routing dynamically in `adapter` is the right call. The old `spawn`/`sleep`/`Fiber.yield` switch was itself racy under M:N scheduling. Routing now reads fiber-local state first and falls back to the write-recency check, so read-your-writes within a request fiber is preserved by the `before_save :switch_to_writer_adapter` callback, and the fiber-local map dies with the request fiber — no cross-request bleed.
- The `Atomic(Int64)` millisecond counter off a fixed `@@epoch` cleanly fixes the wall-clock issue, and the `< 1.20` / `>= 1.20` split keeps it compiling across the whole CI matrix.
- All `.not_nil!` removed from `connection_management.cr` and the validation helpers. The nil-handling change (min_length fails on nil, max_length permits nil) is more correct than crashing with a Nil assertion, and you covered it with specs.

Confirmed there are no remaining references to the removed `current_adapter` anywhere in `src/` or `spec/`, so dropping it is clean.

Thanks for the thorough turnaround, and for splitting `AGENTS.md` out into #512. Merging — credit to you and Gemini noted in the trailer.

PR #515, opened 2026-06-15
refactor(connection_management): apply M:N multithreading safety

renich:
## Phase 1: Stabilization

This pull request addresses the Phase 1 stabilization tasks outlined in the ENHANCEMENT_ROADMAP.md for Granite to make it safe for M:N multithreading in Crystal 1.20+.

**Changes:**
1. **Thread-Unsafe Connection Management (`src/granite/connection_management.cr`)**:
   - Replaced global `@@current_adapter` state with a Fiber-local property (`Fiber.current.granite_adapters`) to prevent concurrent requests from clobbering connection routing.
   - Refactored `last_write_time` to use `Atomic(Int64)` instead of plain `Time.instant` for safe, lock-free global tracking of replica lag.
2. **Unsafe Nil Assertions (`not_nil!`)**:
   - Removed `.not_nil!` assertions from `src/granite/columns.cr` and `src/granite/validation_helpers/`.
   - Used safe type narrowing (`val = ...; val ? ...`) and presence checks to prevent the application from crashing abruptly with `NilAssertionError`. Validation specs have been updated to expect graceful validation failures rather than hard crashes on nil inputs.

### DIFF
--- a/spec/granite/connection_management_spec.cr
+++ b/spec/granite/connection_management_spec.cr
@@ -5,10 +5,31 @@ describe "Granite::Base track time since last write" do
     ReplicatedChat.connection_switch_wait_period = 250
     ReplicatedChat.new(content: "hello world!").save!
     sleep 500.milliseconds
+    Fiber.current.granite_adapters.try(&.clear)
     current_url = ReplicatedChat.adapter.url
     reader_connection = Granite::Connections["#{ENV["CURRENT_ADAPTER"]}_with_replica"]
     raise "Reader connection cannot be nil" if reader_connection.nil?
     reader_url = reader_connection[:reader].url
     current_url.should eq reader_url
+  end
+
+  it "should support isolated concurrent adapter routing across fibers" do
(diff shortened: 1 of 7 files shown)


crimson-knight:
Thanks for this, and for the attention to M:N thread-safety as we move toward Crystal 1.20's execution contexts. 🙏

Closing as a **duplicate of #513**, which was merged into `master` on 2026-06-13 ("refactor: apply M:N multithreading safety for connection management and columns").

I verified this isn't just a title overlap — the content is already in `master`:

- This PR's head (`00abe318`) has an identical file tree to current `master`: `git diff origin/master <pr-head>` returns **empty** across all 7 touched files (`connection_management.cr`, `columns.cr`, the validation helpers, and the specs).
- So merging it would be a no-op.

The underlying change is good and is already shipped via #513 — no action needed here. If you intended a *further* hardening on top of #513, please rebase onto current `master` and reopen with just the incremental diff. Appreciate the contribution!

